PR #10, opened 2016-03-22
Q has no method 'getUnhandledReasons'

tarmolov:
When I specified `debug` option in the config file, I'll got the following error:
```javascript
TypeError: Object function Q(value) {
   // If the object is already a Promise, return it directly.  This enables
   // the resolve function to both be used to created references from objects,
   // but to tolerably coerce non-promises to promises.
   if (Q_isPromise(value)) {
       return value;
   } else if (isThenable(value)) {
       if (!thenables.has(value)) {
           thenables.set(value, new Promise(new Thenable(value)));
       }
       return thenables.get(value);
   } else {
       return new Promise(new Fulfilled(value));
   }
} has no method 'getUnhandledReasons'
```

`hermione` uses q library v2 which doesn't have `getUnhandledReasons` method anymore. 
It seems that `q-debug-mode` is useless now and can be pruned.

/cc @j0tunn

### DIFF
--- a/lib/hermione.js
+++ b/lib/hermione.js
@@ -6,9 +6,7 @@ var Runner = require('./runner'),
     pathUtils = require('./path-utils'),
     logger = require('./utils').logger,
     _ = require('lodash'),
-    q = require('q'),
     inherit = require('inherit'),
-    qDebugMode = require('q-debug-mode'),
     util = require('util'),
     chalk = require('chalk');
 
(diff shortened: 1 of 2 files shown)


SwinX:
Seems `q` is not being used anywhere except `qDebugMode` in `lib/hermione.js`. Need to remove in order to fix lint issues.

SwinX:
:ok: for PR.

tarmolov:
:shipit:

SwinX:
Thank you!

PR #27, opened 2016-05-16
Decrease default waitTimeout value

sipayRT:
@j0tunn @eGavr @tormozz48


eGavr:
A little bit more information about this change, please :)

PR #29, opened 2016-06-21
Fix some documentation issues

andre487:
I have replaced TDD terms with BDD terms. Don't forget about BDD Mocha specifics.

And colon was added to JSON listing with browsers params.


sipayRT:
tnx!

PR #123, opened 2017-03-01
fix: bug with file path on 'before all' hook

DudaGod:


DudaGod:
@sipayRT @rostik404

rostik404:
🆗

PR #308, opened 2018-09-08
feat: add "antialiasingTolerance" option

DudaGod:


test/lib/config/browser-options.js
```diff
@@ -765,74 +765,76 @@ describe('config browser-options', () => {
         });
     });
 
-    describe('tolerance', () => {
-        describe('should throw an error', () => {
-            it('if value is not number', () => {
+    ['tolerance', 'antialiasingTolerance'].forEach((option) => {
+        describe(`${option}`, () => {
+            describe('should throw an error', () => {
+                it('if value is not number', () => {
```
xrsd: `is not a number`

xrsd: /ok

PR #309, opened 2018-09-11
feat: Add event RUNNER_BEFORE_END

WalkInWay:
It need when you have to make operations after run tests but before close threads


j0tunn:
Can you describe your use case please?

WalkInWay:
Sorry, my bad. It's not needed

PR #358, opened 2019-01-18
fix: meta randomly does not contain 'pid'

eGavr:


lib/browser/existing-browser.js
```diff
@@ -14,7 +14,11 @@ module.exports = class ExistingBrowser extends Browser {
 
         this._camera = Camera.create(this._config.screenshotMode, () => this._takeScreenshot());
 
-        this._meta = _.extend({pid: process.pid}, this._config.meta);
+        this._meta = this._initMeta();
+    }
+
+    _initMeta() {
+        return _.extend({pid: process.pid}, this._config.meta);
```
j0tunn:
```js
return {pid: process.pid, ...this._config.meta};
```
?

PR #375, opened 2019-02-14
feat: add tolerance option to assert view command

next0:


lib/browser/commands/assert-view/index.js
```diff
@@ -44,7 +44,13 @@ module.exports = (browser) => {
         }
 
         const {canHaveCaret, pixelRatio} = page;
-        const imageCompareOpts = {tolerance, antialiasingTolerance, canHaveCaret, pixelRatio, compareOpts};
+        const imageCompareOpts = {
+            tolerance: typeof opts.tolerance === 'number' ? opts.tolerance : tolerance,
```
j0tunn: It would be better to set default `opts.tolerance` in line 20. Something like:
```js
opts = _.defaults(opts, {ignoreElements: [], tolerance});
...
const imageCompareOpts = {
    tolerance: opts.tolerance
...
```

test/lib/browser/commands/assert-view/index.js
```diff
@@ -241,6 +241,21 @@ describe('assertView command', () => {
...
+            const browser = stubBrowser_(config);
+
+            browser.prepareScreenshot.resolves({canHaveCaret: 'foo bar', pixelRatio: 300});
+
+            await browser.publicAPI.assertView(null, null, {tolerance: 500});
+
+            assert.calledOnceWith(
+                Image.compare,
+                sinon.match.any, sinon.match.any,
+                {canHaveCaret: 'foo bar', tolerance: 500, antialiasingTolerance: 200, pixelRatio: 300, compareOpts: {stopOnFirstFail: true}}
```
j0tunn: you do not need to copy-paste whole previous test. You should test only tolerance overwriting here:
```js
const config = mkConfig_({tolerance: 100});
...
assert.calledOnceWith(
    Image.compare,
    sinon.match.any,
    sinon.match.any,
    sinon.match({tolerance: 500})
);
```

README.md
```diff
@@ -970,6 +970,7 @@ Parameters:
  - selector (required) `String|String[]` – DOM-node selector that you need to capture
  - opts (optional) `Object`:
    - ignoreElements (optional) `String|String[]` – elements, matching specified selectors will be ignored when comparing images
+   - tolerance (optional) `Number` – maximum allowed [CIEDE2000](http://en.wikipedia.org/wiki/Color_difference#CIEDE2000) difference between colors. See [browsers.tolerance](#tolerance)
```
j0tunn: Please, don't copy-paste tolerance description here. Just tell that this option overwrites tolerance from config for given assertView and give a link to base tolerance description

next0:
Fixed. @j0tunn, can you check it one more time?

j0tunn:
@sipayRT please, add this PR to the next hermione version

PR #379, opened 2019-02-21
feat(config): add parallelLimit option

### LINKED ISSUE #319: Add parallelLimit option as it is done in Gemini.
It is required to have an ability to limix overall amount of simultaneous sessions for testing on Browserstack. Currently I got 30 browser x breakpoint combinations and getting errors from BS that my parallel limit reached.

sipayRT:
fix #319

### DIFF
--- a/README.md
+++ b/README.md
@@ -663,6 +663,9 @@ Hermione runs all tests in subprocesses in order to decrease the main process CP
 #### testsPerWorker
 The maximum number of tests to be run in one worker before it will be restarted.
 
+#### parallelLimit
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -663,6 +663,9 @@ Hermione runs all tests in subprocesses in order to decrease the main process CP
 #### testsPerWorker
 The maximum number of tests to be run in one worker before it will be restarted.
 
+#### parallelLimit
+By default, `hermione` will run all browsers simultaneously. Sometimes (i.e. when using cloud services, such as SauceLabs) you have a limit of a number of browser that can be run once at a time. Use this option to limit the number of browsers that `hermione` will try to run in parallel. Default value is `Infinity`.
```
Flackus: I would rather say:

> ... you have to limit the amount of browsers that can be run at the same time. This option effectively limits how many browsers `hermione` will try to run in parallel.

PR #414, opened 2019-06-13
feat: API: increase test enabling-disabling speed

j0tunn:


j0tunn:
I will create other PR

PR #421, opened 2019-07-17
feat: ability to guarantee tests order after reading

j0tunn:


lib/hermione.js
```diff
@@ -72,6 +72,12 @@ module.exports = class Hermione extends BaseHermione {
         const specs = await testReader.read({paths: testPaths, browsers, ignore, sets, grep});
         const collection = TestCollection.create(specs);
 
+        collection.getBrowsers().forEach((bro) => {
+            if (this._config.forBrowser(bro).strictTestsOrder) {
+                collection.sortTests(bro, ({id: a}, {id: b}) => a < b ? -1 : 1);
```
eGavr:
```js
a < b ? -1 : 1 //вроде можно просто `a - b` 
```

j0tunn: это строчки

PR #457, opened 2019-10-31
feat: add ability to work with async functions in TestCollection

missvalentinep:
add analogue to TestCollection.eachTest that can handle async function as a callback

### DIFF
--- a/lib/test-collection.js
+++ b/lib/test-collection.js
@@ -80,6 +80,23 @@ module.exports = class TestCollection {
         }
     }
 
+    async eachTestAsync(browserId, cb) {
+        if (_.isFunction(browserId)) {
+            cb = browserId;
+            browserId = undefined;
+        }
+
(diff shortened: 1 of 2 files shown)


lib/test-collection.js
```diff
@@ -80,6 +80,23 @@ module.exports = class TestCollection {
...
 
+    async eachTestAsync(browserId, cb) {
+        if (_.isFunction(browserId)) {
+            cb = browserId;
+            browserId = undefined;
+        }
+
+        if (browserId) {
+            for (const test of this._specs[browserId]) {
+                await cb(test, browserId);
```
miripiruni: Тут можно паралельно выполнять… можно попробовать взять `p-map` c `concurrency`.

missvalentinep:
using TestCollection.mapTests can do the same thing and we won't need to add an extra method

PR #474, opened 2020-02-20
fix: use only consistent sort functions

Hypnosphi:
Issue: tests are failing with Node 12, because the used sorting functions are not consistent (i.e. asymmetric)

Context:
https://github.com/nodejs/node/issues/24294
https://v8.dev/blog/array-sort

### DIFF
--- a/lib/hermione.js
+++ b/lib/hermione.js
@@ -74,7 +74,15 @@ module.exports = class Hermione extends BaseHermione {
 
         collection.getBrowsers().forEach((bro) => {
             if (this._config.forBrowser(bro).strictTestsOrder) {
-                collection.sortTests(bro, ({id: a}, {id: b}) => a < b ? -1 : 1);
+                collection.sortTests(bro, ({id: a}, {id: b}) => {
+                    if (a < b) {
+                        return -1;
+                    } else if (a > b) {
+                        return 1;
+                    } else {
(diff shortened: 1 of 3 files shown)


DudaGod:
Hello.

In hermione all tests have uniq full titles. It means there are no case in which `test1.id` equal to `test2.id`. And tests are sorted for each browser separately. So it looks like that there is no point in returning 0 for equal values.

Are you sure that your tests were failing because of this problem? Can you provide minimal case to reproduce?

DudaGod:
> Maybe the problem was that Number(a.title < b.title) is 0 if a.title < b.title is false. But it should be -1 instead

No, we use code like - `collection.sortTests(bro, ({id: a}, {id: b}) => a < b ? -1 : 1);`.
So if `a` less than `b` (`a` and `b` string hashes) then we will use `-1` otherwise we use `1`. Case in which `a === b` does not exist.

> Sorry, I don't remember

Sorry for the very long reply =(
I'm closing this issue because I don't see the problem. 
Thank you.

DudaGod:
Yeah, but it's just a simple unit test and it doesn't affect any users. Moreover you can set any sort function to `collection.sortTests` even if it doesn't work correctly.

But okay, I will fix it in this PR - https://github.com/gemini-testing/hermione/pull/643.

Issue #442, opened 2019-09-17
Use readable folder names for screenshots


utrotzek:
When I configure the screenshot folder as follows

`screenshotsDir: 'tests/screenshots'`

and define a test like this: 

```javascript
describe('headlines', function() {
    it('will render global configured headlines properly', function() {
        return this.browser
            .url('/headlines')
            .assertView('headlines', '#content');
    });
});
```
the screenshots will be stored in

`tests/screenthos/abfebfdab/chrome/headlines.jpg`

Is there a way to replace the hashed folder name (abfebfdab) by the name of the test suite? Otherwise the folder structure is a little bit confusing to me.

qjac:
I also want folders to be named using the test name and, ideally, grouped according to test suite.

web-padawan:
For anyone who needs this, there is an undocumented option to pass a function:

https://github.com/gemini-testing/hermione/blob/d9fe04bf1aad2c70c88028b30bf39ec2f6e39d92/lib/config/browser-config.js#L15-L17

So in my project, I have configured it like this:

```
  browsers: {
    chrome: {
      baseUrl: 'http://localhost:8080/test/visual/',
      screenshotsDir: () => 'test/visual/screens/chrome',
      desiredCapabilities: {
        browserName: 'chrome',
        version: '80.0',
        platform: 'Windows 10'
      }
    }
  },
```

Note, function takes a single parameter which is basically a mocha test context.
So you can use that to get the test title or any other information you need.

DudaGod:
@web-padawan thank you for the answer. You are right. For described case function `screenshotsDir` is the right way to save screenshot with name from suite.

To get suite title from test you should write:
```
screenshotsDir: (test) => test.parent.title
```

utrotzek:
Awesome, this works like a charm. Thank you!

PR #524, opened 2020-09-17
chore: add screenshotsDir function example

web-padawan:
While the function syntax for `screenshotsDir` is documented, a code example would be nice for discoverability. 

This prevent recurring questions like #442 (asked again today in the self-support Telegram chat).


DudaGod:
Hello.

Thank you for your contribution.

PR #530, opened 2020-09-27
feat: added 'hasBrowserVersionOverwriten' property for test

unlok:
Added property for a test to be able to determine if a browser version of the test was overwritten from the native helper `browser().version()`.

### DIFF
--- a/lib/test-reader/browser/commands/version.js
+++ b/lib/test-reader/browser/commands/version.js
@@ -9,7 +9,10 @@ module.exports = class VersionCommand extends BaseCommand {
     }
 
     execute(browserVersionToSet) {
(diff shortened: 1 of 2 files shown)


lib/test-reader/browser/commands/version.js
```diff
@@ -9,7 +9,10 @@ module.exports = class VersionCommand extends BaseCommand {
     }
 
     execute(browserVersionToSet) {
-        this._oneTimeHandler = (test) => test.browserVersion = browserVersionToSet;
+        this._oneTimeHandler = (test) => {
+            test.browserVersion = browserVersionToSet;
+            test.hasBrowserVersionOverwritten = true;
```
DudaGod, line 14: почему has, а не is?

PR #532, opened 2020-10-18
feat: Report stale screenshots

### LINKED ISSUE #334: Ability to cleanup outdated screens
Hi!
Is there is any ability to cleanup all screenshots tests for which were removed/moved to another suite?

Hypnosphi:
Fixes #334

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,6 +81,8 @@ Hermione is a utility for integration testing of web pages using [WebdriverIO v4
     - [Parallel execution plugin code](#parallel-execution-plugin-code)
   - [prepareBrowser](#preparebrowser)
   - [prepareEnvironment](#prepareenvironment)
+  - [reportStaleScreenshots](#reportstalescreenshots)
+  - [screenshotsRoot](#screenshotsroot)
 - [CLI](#cli)
   - [Reporters](#reporters)
   - [Require modules](#require-modules)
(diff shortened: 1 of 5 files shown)


DudaGod:
Hello.

To my mind it should be implemented as a separate command (for example - `npx hermione clean-screenshots`).  We already had an attempt to implement this but did not complete it to the end

Hypnosphi:
@DudaGod how can I get state names without running all the tests?

DudaGod:
> how can I get state names without running all the tests?

You need to implement custom runner in order to read all tests with a redefined `assertView` command. In this redefined command you will be able to collect all image states. Custom runner should not launch browsers. It's not an easy task but this is how it should be done.

Hypnosphi:
> Custom runner should not launch browsers

Then it's quite likely that tests won't reach the `assertView` stage

DudaGod:
> Then it's quite likely that some tests won't even reach the assertView stage, e.g. if they wait for some condition before it

If we overwrite all commands to return `Promise.resolve()` then `assertView` will be reach in all tests.

DudaGod:
Hello.

Ability to remove unused screens is already implemented in html-reporter@8.4.2. More info about this command is here - https://github.com/gemini-testing/html-reporter#remove-unused-screens.

I close it, thank you.

PR #547, opened 2021-02-18
fix: set timeout for execution screenshot on reject

DudaGod:
The example with hermione-test which hangs on execution:
```
it('test1', async function() {
  this.browser.addCommand('screenshot', () => {
    return new Promise((resolve, reject) => {
      console.log('BEFORE REJECT1!!!');
      reject(`some error!!!`);
    });
  }, true);
  
  await this.browser.foo();
});
```

This test will try to take screenshot on reject because `foo` method does not exists in `browser` but it will hangs forever because wdio@4 has custom promises implementation which does not work in that case correctly.

To fix this problem add timeout for execution screenshot on reject. Moreover this problem does not reproduce with wdio@6.

### DIFF
--- a/test/lib/worker/runner/test-runner/one-time-screenshooter.js
+++ b/test/lib/worker/runner/test-runner/one-time-screenshooter.js
@@ -4,6 +4,7 @@ const _ = require('lodash');
 const Promise = require('bluebird');
 const {Image} = require('gemini-core');
 const OneTimeScreenshooter = require('lib/worker/runner/test-runner/one-time-screenshooter');
+const logger = require('lib/utils/logger');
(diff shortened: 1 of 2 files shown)


test/lib/worker/runner/test-runner/one-time-screenshooter.js
```diff
@@ -148,5 +150,15 @@ describe('worker/runner/test-runner/one-time-screenshooter', () => {
 
             assert.calledOnce(browser.restoreHttpTimeout);
         });
+
+        it('should fail with timeout error on long execution "screenshot" command', async () => {
```
xrsd, line 154: `of "screenshot" command`

xrsd (approved):

PR #566, opened 2021-06-09
fix: ability to run tests using "devtools" protocol

DudaGod:


lib/browser/browser.js
```diff
@@ -73,7 +73,9 @@ module.exports = class Browser {
             sessionId,
             ...sessionOpts,
             ...detectedSessionEnvFlags,
-            ...this._config.sessionEnvFlags
+            ...this._config.sessionEnvFlags,
+            options: sessionOpts,
```
DudaGod, line 77: In new wdio version attach to existing browser with `devtools` protocol works without any problem when you pass instance of browser created using `remote` method. Like this:
```
const client = await remote(opts);
const client2 = await attach(client);
```

But it does not work when you pass instance from one process to another. I get error `route not found`. Looking to wdio code I found that `options` should be passed explicitly - https://github.com/webdriverio/webdriverio/blob/main/packages/webdriverio/src/index.ts#L85.

lib/browser/browser.js
```diff
@@ -73,7 +73,9 @@ module.exports = class Browser {
...
+            capabilities: {...sessionOpts.capabilities, ...sessionCaps}
```
DudaGod, line 78: It is necessary because otherwise when run tests using `devtools` protocol creates two instance of chrome browser for each test. First when call `remote` method in master process and second when call `attach` method in worker process.

The problem is here - https://github.com/webdriverio/webdriverio/blob/main/packages/devtools/src/launcher.ts#L196-L203.

When browser creates in master process it returns capabilities with - `'goog:chromeOptions': {debuggerAddress: 'localhost:12345'}` which is necessary to connect to it when call `attach` method. But capabilities from browser config don't know anything about it option. So i merge it.

lib/browser/browser.js
```diff
@@ -73,7 +73,9 @@ module.exports = class Browser {
             sessionId,
             ...sessionOpts,
```
sipayRT, line 74: are we really need `sessionOpts` here now?

sipayRT (approved):

lib/browser/browser.js
DudaGod: Yeah, they are used to correctly detect backend configuration, here - https://github.com/webdriverio/webdriverio/blob/main/packages/webdriverio/src/index.ts#L86.

Without this session does not attached when used webdriver protocol.

PR #636, opened 2022-04-12
Feat: pass browser into test as an argument

j0tunn:
```js
it('some test', async function() {
    await this.browser.url('foo/bar');
});
```
=>
```js
it('some test', async ({ browser }) => {
    await browser.url('foo/bar');
});
```


DudaGod (approved):
🔥

PR #707, opened 2023-01-04
fix: add err field to test result

epszaw:
`TestResult` contains `err` field, but it doesn't have `screenshot` field inside, but it exists.


DudaGod:
Thank you.

PR #812, opened 2023-11-29
fix: fix client code to work on IE11 (#811)

shadowusr:


shadowusr:
Merging this w/o review since it's just a cherry-pick from master.

PR #839, opened 2024-02-02
feat: issue templates

miripiruni:
How it works: https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#creating-issue-forms

### DIFF
--- a/.github/ISSUE_TEMPLATE/1-bug-report.yaml
+++ b/.github/ISSUE_TEMPLATE/1-bug-report.yaml
@@ -1,0 +1,78 @@
+name: Bug Report
+description: Create a bug report for Hermione
+labels: ["type: bug"]
+body:
+    - type: markdown
+      attributes:
+          value: |
+              If this issue affects many people in a company/big team, create a post for your company in the following discussion:
+              [https://github.com/gemini-testing/hermione/discussions/838](https://github.com/gemini-testing/hermione/discussions/838)
+              and link the issue in your post.
+
+              This will help us prioritize issues that affect more people.
+    - type: checkboxes
+      attributes:
+          label: Verify latest release
+          description: "Please run `npm install hermione@latest` to try the latest version of Hermione. Some issues may already be fixed in the latest release, so please verify that your issue reproduces before opening a new issue."
+          options:
+              - label: I verified that the issue exists in the latest Hermione release
+    - type: input
+      attributes:
+          label: Hermione version
+          description: Fill only if you are not using the latest major version (v8)
+    - type: dropdown
+      attributes:
+          label: Which area(s) of Hermione are affected? (leave empty if unsure)
+          multiple: true
+          options:
+              - "Hermione core"
+              - "Plugins"
+              - "Configuration"
+              - "CLI"
+              - "Runner"
+              - "Reporters"
+              - "API"
+    - type: input
+      attributes:
+          label: Link to the code that reproduces this issue or a replay of the bug
+          description: |
+              A link to a GitHub repository minimal reproduction. If a minimal reproduction can't be created please share a replay of the bug which doesn't require sharing a private repo.
+    - type: textarea
+      id: repro
+      attributes:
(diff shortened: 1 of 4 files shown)


miripiruni:
Later we can add more specific fields (like browsers, config, workflow, etc). But, I think, its good starting point :)

miripiruni:
Preview: https://github.com/miripiruni/hermione/issues/new/choose

Test bug issue: https://github.com/miripiruni/hermione/issues/1

You can try it on my fork :)

.github/ISSUE_TEMPLATE/3-feature-request.yaml
```diff
@@ -0,0 +1,51 @@
...
+              A clear and concise description of what workflow is meant to be improved. Example: "As a developer, I often want to do <something>, but I often face <problem>".
+      validations:
+          required: true
+    - type: textarea
+      id: solution
+      attributes:
+          label: Describe the solution you'd like
+          description: A clear and concise description of what you want to happen. Consider that Hermione is used by many people, and your particular use case might not make sense to implement in the core.
+      validations:
+          required: true
```
sipayRT, line 37: This is the only field that should be required. Please, specify other fields as optional

.github/ISSUE_TEMPLATE/2-regression-report.yaml
```diff
@@ -0,0 +1,62 @@
...
+description: Report a regression in Hermione
+labels: ["type: regression"]
+body:
+    - type: input
+      id: last-working-version
+      attributes:
+          label: Last Hermione version that worked
+          placeholder: e.g. 8.0.3
+      validations:
+          required: true
```
sipayRT, line 11: We can set this field as optional - if a minimal case is provided, we can find the broken version on our own

.github/ISSUE_TEMPLATE/1-bug-report.yaml
```diff
@@ -0,0 +1,73 @@
...
+    - type: input
+      attributes:
+          label: Link to the code that reproduces this issue or a replay of the bug
+          description: |
+              A link to a GitHub repository minimal reproduction. If a minimal reproduction can't be created please share a replay of the bug which doesn't require sharing a private repo.
+    - type: textarea
+      id: repro
+      attributes:
+          label: Reproduction steps
+          description: "How do you trigger this bug? Please walk us through it step by step."
```
sipayRT, line 44: Maybe, we should add some example with steps such as:
```
1. npm init hermione-app my-example
2. add reproduction step to the file tests/some.hermione.ts
3. ...
```

.github/ISSUE_TEMPLATE/1-bug-report.yaml
```diff
@@ -0,0 +1,73 @@
...
+          label: Describe the Bug
+          description: A clear and concise description of what the bug is.
+      validations:
+          required: true
+    - type: markdown
+      attributes:
+          value: ----------------------------
+    - type: textarea
+      attributes:
+          label: Expected Behavior
```
sipayRT, line 61: I think these 2 fields (current and previous) should be together, because when describing a bug, it will be easier to describe everything at once

sipayRT (approved):

.github/ISSUE_TEMPLATE/1-bug-report.yaml
KuznetsovRoman: I would swap this and `Describe the Bug` sections, so you could write:
- I expected <something>
- But got <something>

KuznetsovRoman (approved):

.github/ISSUE_TEMPLATE/1-bug-report.yaml
KuznetsovRoman: Maybe "2. add reproduction steps to the `my-example` dir"? Reproduction steps may also include hermione-config modifications

PR #845, opened 2024-02-19
chore(npm): removed @types/chalk

nerlihmax:
```
npm WARN deprecated @types/chalk@2.2.0: This is a stub types definition for chalk (https://github.com/chalk/chalk). chalk provides its own type definitions, so you don't need @types/chalk installed!
```


KuznetsovRoman:
Thank you for the contribution!

PR #848, opened 2024-02-21
Export `TestError` type

epszaw:
The type is required for Allure Hermione integration


DudaGod:
Thank you for contribution. When do you need a new version with this change?

epszaw:
@DudaGod it's not critical, I just add ts ban comments where it's required, but in future it'll be nice to get rid of the comments

PR #850, opened 2024-02-28
fix: ability to run tests in ".jsx" and ".tsx" files by default

DudaGod:
## What is done

User can write regular e2e tests in files with extension `.jsx` and `.tsx`. So I add it to option `fileExtensions` by default. When using components (like `<h1>hello</h1>`) inside these files, an error `SyntaxError: Unexpected token '<'` will occur. Will be fixed in next PR.


DudaGod:
> Will be fixed in next PR.

Will be merged with second PR in order to not destroy run if user mask will be matched on `jsx` or `tsx` components.

PR #851, opened 2024-02-29
feat: ability to read tests in ".jsx" and ".tsx" files

DudaGod:
## What is done

Added ability to read tests in `.jsx` and `.tsx` files. To do this, it was necessary to add a code transpiler - babel. Used in order to convert components like `<h1>foo</h1>` to code which can be required.

In this PR supports only commonjs because esm projects not working with Hermione at the moment.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,8 @@
     "typings"
   ],
   "scripts": {
-    "build": "tsc && npm run copy-static",
+    "build": "tsc && npm run copy-static && npm run build-bundle -- --minify",
+    "build-bundle": "esbuild ./src/bundle/index.ts --outdir=./build/src/bundle --bundle --format=cjs --platform=node --target=ES2021",
     "copy-static": "copyfiles 'src/browser/client-scripts/*' build",
     "check-types": "tsc --project tsconfig.spec.json",
     "clean": "rimraf build/ *.tsbuildinfo",
@@ -79,10 +80,16 @@
     "yallist": "3.1.1"
   },
   "devDependencies": {
+    "@babel/core": "7.24.1",
+    "@babel/plugin-transform-modules-commonjs": "7.24.1",
+    "@babel/plugin-transform-react-jsx": "7.23.4",
+    "@babel/preset-react": "7.24.1",
+    "@babel/preset-typescript": "7.24.1",
     "@commitlint/cli": "^19.0.3",
     "@commitlint/config-conventional": "^19.0.3",
     "@sinonjs/fake-timers": "10.3.0",
     "@swc/core": "1.3.40",
+    "@types/babel__core": "7.20.5",
     "@types/bluebird": "3.5.38",
     "@types/chai": "4.3.4",
     "@types/chai-as-promised": "7.1.5",
@@ -99,6 +106,7 @@
     "chai-as-promised": "7.1.1",
     "copyfiles": "2.4.1",
     "doctoc": "2.2.0",
+    "esbuild": "0.20.2",
     "escape-string-regexp": "1.0.5",
     "eslint": "8.25.0",
     "eslint-config-gemini-testing": "2.8.0",
@@ -107,6 +115,7 @@
     "jsdom": "^24.0.0",
     "jsdom-global": "3.0.2",
     "onchange": "7.1.0",
+    "pirates": "4.0.6",
     "prettier": "2.8.4",
     "proxyquire": "1.8.0",
     "rimraf": "4.1.2",
--- a/src/bundle/constants.ts
+++ b/src/bundle/constants.ts
@@ -1,0 +1,2 @@
+export const TRANSFORM_EXTENSIONS = [".jsx", ".cjsx", ".mjsx", ".tsx", ".ctsx", ".mtsx"];
+export const JS_EXTENSION_RE = /^\.([cm]?[tj]sx?|json)$/;
--- a/src/bundle/test-transformer.ts
+++ b/src/bundle/test-transformer.ts
@@ -1,0 +1,49 @@
+import * as nodePath from "node:path";
+import * as babel from "@babel/core";
+import { addHook } from "pirates";
+import { TRANSFORM_EXTENSIONS, JS_EXTENSION_RE } from "./constants";
+
+import type { NodePath, PluginObj, TransformOptions } from "@babel/core";
+import type { ImportDeclaration } from "@babel/types";
+
+export const setupTransformHook = (): VoidFunction => {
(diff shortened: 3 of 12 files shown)


src/test-reader/test-transformer.ts
```diff
@@ -0,0 +1,47 @@
...
+        browserslistConfigFile: false,
+        babelrc: false,
+        configFile: false,
+        compact: false,
+        plugins: [
+            [
+                require("@babel/plugin-transform-react-jsx"),
+                {
+                    throwIfNamespace: false,
+                    runtime: "automatic",
```
DudaGod, line 20: adds import for react even if it doesn't specified in file with component

src/test-reader/test-transformer.ts
```diff
@@ -0,0 +1,47 @@
...
+    const transformOptions: TransformOptions = {
+        browserslistConfigFile: false,
+        babelrc: false,
+        configFile: false,
+        compact: false,
+        plugins: [
+            [
+                require("@babel/plugin-transform-react-jsx"),
+                {
+                    throwIfNamespace: false,
```
DudaGod, line 19: https://babeljs.io/docs/babel-preset-react#throwifnamespace

src/test-reader/test-transformer.ts
```diff
@@ -0,0 +1,47 @@
...
+                    runtime: "automatic",
+                },
+            ],
+            require("@babel/plugin-transform-modules-commonjs"),
```
DudaGod, line 23: in order to modify imports to requires

src/test-reader/test-transformer.ts
```diff
@@ -0,0 +1,47 @@
...
+            [
+                (): PluginObj => ({
+                    name: "remove-css-imports",
+                    visitor: {
+                        ImportDeclaration(path: NodePath<ImportDeclaration>): void {
+                            if (path.node.source.value.match(/\.(css|less|scss)$/)) {
+                                path.remove();
```
DudaGod, line 30: In order to remove css imports. I don't find solution how to transpile them, so I remove them because we need only read tests and not execute in nodejs (will be executed in browser in next PR)

package.json
```diff
@@ -8,7 +8,8 @@
     "typings"
   ],
   "scripts": {
-    "build": "tsc && npm run copy-static",
+    "build": "tsc && npm run copy-static && npm run build-bundle -- --minify",
+    "build-bundle": "esbuild ./src/bundle/index.ts --outdir=./build/src/bundle --bundle --format=cjs --platform=node --target=ES2021",
```
DudaGod, line 12: Here I build a bundle in order to not install all heavy babel deps in `dependencies`

package.json
```diff
@@ -107,6 +115,7 @@
     "jsdom": "^24.0.0",
     "jsdom-global": "3.0.2",
     "onchange": "7.1.0",
+    "pirates": "4.0.6",
```
DudaGod, line 118: all deps installed in dev deps and used only through bundle

src/bundle/constants.ts
```diff
@@ -0,0 +1,2 @@
+export const TRANSFORM_EXTENSIONS = [".jsx", ".cjsx", ".mjsx", ".tsx", ".ctsx", ".mtsx"];
```
DudaGod, line 1: extensions which should be transformed. I took various types of dependencies in which components can be described from this site - https://www.totaltypescript.com/concepts/mjs-cjs-mts-and-cts-extensions.

src/bundle/constants.ts
```diff
@@ -0,0 +1,2 @@
...
+export const JS_EXTENSION_RE = /^\.([cm]?[tj]sx?|json)$/;
```
DudaGod, line 2: used in order to transform only javascript files and json. Omit all other extensions like `.png`, `.woff`, `.css` and etc.

src/bundle/test-transformer.ts
```diff
@@ -0,0 +1,49 @@
...
+            require("@babel/plugin-transform-modules-commonjs"),
+            [
+                (): PluginObj => ({
+                    name: "ignore-imports",
+                    visitor: {
+                        ImportDeclaration(path: NodePath<ImportDeclaration>): void {
+                            const extname = nodePath.extname(path.node.source.value);
+
+                            if (extname && !extname.match(JS_EXTENSION_RE)) {
+                                path.remove();
```
DudaGod, line 32: Here omit imports like `.css` and others which can be transformed

tsconfig.json
```diff
@@ -1,7 +1,7 @@
 {
     "extends": "./tsconfig.common.json",
     "include": ["src"],
-    "exclude": ["src/**/__*", "src/**/*.test.ts", "src/browser/client-scripts"],
+    "exclude": ["src/**/__*", "src/**/*.test.ts", "src/browser/client-scripts", "src/bundle"],
```
DudaGod, line 4: `src/bundle` should not be build by typescript. I build it using esbuild

src/test-reader/test-transformer.ts
```diff
@@ -0,0 +1,3 @@
+/* eslint-disable @typescript-eslint/no-var-requires */
+export const setupTransformHook: () => VoidFunction = require("../bundle").setupTransformHook;
+export const TRANSFORM_EXTENSIONS: string[] = require("../bundle").TRANSFORM_EXTENSIONS;
```
DudaGod, line 3: Here I use `require` in order to ts not understand that folder `bundle` should be builded. If I use `import` tsc build it even if exclude it in tsconfig

package.json
shadowusr: I think we should add it to the `watch` script as well.

In watch script, we can either build the bundle once on startup, or make 2 separate scripts like `watch:bundle` `watch:src` and unite them in `watch` script via `concurrently` (we use the [same approach](https://github.com/gemini-testing/html-reporter/blob/9d860fef7b455d9b63b4fcbfc91ad5ed38754eae/package.json#L34) in html-reporter).

I'm not sure how often we'll want to change (watch) bundle contents, so it's up to you to decide.

src/test-reader/test-transformer.ts
shadowusr: I suggest adding this comment to the code as well.

To me it seems like typescript's [project references](https://www.typescriptlang.org/docs/handbook/project-references.html) would allow the use of imports here, but don't include those files in build. But probably not worth spending time on this right now, maybe just leave a todo in the comment.

shadowusr (approved):
Great additions overall! =)

PR #856, opened 2024-03-06
fix: correctly screen elements inside iframes

DudaGod:
## What is done

`getBoundingClientRect` inside iframe returns DOMRect only inside this iframe and don't know nothing about offsets between this iframe and viewport. Therefore, a solution has been implemented that takes into account the indentation of the iframe

Fix issue - https://github.com/gemini-testing/hermione/issues/349

### DIFF
--- a/src/browser/client-scripts/index.js
+++ b/src/browser/client-scripts/index.js
@@ -70,10 +70,11 @@ function prepareScreenshotUnsafe(areas, opts) {
         }
     }
 
-    var viewportWidth = document.documentElement.clientWidth,
-        viewportHeight = document.documentElement.clientHeight,
-        documentWidth = document.documentElement.scrollWidth,
-        documentHeight = document.documentElement.scrollHeight,
+    var mainDocumentElem = util.getMainDocumentElem(),
+        viewportWidth = mainDocumentElem.clientWidth,
+        viewportHeight = mainDocumentElem.clientHeight,
+        documentWidth = mainDocumentElem.scrollWidth,
+        documentHeight = mainDocumentElem.scrollHeight,
         viewPort = new Rect({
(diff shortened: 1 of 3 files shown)


src/browser/client-scripts/index.js
```diff
@@ -70,10 +70,11 @@ function prepareScreenshotUnsafe(areas, opts) {
         }
     }
 
-    var viewportWidth = document.documentElement.clientWidth,
-        viewportHeight = document.documentElement.clientHeight,
-        documentWidth = document.documentElement.scrollWidth,
-        documentHeight = document.documentElement.scrollHeight,
+    var mainDocumentElem = util.getMainDocumentElem(),
```
DudaGod, line 73: If user use command `browser.switchToFrame(<iframe>)` in test then `document.documentElement` references to the document inside frame which has its own dimensions. It can't be used because `getBoundingClientRect` inside iframe returns rect only inside this iframe.

So I should use main document in order to correctly screen element inside iframe.

src/browser/client-scripts/rect.js
```diff
@@ -164,3 +164,76 @@ exports.getAbsoluteClientRect = function getAbsoluteClientRect(element, opts) {
 
     return clientRect.translate(util.getScrollLeft(opts.scrollElem), util.getScrollTop(opts.scrollElem));
 };
+
+function getNestedBoundingClientRect(node, boundaryWindow) {
```
DudaGod, line 168: Found solution in this package - https://github.com/suchipi/get-nested-bounding-client-rect/tree/master. The basic logic is taken from it with some modification for our use case.

KuznetsovRoman (approved):
LGTM

PR #863, opened 2024-03-07
feat: add ability to use import and export in test files

KuznetsovRoman:
## What is done
- Add ability to use "import" and "export" statements in tests

https://www.typescriptlang.org/tsconfig/#allowJs

### DIFF
--- a/src/utils/typescript.ts
+++ b/src/utils/typescript.ts
@@ -27,6 +27,9 @@ export const tryToRegisterTsNode = (): void => {
                 skipProject: JSON.parse(skipProjectRaw),
                 transpileOnly: JSON.parse(transpileOnlyRaw),
                 swc: JSON.parse(swcRaw),
+                compilerOptions: {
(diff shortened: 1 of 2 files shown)


src/utils/typescript.ts
```diff
@@ -27,6 +27,9 @@ export const tryToRegisterTsNode = (): void => {
                 skipProject: JSON.parse(skipProjectRaw),
                 transpileOnly: JSON.parse(transpileOnlyRaw),
                 swc: JSON.parse(swcRaw),
+                compilerOptions: {
+                    allowJs: true,
+                },
```
DudaGod, line 32: It gives ability to use `import/export` inside tests on javascript?

DudaGod (approved):

src/utils/typescript.ts
KuznetsovRoman: If you use ts-node, yes

PR #866, opened 2024-03-07
fix: add getConfig command typings

KuznetsovRoman:
## What is done
- Add `browser.getConfig` typings

### DIFF
--- a/src/browser/types.ts
+++ b/src/browser/types.ts
@@ -26,6 +26,8 @@ declare global {
 
             extendOptions(opts: { [name: string]: unknown }): Promise<void>;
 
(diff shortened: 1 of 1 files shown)


src/browser/types.ts
```diff
@@ -26,6 +26,8 @@ declare global {
 
             extendOptions(opts: { [name: string]: unknown }): Promise<void>;
 
+            getConfig(): Promise<BrowserConfig>;
```
DudaGod, line 29: It looks like as `fix`, because it should have been described when write this file

DudaGod (approved):

PR #869, opened 2024-03-11
fix: add typings for overwrireCommand: custom commands

KuznetsovRoman:


src/browser/types.ts
```diff
@@ -18,6 +18,12 @@ export interface Browser {
 declare global {
     // eslint-disable-next-line @typescript-eslint/no-namespace
     namespace WebdriverIO {
+        type OverwriteCommandFn<IsElement extends boolean = false> = (
+            this: IsElement extends true ? WebdriverIO.Element : WebdriverIO.Browser,
+            origCommand: (...args: any[]) => Promise<any>, // eslint-disable-line @typescript-eslint/no-explicit-any
+            ...args: any[] // eslint-disable-line @typescript-eslint/no-explicit-any
+        ) => Promise<any>; // eslint-disable-line @typescript-eslint/no-explicit-any
```
KuznetsovRoman, line 25: Took the types from webdriverio.
It uses `any` and prettier spreads this into multiple lines, so i have to manually disable this eslint rule for each line of typings

DudaGod: I suggest to send PR/issue to wdio with this problem. It looks weird that we should modify type in order to be able to use custom commands.

DudaGod (approved):

src/browser/types.ts
KuznetsovRoman: https://github.com/webdriverio/webdriverio/issues/12485

PR #870, opened 2024-03-11
fix: add ability to specify runStep resolved value

KuznetsovRoman:


src/browser/types.ts
```diff
@@ -76,9 +76,9 @@ declare global {
              *
              * @param stepName step name
              * @param stepCb step callback
-             * @returns {Promise<any>} value, returned by `stepCb`
+             * @returns {Promise<T>} value, returned by `stepCb`
              */
-            runStep(stepName: string, stepCb: () => Promise<unknown> | unknown): Promise<unknown>;
+            runStep<T = unknown>(stepName: string, stepCb: () => Promise<T> | T): Promise<T>;
```
KuznetsovRoman, line 81: This type does not need to be specified manually. It would be inferred from stepCb return value

DudaGod (approved):

PR #876, opened 2024-03-19
fix: add missing command types (moveCursorTo and openAndWait)

shadowusr:
This PR adds missing types for `moveCursorTo` and `openAndWait` commands and changes command exports from `export =` to `export default`.


DudaGod (approved):
🔥 

Looks like you forgot to run prettier.

PR #880, opened 2024-03-21
feat: ability to run unit tests in browser

DudaGod:
## What is done

The ability to run tests in a browser environment is implemented (previously, it was possible to run only in nodejs env). Now you can run the unit tests in the browser. The launch of the component tests will be implemented in the next PR.

To run browser tests, you must specify `system.testRunEnv: "browser"`. And to use your vite config - `system.testRunEnv: ["browser", {viteConfig: {...}}]`. Browser tests are run separately from tests in node js.

Implementation scheme:
- [light](https://github.com/gemini-testing/hermione/assets/5900697/f65284b3-aacb-416c-b674-de6bebd1c0f3)
- [dark](https://github.com/gemini-testing/hermione/assets/5900697/0076a479-e709-40f2-89c8-0d94ff6f87e7)

[Vite](https://vitejs.dev/) is used as the server that compiles the test code and displays it in the browser. All browser tests go to the url of the vite server. Vite also has presets with support popular frameworks: react, vue, etc.

The logic about launching browsers, managing them, and reading tests in the master and the worker process has not changed. Only a new browser runner has been added, which start the Vite server and sets up communication between workers and browsers via websocket. It also generates the necessary index.html page in memory that the browser opens. On this page, mocha is connected, which downloads passed test file, parses all the tests and saves them in its data structure. The worker is responsible for starting the launch of hooks (`beforEach`, `afterEach`) and tests (`it`). Which sends a vite message via the websocket which test needs to be run. fullTitle is used to identify which runnable must be run. The browser launches the required runnable and returns the result. If there were errors, the test fails, and if everything is OK, the test is successful.

---

User documentation will be added along with the implementation of component testing (next PR).


src/worker/browser-env/runner/test-runner/execution-thread.ts
```diff
@@ -0,0 +1,49 @@
...
+
+import { WorkerEventNames } from "../../types";
+
+import type { ViteWorkerCommunicator } from "../../communicator";
+import type { BrowserEnvRuntimeConfig } from "../../types";
+import type { ExecutionThreadCtorOpts } from "../../../runner/test-runner/types";
+import type { Test } from "../../../../test-reader/test-object/test";
+import type { Hook } from "../../../../test-reader/test-object/hook";
+
+export const wrapExecutionThread = (runUuid: string): typeof NodejsEnvExecutionThread => {
```
DudaGod, line 14: Solution which gives me ability to pass `runUuid` and generate `ExecutionThread` class for browser. I use it because `ExecutionThread` class is used inside `NodejsEnvTestRunner` and I need some api to set it, but `ExecutionThread` from nodejs don't need this.

src/worker/browser-env/runner/test-runner/execution-thread.ts
```diff
@@ -0,0 +1,49 @@
...
+        #runUuid: string = runUuid;
+
+        constructor(opts: ExecutionThreadCtorOpts & { runUuid: string }) {
+            super(opts);
+
+            this.#communicator = (RuntimeConfig.getInstance() as BrowserEnvRuntimeConfig).viteWorkerCommunicator;
+        }
+
+        async _call(runnable: Test | Hook): Promise<void> {
+            const cmdUuid = crypto.randomUUID();
```
DudaGod, line 26: generates uniq for each event

src/worker/browser-env/runner/test-runner/index.ts
```diff
@@ -0,0 +1,58 @@
...
+        super(opts);
+
+        this.#communicator = (RuntimeConfig.getInstance() as BrowserEnvRuntimeConfig).viteWorkerCommunicator;
+    }
+
+    async run(opts: WorkerTestRunnerRunOpts): Promise<WorkerRunTestResult> {
+        return super.run({ ...opts, ExecutionThreadCls: wrapExecutionThread(this.#runUuid) });
+    }
+
+    _getPreparePageActions(browser: Browser, history: BrowserHistory): (() => Promise<void>)[] {
```
DudaGod, line 29: redefine method from parent in order to be able run action before user hook/test

src/worker/runner/test-runner/index.js
```diff
@@ -117,6 +118,19 @@ module.exports = class TestRunner {
         return results;
     }
 
+    _getPreparePageActions(browser, history) {
```
DudaGod, line 121: Add API to be able to redefine it through browser runner

src/worker/browser-env/runner/test-runner/index.ts
```diff
@@ -0,0 +1,59 @@
...
+        return [
+            async (): Promise<void> => {
+                await history.runGroup(browser.callstackHistory, "openVite", async () => {
```
KuznetsovRoman, line 32: > history.runGroup(browser.callstackHistory

Why not

`browser.runStep("openVite", async () => {...)`

package.json
```diff
@@ -62,7 +62,9 @@
     "fastq": "1.13.0",
     "fs-extra": "5.0.0",
     "gemini-configparser": "1.3.0",
+    "get-port": "5.1.1",
     "glob-extra": "5.0.2",
+    "import-meta-resolve": "4.0.0",
```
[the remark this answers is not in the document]
shadowusr: 🐴fill =)

src/runner/browser-env/vite/plugins/resolve-module-paths.ts
```diff
@@ -0,0 +1,36 @@
...
+                if (id.endsWith(MODULE_NAMES.browserRunner)) {
+                    return browserRunnerModulePath;
+                }
+
+                if (id.endsWith(MODULE_NAMES.globals)) {
+                    return globalsModulePath;
+                }
+
+                if (id.endsWith(MODULE_NAMES.mocha)) {
+                    return options.modulePaths.mocha;
```
shadowusr, line 31: hm, isn't it a bit weird that some resolutions are hard-coded here, but some are passed as options, but their types are still hard-coded? Maybe we should stick to the single approach (include all resolutions here or move them all to options)?

Also, thinking out loud, do you think it's worth it to have a separate plugin for such resolutions? An alternative would be to just write complete paths at the generate-index-html helper, which sounds easier? But I may be missing something.

Not critical for me. Leaving this up to you.

src/runner/browser-env/vite/plugins/generate-index-html.ts
```diff
@@ -0,0 +1,73 @@
...
+    const file = urlParamString.get("file");
+    const runUuid = urlParamString.get("runUuid");
+    const cmdUuid = urlParamString.get("cmdUuid");
+
+    return `
+<!DOCTYPE html>
+<html>
+    <head>
+        <title>Hermione Browser Test</title>
+        <script type="module" src="${MODULE_NAMES.globals}" pid="${pid}" file="${file}" runUuid="${runUuid}" cmdUuid="${cmdUuid}"></script>
```
shadowusr, line 55: Hmm how do you use these inside globals module? I see you still refer to window.location there. Or does the module has its own window.location, and you pass these params to that context?

shadowusr (approved):
Overall, really cool addition! I have no critical review issues.

Left a few questions and comments below, also discussed  if it's worth it to consider using socket.io and separate WS server in the future. For now, I don't think any severe changes are necessary and it's surely better to get this shipped faster.

src/runner/browser-env/vite/plugins/resolve-module-paths.ts
DudaGod: I agree it wasn't the best decision. Fixed.

PR #885, opened 2024-03-26
fix: type-safe overwriteCommand

KuznetsovRoman:


shadowusr (approved):
I'm ok with this solution as is. 

---

Thinking out loud, this code shouldn't be at hermione at all. Also, this gives user an illusion of being type-safe, because you can overwrite non-existent commands, which are methods of browser object. When there's just `string`, you essentially let users know that it's their responsibility to check command name.

Also, I'd have a look at the `SetParameterType` helper in `type-fest`, looks like it does exactly what we need — changes one parameter type, while keeping all other things intact.

KuznetsovRoman:
> type-fest

While we have it in devDeps only, it won't work unless user has it.

PR #889, opened 2024-03-28
fix: return origCommand type to overwriteCommand

KuznetsovRoman:
## What is done
- add absent `origCommand` parameter to OverwriteCommandFn

### DIFF
--- a/src/browser/types.ts
+++ b/src/browser/types.ts
@@ -33,6 +33,12 @@ declare global {
     namespace WebdriverIO {
         type BrowserCommand = Exclude<FunctionProperties<WebdriverIO.Browser>, EventEmitterMethod>;
         type ElementCommand = Exclude<FunctionProperties<WebdriverIO.Element>, EventEmitterMethod>;
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        type OverwriteCommandFn<ThisArg, Fn extends (...args: any) => any> = (
+            this: ThisArg,
+            origCommand: OmitThisParameter<Fn>,
+            ...args: Parameters<Fn>
+        ) => ReturnType<Fn>;
 
         interface Browser {
             getMeta(this: WebdriverIO.Browser): Promise<BrowserMeta>;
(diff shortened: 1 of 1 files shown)


src/browser/types.ts
```diff
@@ -33,6 +33,12 @@ declare global {
     namespace WebdriverIO {
         type BrowserCommand = Exclude<FunctionProperties<WebdriverIO.Browser>, EventEmitterMethod>;
         type ElementCommand = Exclude<FunctionProperties<WebdriverIO.Element>, EventEmitterMethod>;
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        type OverwriteCommandFn<ThisArg, Fn extends (...args: any) => any> = (
+            this: ThisArg,
+            origCommand: OmitThisParameter<Fn>,
```
KuznetsovRoman, line 39: Removing `this` parameter from the function we are overwriting, so we can call it without `bind`/`call`/`appy` functions:

```ts
browser.overwriteCommand('url', function (origCommand, path) {
    return origCommand(path); // no `call`/`bind`/`apply` is needed
})
```

src/browser/types.ts
```diff
@@ -33,6 +33,12 @@ declare global {
     namespace WebdriverIO {
         type BrowserCommand = Exclude<FunctionProperties<WebdriverIO.Browser>, EventEmitterMethod>;
         type ElementCommand = Exclude<FunctionProperties<WebdriverIO.Element>, EventEmitterMethod>;
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        type OverwriteCommandFn<ThisArg, Fn extends (...args: any) => any> = (
+            this: ThisArg,
```
KuznetsovRoman, line 38: Adding `this` parameter to the `OverwriteCommandFn`, so we can use `this` (usefull for elements):

```ts
browser.overwriteCommand('click', function (origCommand) {
    this // WebdriverIO.Element

    return origCommand(path);
})
```

KuznetsovRoman: At the time, we can use arrow function here

src/browser/types.ts
```diff
@@ -33,6 +33,12 @@ declare global {
...
+            origCommand: OmitThisParameter<Fn>,
+            ...args: Parameters<Fn>
+        ) => ReturnType<Fn>;
```
KuznetsovRoman, line 41: > 38

Adding `this` parameter to the `OverwriteCommandFn`, so we can use `this` (which is usefull for elements):

```ts
browser.overwriteCommand('click', function (origCommand) {
    this // WebdriverIO.Element

    return origCommand(path);
})
```

At the time, we dont have to

> 39

Removing `this` parameter from the function we are overwriting, so we can call it without `bind`/`call`/`appy` functions:

```ts
browser.overwriteCommand('url', function (origCommand, path) {
    return origCommand(path); // no `call`/`bind`/`apply` is needed
})
```

shadowusr (approved):

PR #892, opened 2024-03-31
feat: new ./bin/testplane

miripiruni:


package.json
```diff
@@ -39,7 +39,8 @@
     "test"
   ],
   "bin": {
-    "hermione": "./bin/hermione"
+    "hermione": "./bin/hermione",
+    "testplane": "./bin/hermione"
```
KuznetsovRoman, line 43: You still have one binary file, so `npx testplane` will launch `./bin/hermione`, and `{cliName: 'hermione'}` would be passed to `run` method either way.

test/src/cli/index.js
```diff
@@ -38,11 +38,27 @@ describe("cli", () => {
...
+        });
+
+        it("should show information about testplane by default", async () => {
+            const defaultResult = info.configOverriding();
+
+            assert.isTrue(defaultResult.includes("testplane"));
+            assert.isFalse(defaultResult.includes("hermione"));
+        });
+
+        it("should show information about hermione by default", async () => {
```
KuznetsovRoman, line 56: > by default

Does not look like `default`, if you pass `{ cliName: "hermione" }`

src/cli/info.js
```diff
@@ -1,20 +1,24 @@
 "use strict";
 
-exports.configOverriding = `  Overriding config
+exports.configOverriding = (opts = {}) => {
```
KuznetsovRoman, line 3: You lost two extra spaces at the beginning.

Here we could do this:
```js
exports.configOverriding = ({ cliName = "testplane" } = {}) => `  Overriding config
    To override any config...
```

This way we won't need to have extra `return`, and spacing would be untouched

src/cli/index.js
```diff
@@ -39,16 +39,16 @@ process.on("unhandledRejection", (reason, p) => {
...
+    const program = new Command(opts.cliName || "testplane");
 
     program.version(pkg.version).allowUnknownOption().option("-c, --config <path>", "path to configuration file");
 
     const configPath = preparseOption(program, "config");
     hermione = Hermione.create(configPath);
 
     program
-        .on("--help", () => logger.log(info.configOverriding))
+        .on("--help", () => logger.log(info.configOverriding(opts)))
```
KuznetsovRoman, line 51: name `configOverriding` looks like it is a string.

If you change its type to `Function`, i suggest to rename it to something like `getConfigOverriding`, to emphasize it is a function

src/cli/index.js
```diff
@@ -39,16 +39,16 @@ process.on("unhandledRejection", (reason, p) => {
     }
 });
 
-exports.run = () => {
-    const program = new Command();
+exports.run = (opts = {}) => {
+    const program = new Command(opts.cliName || "testplane");
```
KuznetsovRoman, line 43: Why do we need this?

test/src/cli/index.js
```diff
@@ -38,11 +38,27 @@ describe("cli", () => {
...
+            const defaultResult = info.configOverriding({ cliName: "hermione" });
```
KuznetsovRoman, line 57: So it is not a `defaultResult` eigher

DudaGod (commented):
Branch `rebranding` should be add to the protection rules. Moreover you should enable tests in this branch inside `.github/workflows/node.js.yml` (temporarily)

package.json
miripiruni: Fixed: `"testplane": "./bin/tesplane"`.

src/cli/index.js
miripiruni: My changes is optimized for future situation: just get rid of old staff.

I'm looking forward to delete opts here and in similar places.

src/cli/info.js
miripiruni: > You lost two extra spaces at the beginning.

thank you!

Fixed.

test/src/cli/index.js
miripiruni: Fixed

miripiruni:
@DudaGod setup CI for new branch — done. Thanks.

@KuznetsovRoman PR updated. "Resolve conversation" buttons is up to you.

package.json
```diff
@@ -39,7 +39,8 @@
     "test"
   ],
   "bin": {
-    "hermione": "./bin/hermione"
+    "hermione": "./bin/hermione",
+    "testplane": "./bin/testplane"
```
KuznetsovRoman, line 43: `./bin/testplane` is missing

.github/workflows/node.js.yml
```diff
@@ -5,9 +5,9 @@ name: Node.js CI
 
 on:
     push:
-        branches: [master]
+        branches: [master, rebranding]
     pull_request:
-        branches: [master]
+        branches: [master, rebranding]
```
KuznetsovRoman, line 10: Your branch is called `gemini-testing:rebranding`, so the auto-checks are still not launching

src/cli/index.js
KuznetsovRoman: And still, why do we need this? What for?

test/src/cli/index.js
KuznetsovRoman: name `defaultResult` is a lie.

KuznetsovRoman (changes requested):

src/cli/index.js
miripiruni: [src/cli/index.js](https://github.com/gemini-testing/hermione/pull/892/commits/d062638402a43920027c0b0c89eeb81fe9584167#diff-f00e7c78fa83a1316c2844a481cc2eeda7c7c2c4f958d39e9e457ddbc6b5c20f) used twice:

In bin/hermione and bin/testplane.

In future we will delete ./bin/hermione and opts in src/cli/index.js::run() also.

Right now, I introduced opts to pass cliName to compose proper help in CLI.

I decided to use opts as object with cliName field to make in more readable in on the call side.

Compate variants:

```
.run('hermione');
.run({ cliName: 'hermione' });
```

…for me second is much better.

KuznetsovRoman: I mean, why do we need to pass `"hermione" / "testplane"` into Command constructor?

PR #895, opened 2024-04-02
chore: replace hermione tokens

KuznetsovRoman:
## What is done
Changed `hermione` to `testplane` in all necessary places.

Public api change list:
- Instead of `HERMIONE_SETS` and `HERMIONE_SKIP_BROWSERS` env vars, you can use `TESTCOP_` analogs
- `screenshotsDir` default value changed. But if default value is used and there is `"hermione/screens"` directory, but no `"testplane/screens"` (new default) directory, deprecated path will be chosen
- marked all `hermione*` stuff as `deprecated`
- added copy `testplane` stuff


README.md
```diff
@@ -862,7 +862,7 @@ If you want to extend testplane.ctx typings, you could use module augmentation:
 import type { TestplaneCtx } from "testplane";
 
 declare module "testplane" {
-    interface HermioneCtx {
+    interface TestplaneCtx {
```
KuznetsovRoman, line 865: Changed interface name in README.
You can still use `HermioneCtx` though

package.json
```diff
@@ -1,6 +1,6 @@
 {
-  "name": "hermione",
-  "version": "8.7.2",
+  "name": "testplane",
+  "version": "0.0.1",
```
KuznetsovRoman, line 3: first ready version will be marked as 0.1.0

src/base-testplane.ts
```diff
@@ -72,6 +69,6 @@ export abstract class BaseHermione extends AsyncEmitter {
     }
 
     protected _loadPlugins(): void {
-        pluginsLoader.load(this, this.config.plugins, PREFIX);
+        pluginsLoader.load(this, this.config.plugins, "hermione-");
```
KuznetsovRoman, line 72: Using hard-coded `"hermione-"` here, because it no longer depends on `package.json`-s name, and we are not planning to add ability to use implicit plugin names in Testplane.

src/browser/types.ts
```diff
@@ -122,9 +122,11 @@ declare global {
...
             // TODO: describe executionContext more precisely
             executionContext: (RunnerTest | RunnerHook) & {
-                hermioneCtx: {
-                    assertViewResults: Array<AssertViewResult>;
-                };
+                testplaneCtx: ExecutionThreadToolCtx;
+                /**
+                 * @deprecated Use `testplaneCtx` instead
+                 */
+                hermioneCtx: ExecutionThreadToolCtx;
```
KuznetsovRoman, line 129: This label would style every usage with strikethrough with specific label:
Screenshot 2024-04-03 at 02 34 59

src/config/browser-options.js
```diff
@@ -181,7 +182,23 @@ function buildBrowserOptions(defaultFactory, extra) {
...
+            },
+            map: (value, _, __, { isSetByUser }) => {
+                const deprecatedScreensPath = "hermione/screens";
+
+                if (!isSetByUser && fs.existsSync(deprecatedScreensPath) && !fs.existsSync(value)) {
+                    return deprecatedScreensPath;
+                }
+
+                return value;
+            },
```
KuznetsovRoman, line 200: If user uses default value, we check for "hermione/screens" folder. If it exists, and "testplane/screens" does not, switch to "hermione/screens"

Only works when default value is untouched, and new default directory does not exist, but old one - does

src/config/defaults.js
```diff
@@ -9,7 +9,7 @@ module.exports = {
     desiredCapabilities: null,
     automationProtocol: WEBDRIVER_PROTOCOL,
     sessionEnvFlags: {},
-    screenshotsDir: "hermione/screens",
+    screenshotsDir: "testplane/screens",
```
KuznetsovRoman, line 12: not a breaking change, though

src/test-reader/build-instructions.js
```diff
@@ -64,7 +64,11 @@ function extendWithTimeout({ treeBuilder, config }) {
 }
 
 function buildGlobalSkipInstruction(config) {
-    const skipBrowsers = env.parseCommaSeparatedValue("HERMIONE_SKIP_BROWSERS");
+    const { value: skipBrowsers, key: envKey } = env.parseCommaSeparatedValue([
+        "TESTPLANE_SKIP_BROWSERS",
+        "HERMIONE_SKIP_BROWSERS",
+    ]);
```
KuznetsovRoman, line 70: In these cases we check both env variables. We need used env key back, so we can define correct skip reason

src/test-reader/index.js
```diff
@@ -23,9 +23,10 @@ module.exports = class TestReader extends EventEmitter {
         const { paths, browsers, ignore, sets, grep } = options;
 
         const { fileExtensions } = this.#config.system;
-        const setCollection = await SetsBuilder.create(this.#config.sets, { defaultDir: require("../../package").name })
+        const envSets = env.parseCommaSeparatedValue(["TESTPLANE_SETS", "HERMIONE_SETS"]).value;
+        const setCollection = await SetsBuilder.create(this.#config.sets, { defaultPaths: ["testplane", "hermione"] })
```
KuznetsovRoman, line 27: It used to be `defaultDir` with `package.json`-s name. Now it is straight up passed as an array with two places to find tests.

Not sure if anybody uses sets with unspecified `files` section though.

src/test-reader/test-parser.js
```diff
@@ -28,22 +28,25 @@ class TestParser extends EventEmitter {
...
+        const toolGlobals = {
             browser: browserVersionController.mkProvider(config.getBrowserIds(), eventBus),
             config: ConfigController.create(eventBus),
             ctx: _.clone(ctx),
             only: OnlyController.create(eventBus),
             skip: SkipController.create(eventBus),
         };
 
+        global.testplane = toolGlobals;
+        global.hermione = toolGlobals;
```
KuznetsovRoman, line 40: all public api instances of `hermione*` are also exported as copies (with Object.is equality) for `testplane*`

test/src/config/browser-options.js
```diff
@@ -528,6 +529,66 @@ describe("config browser-options", () => {
             assert.equal(config.browsers.b1.screenshotsDir, "/some/dir");
             assert.equal(config.browsers.b2.screenshotsDir, "/screens");
         });
+
+        it("should fallback default value to hermione/screens, if it exists and default dir not exists", () => {
```
KuznetsovRoman, line 533: New logic tests added in this file

test/src/utils/env.js
```diff
@@ -9,7 +9,17 @@ describe("env-utils", () => {
...
+            assert.equal(key, "foo");
+        });
+
+        it("should fallback to other env keys", () => {
+            process.env.foo = "a, b,c";
+
+            const { key, value } = env.parseCommaSeparatedValue(["bar", "foo"]);
+            assert.deepEqual(value, ["a", "b", "c"]);
+            assert.equal(key, "foo");
         });
```
KuznetsovRoman, line 23: And these ones

test/src/worker/testplane.js
```diff
@@ -77,50 +77,51 @@ describe("worker/hermione", () => {
...
             });
 
-            it("should load plugins with appropriate prefix", () => {
-                Hermione.create();
+            // testplane does not support its own plugin prefixes.
+            it("should load plugins with deprecated hermione prefix", () => {
+                Testplane.create();
 
                 assert.calledWith(pluginsLoader.load, sinon.match.any, sinon.match.any, "hermione-");
             });
```
KuznetsovRoman, line 104: Loading plugins with implicit "testplane-" or "@testplane/" would not be supported.

test/src/testplane.js
```diff
@@ -49,10 +48,12 @@ describe("hermione", () => {
...
 
         initReporters = sandbox.stub().resolves();
+        signalHandler = new AsyncEmitter();
 
-        Hermione = proxyquire("src/hermione", {
+        Testplane = proxyquire("src/testplane", {
             "./reporters": { initReporters },
-        }).Hermione;
+            "./signal-handler": signalHandler,
+        }).Testplane;
```
KuznetsovRoman, line 56: For some reason, after renaming, `signalHandler` was interfering from other files, so one tests was failing, when launched with all other test files.

package.json
miripiruni: Don't forget to change to 0.1.0 in the file

KuznetsovRoman: It will be made automatically. With `npm run release`

src/config/browser-options.js
```diff
@@ -181,7 +182,23 @@ function buildBrowserOptions(defaultFactory, extra) {
 
         prepareBrowser: options.optionalFunction("prepareBrowser"),
 
-        screenshotsDir: options.stringOrFunction("screenshotsDir"),
+        screenshotsDir: option({
+            defaultValue: defaultFactory("screenshotsDir"),
+            validate: value => {
+                if (!_.isString(value) && !_.isFunction(value)) {
```
miripiruni, line 188: I suggest not using lodash where possible.

```js
if (typeof value === "string" || typeof value === "function") {
    return;
}

throw new Error('"screenshotsDir" must be a string or function');
```

miripiruni (approved):

PR #902, opened 2024-04-13
feat: add ability to use browser in component tests

DudaGod:
## What is done

Added the ability to use a browser instance on the client side when running component or unit tests. It works due to the fact that a browser stub is created on the client side and the command call sends a ws message to the server, which executes command and returns the result.


package.json
```diff
@@ -51,8 +51,10 @@
     "@gemini-testing/commander": "2.15.3",
     "@types/mocha": "10.0.1",
     "@wdio/globals": "8.21.0",
+    "@wdio/protocols": "8.22.0",
     "@wdio/types": "8.21.0",
-    "@wdio/utils": "7.26.0",
+    "@wdio/utils": "8.35.0",
+    "@wdio/utils-cjs": "npm:@wdio/utils@7.26.0",
```
DudaGod, line 57: Currently we have `webdriverio: 8.21.0`, but we didn't update `@wdio/utils` because it is not support cjs from `@wdio/utils@8` (only esm). On client side cjs version of `@wdio/utils` didn't build correctly. So use old version for nodejs env (until we move to esm) and new version for browser env.

package.json
```diff
@@ -71,6 +73,7 @@
     "looks-same": "9.0.0",
     "micromatch": "4.0.5",
     "mocha": "10.2.0",
+    "modern-node-polyfills": "1.0.0",
```
DudaGod, line 76: polyfills for nodejs modules (used in client)

package.json
```diff
@@ -118,7 +121,7 @@
     "concurrently": "8.2.2",
     "copyfiles": "2.4.1",
     "doctoc": "2.2.0",
-    "esbuild": "0.20.2",
+    "esbuild": "0.18.20",
```
DudaGod, line 124: I have to downgrade it because `modern-node-polyfills` not support esbuild@0.18. But PR with fix is already merged. And I pinged the guys with a request to release new version

src/browser/browser.js
```diff
@@ -37,14 +37,7 @@ module.exports = class Browser {
             ...opts.state,
             isBroken: false,
         };
-    }
-
-    async attach(sessionId, sessionCaps, sessionOpts) {
```
DudaGod, line 42: Remove this method because it doesn't used anywhere

src/browser/existing-browser.js
```diff
@@ -35,6 +36,10 @@ module.exports = class ExistingBrowser extends Browser {
     async init({ sessionId, sessionCaps, sessionOpts } = {}, calibrator) {
         this._session = await this._attachSession({ sessionId, sessionCaps, sessionOpts });
 
+        if (!isRunInNodeJsEnv(this._config)) {
+            this._collectCustomCommands();
```
DudaGod, line 40: Collect custom commands only in browser environment. It is used in order to mock these methods on client side

src/runner/browser-env/vite/browser-modules/@types/vite.d.ts
```diff
@@ -0,0 +1,3 @@
+declare module "virtual:@testplane/driver" {
```
DudaGod, line 1: declare [virtual vite module](https://vitejs.dev/guide/api-plugin.html#virtual-modules-convention). Used in order to pass `automationProtocolPath` from server to client.

src/runner/browser-env/vite/browser-modules/constants.ts
```diff
@@ -12,3 +12,9 @@ export const VITE_SELECTORS = {
 
 export const BROWSER_EVENT_SUFFIX = "browser";
 export const WORKER_EVENT_SUFFIX = "worker";
+
+// TODO: use from nodejs code after migrate to esm
+export const SOCKET_MAX_TIMEOUT = 2147483647;
```
DudaGod, line 17: All emitted events with using `emitWithAck` on server side (in testplane master process) must set timeout, which cannot be specifies as zero or `Infinity`. This is max value which accepts socket.io

src/runner/browser-env/vite/browser-modules/driver.ts
```diff
@@ -0,0 +1,173 @@
+import {
```
DudaGod, line 1: Implements the interface for the browser stub. I looked at the implementation example in webdriverio - https://github.com/webdriverio/webdriverio/blob/main/packages/wdio-browser-runner/src/browser/driver.ts

src/runner/browser-env/vite/browser-modules/driver.ts
```diff
@@ -0,0 +1,173 @@
...
+    WebDriverProtocol,
+    MJsonWProtocol,
+    JsonWProtocol,
+    AppiumProtocol,
+    ChromiumProtocol,
+    SauceLabsProtocol,
+    SeleniumProtocol,
+    GeckoProtocol,
+    WebDriverBidiProtocol,
+} from "@wdio/protocols";
```
DudaGod, line 11: All available protocols used in order to collect all commands and then mock them

src/runner/browser-env/vite/browser-modules/driver.ts
```diff
@@ -0,0 +1,173 @@
...
+import { webdriverMonad, sessionEnvironmentDetector, devtoolsEnvironmentDetector } from "@wdio/utils";
+import { getEnvironmentVars } from "webdriver";
+import { SOCKET_MAX_TIMEOUT, SOCKET_TIMED_OUT_ERROR, MAX_ARGS_LENGTH } from "./constants.js";
+import { BrowserEventNames } from "./types.js";
+
+type ProtocolCommandFn = (...args: unknown[]) => Promise<unknown>;
+type PropertiesObject = Record<string, PropertyDescriptor>;
+
+const COMMANDS_HANDLED_BY_SERVER = ["debug", "saveScreenshot", "savePDF"];
```
DudaGod, line 20: Most commands are initially executed on the client. Protocol commands are called inside them, which are all mocked. These commands can't be executed on client and they are immediately sent to the server.

src/runner/browser-env/vite/browser-modules/driver.ts
```diff
@@ -0,0 +1,173 @@
...
+        commandWrapper: VoidFunction | undefined,
+    ): unknown {
+        const monad = webdriverMonad(params, modifier, getWdioPrototype(userPrototype));
+        return monad(window.__testplane__.sessionId, commandWrapper);
+    }
+}
+
+function getWdioPrototype(userPrototype: Record<string, PropertyDescriptor>): PropertiesObject {
+    for (const commandName of COMMANDS_HANDLED_BY_SERVER) {
+        delete userPrototype[commandName];
```
DudaGod, line 36: delete commands implementation because they must be mocked (inside `getMockedProtocolCommands`)

src/runner/browser-env/vite/browser-modules/driver.ts
```diff
@@ -0,0 +1,173 @@
...
+                commands.add(command);
+            }
+        }
+    }
+
+    return Array.from(commands);
+}
+
+function getMockedProtocolCommands(): PropertiesObject {
+    return [...getAllProtocolCommands(), ...COMMANDS_HANDLED_BY_SERVER, ...window.__testplane__.customCommands].reduce(
```
DudaGod, line 74: mock all custom commands by default

src/runner/browser-env/vite/browser-modules/driver.ts
```diff
@@ -0,0 +1,173 @@
...
+function mockCommand(commandName: string): ProtocolCommandFn {
+    return async (...args: unknown[]): Promise<unknown> => {
+        const { socket } = window.__testplane__;
+        const timeout = getCommandTimeout(commandName);
+
+        try {
+            // TODO: remove type casting after https://github.com/socketio/socket.io/issues/4925
+            const [error, result] = (await socket
+                .timeout(timeout)
+                .emitWithAck(BrowserEventNames.runBrowserCommand, { name: commandName, args })) as [
```
DudaGod, line 92: Send command to the server (testplane master process) which passes it to worker process and execute command.

src/runner/browser-env/vite/browser-modules/driver.ts
```diff
@@ -0,0 +1,173 @@
...
+            return prev;
+        },
+        {} as PropertiesObject,
+    );
+}
+
+function mockCommand(commandName: string): ProtocolCommandFn {
+    return async (...args: unknown[]): Promise<unknown> => {
+        const { socket } = window.__testplane__;
+        const timeout = getCommandTimeout(commandName);
```
DudaGod, line 86: Use `timeout` from testplane config

src/runner/browser-env/vite/browser-modules/driver.ts
```diff
@@ -0,0 +1,173 @@
...
+    return getEnvironmentVars(environment);
+}
+
+function getCommandTimeout(commandName: string): number {
+    const { config } = window.__testplane__;
+
+    let timeout = config.httpTimeout;
+
+    if (commandName === "debug" || commandName === "switchToRepl") {
+        timeout = SOCKET_MAX_TIMEOUT;
```
DudaGod, line 139: For commands which open repl timeout must be `Infinity`, but I can use only `2147483647` integer

src/runner/browser-env/vite/browser-modules/driver.ts
```diff
@@ -0,0 +1,173 @@
...
+    }
+
+    if (commandName === "navigateTo" && config.urlHttpTimeout) {
+        timeout = config.urlHttpTimeout;
+    }
+
+    return timeout;
+}
+
+function normalizeCommandArgs(name: string, args: unknown[] = []): unknown[] {
```
DudaGod, line 149: Get implementation from server side, but rewrite a little bit in order to not use lodash.
I can't just import it from server side, because client code builds separately and knows nothing about nodejs code

src/runner/browser-env/vite/browser-modules/globals.ts
```diff
@@ -64,9 +62,20 @@ const mockBlockingDialogs = (): void => {
     window.prompt = mockDialog({ name: "prompt", value: null });
 };
 
+const mockBuiltInNodeJsModules = (): void => {
+    window.process = window.process || {
```
DudaGod, line 66: Some inner modules inside wdio used `process`. So I need to mock it.

src/runner/browser-env/vite/browser-modules/globals.ts
```diff
@@ -12,9 +12,7 @@ const connectToSocket = (): BrowserViteSocket => {
         transports: ["websocket"],
     }) as BrowserViteSocket;
 
-    socket.on("connect_error", err => {
-        document.body.insertAdjacentHTML("afterbegin", `<p>${err.message}</p>`);
-    });
```
DudaGod, line 17: It was mistake. I shouldn't change the user's layout here. Now I write it only to the log.

src/runner/browser-env/vite/browser-modules/mocha/index.ts
```diff
@@ -62,7 +62,8 @@ export class MochaWrapper {
             let error: Error | undefined = undefined;
 
             try {
-                await (runnableToRun.fn as Mocha.AsyncFunc)?.call({} as unknown as Mocha.Context);
+                const ctx = { browser: window.browser };
+                await (runnableToRun.fn as unknown as RunnableFn).call(ctx, ctx);
```
DudaGod, line 66: Call runnable with `browser` as `this` and as first argument of runnable callback. I will add `currentTest` in one of the following PR

src/runner/browser-env/vite/socket.ts
```diff
@@ -103,4 +103,21 @@ function handleBrowserEvents(
...
+    socket.on(BrowserEventNames.runBrowserCommand, async (payload, cb) => {
+        const { runUuid } = socket.handshake.auth;
+        socket.join(runUuid);
+
+        try {
+            const [response] = await io
+                .to(runUuid)
+                .except(socket.id)
+                .timeout(SOCKET_MAX_TIMEOUT)
+                .emitWithAck(BrowserEventNames.runBrowserCommand, payload);
```
DudaGod, line 116: Pass event to worker

src/worker/browser-env/runner/test-runner/index.ts
```diff
@@ -53,6 +62,17 @@ export class TestRunner extends NodejsEnvTestRunner {
...
             async (): Promise<void> => {
+                this._socket.on(BrowserEventNames.runBrowserCommand, this._handleRunBrowserCommand(browser));
+
+                this._socket.emit(WorkerEventNames.initialize, {
+                    file: this._file,
+                    sessionId: this._runOpts.sessionId,
+                    capabilities: this._runOpts.sessionCaps,
+                    requestedCapabilities: this._runOpts.sessionOpts.capabilities,
+                    customCommands: browser.customCommands,
+                    config: this._config as BrowserConfig,
```
DudaGod, line 73: in addition to pass `file`, I now need to transfer data to properly create the browser on client

src/worker/browser-env/runner/test-runner/index.ts
```diff
@@ -61,6 +81,36 @@ export class TestRunner extends NodejsEnvTestRunner {
...
     }
 
+    private _handleRunBrowserCommand(browser: Browser): BrowserViteEvents[BrowserEventNames.runBrowserCommand] {
+        const { publicAPI: session } = browser;
+
+        return async (payload, cb): Promise<void> => {
+            const { name, args } = payload;
+            const cmdName = name as keyof typeof session;
+
+            if (typeof session[cmdName] !== "function") {
```
DudaGod, line 91: additional verification that the transmitted command exists

src/runner/browser-env/vite/socket.ts
```diff
@@ -103,4 +103,21 @@ function handleBrowserEvents(
 
         io.to(runUuid).except(socket.id).emit(BrowserEventNames.initialize, payload);
     });
+
+    socket.on(BrowserEventNames.runBrowserCommand, async (payload, cb) => {
+        const { runUuid } = socket.handshake.auth;
+        socket.join(runUuid);
```
DudaGod, line 109: TODO: there is no need to add socket in room on this event, so just remove this line.

src/runner/browser-env/vite/browser-modules/driver.ts
KuznetsovRoman: Would it work with chainings like `browser.$('.element').click()`?

I noticed you are not splitting `customCommands` for browser and elements, and just using them together, so if a user adds a custom command to the element, would he be able to use it,

src/runner/browser-env/vite/browser-modules/driver.ts
KuznetsovRoman: I would use name `SERVER_HANDLED_COMMANDS`

KuznetsovRoman: > Most commands are initially executed on the client

So you are sending requests to remote grid right from the client? Does CORS mind?

src/runner/browser-env/vite/browser-modules/driver.ts
```diff
@@ -0,0 +1,173 @@
...
+            }
+        }
+    }
+
+    return Array.from(commands);
+}
+
+function getMockedProtocolCommands(): PropertiesObject {
+    return [...getAllProtocolCommands(), ...COMMANDS_HANDLED_BY_SERVER, ...window.__testplane__.customCommands].reduce(
+        (prev, commandName) => {
```
KuznetsovRoman, line 75: nit: we usually use `acc`

src/runner/browser-env/vite/browser-modules/driver.ts
KuznetsovRoman, line 92: What is `Ack`?

src/runner/browser-env/vite/browser-modules/driver.ts
```diff
@@ -0,0 +1,173 @@
...
+                result?: unknown,
+            ];
+
+            if (error) {
+                throw error;
+            }
+
+            return result;
+        } catch (err) {
+            let error = err as Error;
```
KuznetsovRoman, line 103: Can't you just use `} catch (err: Error) {`?

src/runner/browser-env/vite/browser-modules/globals.ts
KuznetsovRoman: On the other hand, how would user know if `connect_error` occured in the CI with remote grid? It is hard to check browser console error logs after the run ends

src/runner/browser-env/vite/types.ts
```diff
@@ -14,4 +14,5 @@ export interface ViteClientEvents extends BrowserViteEvents, ViteBrowserEvents {
 // TODO: use from "./browser-modules/types" after migrate to esm
 export enum BrowserEventNames {
     initialize = `${BROWSER_EVENT_SUFFIX}:initialize`,
+    runBrowserCommand = `${BROWSER_EVENT_SUFFIX}:runBrowserCommand`,
```
KuznetsovRoman, line 17: > BROWSER_EVENT_SUFFIX

So it is actually a prefix?)

src/runner/browser-env/vite/browser-modules/index.ts
```diff
@@ -1,4 +1,12 @@
+import { remote } from "webdriverio";
+import { automationProtocolPath } from "virtual:@testplane/driver";
 import { MochaWrapper } from "./mocha/index.js";
 
+window.browser = await remote({
```
KuznetsovRoman, line 5: Why `window.browser`, and not `window.__testplane__.browser`?

If possible, i would move it under `__testplane__`, so user would not overwrite it

KuznetsovRoman (approved):
/ok

src/runner/browser-env/vite/browser-modules/driver.ts
DudaGod: > Would it work with chainings like browser.$('.element').click()?

Yes, because I mock only protocol commands and do nothing with wrappers like `browser.$` and others. Implementation of command `$` - https://github.com/webdriverio/webdriverio/blob/main/packages/webdriverio/src/commands/browser/%24.ts. And protocol commands implement inside `findElement` helper - https://github.com/webdriverio/webdriverio/blob/main/packages/webdriverio/src/commands/browser/%24.ts#L108. And at the end it returns element instanse on which another commands can be called.

> I noticed you are not splitting customCommands for browser and elements, and just using them together, so if a user adds a custom command to the element, would he be able to use it like browser.$('.element').customCommand?

It gives an error from the server that command is not exists on element. Llike this - `Can't call "custom_command" on element with selector "body", it is not a function`

src/runner/browser-env/vite/browser-modules/driver.ts
DudaGod: No. I get an error - `Catch clause variable type annotation must be 'any' or 'unknown' if specified.`

src/runner/browser-env/vite/browser-modules/globals.ts
DudaGod: > On the other hand, how would user know if `connect_error` occured in the CI with remote grid? It is hard to check browser console error logs after the run ends

Adding an element with error to the page won't help much either. In one of the next PR I will add the ability to send logs from the browser to the server. This will help you to see such errors.

TODO: add `runUuid` to the error in order to be able to understand which test fell.

src/runner/browser-env/vite/browser-modules/index.ts
DudaGod: I want the user to be able to quickly execute some command directly in devtools for debug purpose. And call command like `__testplane__.browser.getUrl()` will not be very convenient.

src/browser/browser.js
```diff
@@ -107,6 +100,14 @@ module.exports = class Browser {
         }, {});
     }
 
+    _collectCustomCommands() {
+        this._session.overwriteCommand("addCommand", (origCommand, name, wrapper, elementScope) => {
```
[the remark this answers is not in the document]
KuznetsovRoman: > Do you mean that new arguments may appear in the near future?

User might specify his own. Like we did for `click`: https://github.com/gemini-testing/testplane-safari-commands/blob/hermione/v1.0.0/lib/commands/click.js#L4

Generally it is always better to mimic the original API for wrappers.

I usually use `return origCommand` even for commands with ReturnType: `Promise<void>`, because you might want to change it in the future. For example: https://nda.ya.ru/t/GVBoakzU75StY3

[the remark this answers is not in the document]
DudaGod: > User might specify his own. Like we did for click: https://github.com/gemini-testing/testplane-safari-commands/blob/hermione/v1.0.0/lib/commands/click.js#L4

These arguments passed inside `wrapper` function so in this case everything works correctly.

It can breaks only if someone add custom arguments to the `addCommand`. So I will use `...rest` anyway

src/browser/existing-browser.js
DudaGod: This public field doesn't available for user. User can use only `publicAPI` field from browser. Which can be used as `browser` instance inside test. Or we emit it on `NEW_BROWSER` event - https://github.com/gemini-testing/testplane/blob/master/src/worker/runner/browser-pool.js#L30.

As a result, since this field doesn't public for user and it's not used in any way inside (in nodejs runner), therefore, I don't see the reason to collect them.

We can start collecting them when we really need it. What do you think?

src/runner/browser-env/vite/browser-modules/index.ts
DudaGod: Yes, this is a good solution.

> Do we use the browser from the window?

I used it inside mocha and send to runnables. But for now I use `window. __testplane__.browser` in it.

PR #904, opened 2024-04-15
feat: add ability to use expect in component tests

DudaGod:
## What is done

Added the ability to use a `expect` instance on the client side when running component or unit tests.

### DIFF
--- a/src/runner/browser-env/vite/browser-modules/constants.ts
+++ b/src/runner/browser-env/vite/browser-modules/constants.ts
@@ -18,3 +18,7 @@ export const SOCKET_MAX_TIMEOUT = 2147483647;
 export const SOCKET_TIMED_OUT_ERROR = "operation has timed out";
 
 export const MAX_ARGS_LENGTH = 50;
+
+// used from - https://github.com/jestjs/jest/blob/726ca20752e38c18e20aa21740cec7aba7891946/packages/pretty-format/src/plugins/AsymmetricMatcher.ts#L11-L14
+export const ASYMMETRIC_MATCHER =
+    typeof Symbol === "function" && Symbol.for ? Symbol.for("jest.asymmetricMatcher") : 0x13_57_a5;
--- a/src/runner/browser-env/vite/browser-modules/expect.ts
+++ b/src/runner/browser-env/vite/browser-modules/expect.ts
@@ -1,0 +1,96 @@
+import { expect, type Expect } from "expect";
+import { ASYMMETRIC_MATCHER } from "./constants.js";
+import { BrowserEventNames, type BrowserRunExpectMatcherPayload, MockMatcherFn } from "./types.js";
+
+export const initExpect = (): Expect => {
+    const mockedMatchers = mockMatchers(window.__testplane__.expectMatchers);
+    expect.extend(mockedMatchers);
+
+    return expect;
+};
+
+function mockMatchers(matcherNames: string[]): Record<string, MockMatcherFn> {
+    return matcherNames.reduce((acc, matcherName) => {
+        acc[matcherName] = mockMatcher(matcherName);
+        return acc;
+    }, {} as Record<string, MockMatcherFn>);
+}
+
+function mockMatcher(matcherName: string): MockMatcherFn {
+    const mockMatcherFn: MockMatcherFn = async function (context, ...args) {
+        if (
+            typeof args[0] === "object" &&
+            "$$typeof" in args[0] &&
+            args[0].$$typeof === ASYMMETRIC_MATCHER &&
+            args[0].asymmetricMatch
+        ) {
+            args[0] = {
+                $$typeof: args[0].toString(),
+                sample: args[0].sample,
+                inverse: args[0].inverse,
+            };
+        }
+
+        const matcherPayload: BrowserRunExpectMatcherPayload = {
+            name: matcherName,
+            scope: this,
+            args,
+        };
+
+        const isContextObject = typeof context === "object";
+
+        /**
+         * Check if context is a WebdriverIO.Element
+         */
+        if (isContextObject && "elementId" in context && typeof context.elementId === "string") {
+            matcherPayload.element = context;
+        }
+
+        /**
+         * Check if context is ChainablePromiseElement
+         */
+        if (isContextObject && "then" in context && typeof context.selector === "object") {
+            matcherPayload.element = await context;
+        }
+
+        /**
+         * Check if context is an `Element` and transform it into a WebdriverIO.Element
+         */
+        if (context instanceof Element) {
+            matcherPayload.element = await window.browser.$(context as unknown as HTMLElement);
+        } else if (isContextObject && !("sessionId" in context)) {
+            /**
+             * check if context is an object or promise and resolve it
+             * but not pass through the browser object
+             */
+            matcherPayload.context = context;
+            if ("then" in context) {
+                matcherPayload.context = await context;
+            }
+        } else if (!isContextObject) {
+            /**
+             * if context is not an object or promise, pass it through
+             */
+            matcherPayload.context = context;
+        }
+
+        /**
+         * Avoid serialization issues when sending over the element. If we create
+         * an element from an existing HTMLElement, it might have custom properties
+         * attached to it that can't be serialized.
+         */
+        if (matcherPayload.element && typeof matcherPayload.element.selector !== "string") {
+            matcherPayload.element.selector = "";
+        }
+
+        const { socket, config } = window.__testplane__;
+        // TODO: remove type casting after https://github.com/socketio/socket.io/issues/4925
+        const [{ pass, message }] = (await socket
+            .timeout(config.httpTimeout)
+            .emitWithAck(BrowserEventNames.runExpectMatcher, matcherPayload)) as [{ pass: boolean; message: string }];
+
+        return { pass, message: () => message };
+    };
+
+    return mockMatcherFn;
+}
--- a/src/runner/browser-env/vite/browser-modules/index.ts
+++ b/src/runner/browser-env/vite/browser-modules/index.ts
@@ -1,12 +1,14 @@
 import { remote } from "webdriverio";
 import { automationProtocolPath } from "virtual:@testplane/driver";
 import { MochaWrapper } from "./mocha/index.js";
+import { initExpect } from "./expect.js";
 
 window.__testplane__.browser = window.browser = await remote({
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     automationProtocol: automationProtocolPath as any,
     capabilities: window.__testplane__.capabilities,
 });
+window.expect = initExpect();
 
 const mocha = MochaWrapper.create();
 await mocha.init();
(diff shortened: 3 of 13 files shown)


typings/global.d.ts
```diff
@@ -33,3 +33,8 @@ declare namespace globalThis {
     // eslint-disable-next-line no-var
     var expect: ExpectWebdriverIO.Expect;
 }
+
+declare module "expect-webdriverio/lib/matchers" {
+    const matchers: ExpectWebdriverIO.Matchers<Promise<{ pass: boolean; message(): string }>, unknown>;
+    export default matchers;
```
KuznetsovRoman, line 39: What is this for? Would it cause issues, if user adds testing library matchers?
https://github.com/gemini-testing/testplane/discussions/887#discussioncomment-8935903

src/runner/browser-env/vite/browser-modules/constants.ts
```diff
@@ -18,3 +18,6 @@ export const SOCKET_MAX_TIMEOUT = 2147483647;
 export const SOCKET_TIMED_OUT_ERROR = "operation has timed out";
 
 export const MAX_ARGS_LENGTH = 50;
+
+export const ASYMMETRIC_MATCHER =
```
KuznetsovRoman, line 22: I suggest to add a comment here: https://github.com/jestjs/jest/blob/726ca20752e38c18e20aa21740cec7aba7891946/packages/pretty-format/src/plugins/AsymmetricMatcher.ts#L11

src/runner/browser-env/vite/browser-modules/expect.ts
```diff
@@ -0,0 +1,96 @@
...
+        const matcherPayload: BrowserRunExpectMatcherPayload = {
+            name: matcherName,
+            scope: this,
+            args,
+        };
+
+        const isContextObject = typeof context === "object";
+
+        /**
+         * Check if context is an WebdriverIO.Element
```
KuznetsovRoman, line 43: `is an` -> `is a`

src/runner/browser-env/vite/browser-modules/expect.ts
```diff
@@ -0,0 +1,96 @@
...
+
+        /**
+         * Check if context is ChainablePromiseElement
+         */
+        if (isContextObject && "then" in context && typeof context.selector === "object") {
+            matcherPayload.element = await context;
+        }
+
+        /**
+         * Check if context is a `Element` and transform it into a WebdriverIO.Element
```
KuznetsovRoman, line 57: `is a` -> `is an`

src/worker/browser-env/runner/test-runner/constants.ts
```diff
@@ -1 +1,11 @@
 export const WORKER_EVENT_PREFIX = "worker";
+
+export const SUPPORTED_ASYMMETRIC_MATCHER = {
```
KuznetsovRoman, line 3: Why do you use `enum` in `browser-env/vite/types` and in `browser-env/vite/browser-modules/types.ts`, but `export const ... = {...} as const` here?

src/worker/browser-env/runner/test-runner/index.ts
```diff
@@ -127,3 +179,26 @@ export class TestRunner extends NodejsEnvTestRunner {
         ]);
     }
 }
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+function transformExpectArgs(arg: any): unknown {
```
KuznetsovRoman, line 184: Function is called `transformExpectARGS`, but it only takes one argument

src/worker/browser-env/runner/test-runner/index.ts
```diff
@@ -127,3 +179,26 @@ export class TestRunner extends NodejsEnvTestRunner {
...
+    if (
+        typeof arg === "object" &&
+        "$$typeof" in arg &&
+        Object.keys(SUPPORTED_ASYMMETRIC_MATCHER).includes(arg.$$typeof)
```
KuznetsovRoman, line 188: If `SUPPORTED_ASYMMETRIC_MATCHER` does not includes `arg.$$typeof` (which should mean "asymmetric matcher is not supported") you are just returning the argument raw.

test/src/worker/browser-env/runner/test-runner/index.ts
```diff
@@ -427,6 +441,252 @@ describe("worker/browser-env/runner/test-runner", () => {
...
+
+            describe("should return error if", () => {
+                it("expect module is not found", done => {
+                    global.expect = undefined as unknown as ExpectWebdriverIO.Expect;
+
+                    const socket = mkSocket_() as BrowserViteSocket;
+                    socketClientStub.returns(socket);
+
+                    runWithEmitBrowserInit(socket).then(() => {
+                        socket.emit(
```
KuznetsovRoman, line 466: I would make some helper function like:

```ts
const emitAsync = (socket, ...args) => new Promise(resolve => {
    socket.emit(...args, resolve);
}):
```

Then we could use it to write usual async tests, which look a lot simpler:

```ts
it("expect module is not found", async () => {
    ...
    await runWithEmitBrowserInit(socket);

    await emitAsync(socket, BrowserEventNames.runExpectMatcher, { name: "foo", args: [], scope: {} as MatcherState });

    assert.deepEqual(response, [{ pass: false, message: "Couldn't find expect module" }]);
    ....
});
```

KuznetsovRoman (approved):
`browser-modules/expect.ts` looks unreliable. Especially the types. It needs to be double-checked with custom typings of usual wdio expect.

Other than that, /ok

src/runner/browser-env/vite/browser-modules/expect.ts
```diff
@@ -0,0 +1,96 @@
+import { expect, type Expect } from "expect";
+import { ASYMMETRIC_MATCHER } from "./constants.js";
+import { BrowserEventNames, type BrowserRunExpectMatcherPayload, MockMatcherFn } from "./types.js";
+
+export const initExpect = (): Expect => {
+    const mockedMatchers = mockMatchers(window.__testplane__.expectMatchers);
```
DudaGod, line 6: Inside `expectMatchers` stored the matchers from `expect-webdriverio` which must be run in nodejs runtime

src/runner/browser-env/vite/browser-modules/expect.ts
```diff
@@ -0,0 +1,96 @@
...
+};
+
+function mockMatchers(matcherNames: string[]): Record<string, MockMatcherFn> {
+    return matcherNames.reduce((acc, matcherName) => {
+        acc[matcherName] = mockMatcher(matcherName);
+        return acc;
+    }, {} as Record<string, MockMatcherFn>);
+}
+
+function mockMatcher(matcherName: string): MockMatcherFn {
```
DudaGod, line 19: There's a complicated logic here that I've used from wdio - https://github.com/webdriverio/webdriverio/blob/main/packages/wdio-browser-runner/src/browser/expect.ts#L33-L142

src/runner/browser-env/vite/browser-modules/index.ts
```diff
@@ -1,12 +1,14 @@
...
 import { automationProtocolPath } from "virtual:@testplane/driver";
 import { MochaWrapper } from "./mocha/index.js";
+import { initExpect } from "./expect.js";
 
 window.__testplane__.browser = window.browser = await remote({
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     automationProtocol: automationProtocolPath as any,
     capabilities: window.__testplane__.capabilities,
 });
+window.expect = initExpect();
```
DudaGod, line 11: Here I can't take this declaring to `window.__testplane__` because it is always used as a global

src/runner/browser-env/vite/browser-modules/types.ts
```diff
@@ -10,19 +12,32 @@ export type RunnableFn = (
 export enum BrowserEventNames {
     initialize = `${BROWSER_EVENT_PREFIX}:initialize`,
     runBrowserCommand = `${BROWSER_EVENT_PREFIX}:runBrowserCommand`,
+    runExpectMatcher = `${BROWSER_EVENT_PREFIX}:runExpectMatcher`,
```
DudaGod, line 15: New event which is emitted from browser and listen inside testplane worker

src/runner/browser-env/vite/browser-modules/expect.ts
```diff
@@ -0,0 +1,96 @@
+import { expect, type Expect } from "expect";
```
DudaGod, line 1: Here I use `expect` from jest which is not installed as explicit dependency (installed from `expect-webdriverio`)

src/runner/browser-env/vite/plugins/generate-index-html.ts
```diff
@@ -129,6 +129,16 @@ export const plugin = async (): Promise<Plugin[]> => {
                     return ["export default () => {};", "export const resolve = () => ''"].join("\n");
                 }
             },
+
+            transform(code, id): Rollup.TransformResult {
+                if (id.includes(".vite/deps/expect.js")) {
+                    return {
+                        code: code.replace("var fs = _interopRequireWildcard(require_graceful_fs());", "var fs = {};"),
```
DudaGod, line 136: Here is hack in order to correctly use `expect` in browser environment. I'm replacing graceful fs here

src/runner/browser-env/vite/server.ts
```diff
@@ -38,6 +38,7 @@ export class ViteServer {
             optimizeDeps: {
                 // listed deps are CJS packages and need to be compiled to ESM by Vite
                 include: [
+                    "expect",
```
DudaGod, line 41: This module is cjs only and must be compiled to esm

src/runner/browser-env/vite/socket.ts
```diff
@@ -119,4 +119,20 @@ function handleBrowserEvents(
             cb([err as Error]);
         }
     });
+
+    socket.on(BrowserEventNames.runExpectMatcher, async (payload, cb) => {
```
DudaGod, line 123: Just translate event to worker from browser and back

src/worker/browser-env/runner/test-runner/constants.ts
DudaGod: Just used it from wdio source code. To my mind there doesn't seem to be much difference between `enum` and `as const`

src/worker/browser-env/runner/test-runner/index.ts
```diff
@@ -111,6 +123,46 @@ export class TestRunner extends NodejsEnvTestRunner {
         };
     }
 
+    private _handleRunExpectMatcher(
+        browser: Browser,
+        expectMatchers: ExpectWdioMatchers,
+    ): BrowserViteEvents[BrowserEventNames.runExpectMatcher] {
+        const { publicAPI: session } = browser;
+
+        return async (payload, cb): Promise<void> => {
```
DudaGod, line 132: I saw this approach from wdio - https://github.com/webdriverio/webdriverio/blob/v8.36.0/packages/wdio-runner/src/browser.ts#L368-L412

src/worker/browser-env/runner/test-runner/index.ts
DudaGod: here all arguments are processed by this function (not only assymetric matcher). So in some cases argument must be return as is.

Found this solution in wdio - https://github.com/webdriverio/webdriverio/blob/v8.36.0/packages/wdio-runner/src/utils.ts#L200-L213

typings/global.d.ts
DudaGod: I can't import only available matchers from `expect-webdriverio`. This has been implemented only in the new version - https://github.com/webdriverio/expect-webdriverio/blob/v4.13.0/src/index.ts#L8, which support only esm modules. So this is hack to be able to use only matchers

PR #905, opened 2024-04-16
feat: add ability to show user log in component tests

DudaGod:
## What is done

Added the ability to show user log in nodejs environment when tests run in browser environment

### DIFF
--- a/src/runner/browser-env/vite/browser-modules/constants.ts
+++ b/src/runner/browser-env/vite/browser-modules/constants.ts
@@ -22,3 +22,5 @@ export const MAX_ARGS_LENGTH = 50;
 // used from - https://github.com/jestjs/jest/blob/726ca20752e38c18e20aa21740cec7aba7891946/packages/pretty-format/src/plugins/AsymmetricMatcher.ts#L11-L14
 export const ASYMMETRIC_MATCHER =
     typeof Symbol === "function" && Symbol.for ? Symbol.for("jest.asymmetricMatcher") : 0x13_57_a5;
+
+export const CONSOLE_METHODS = ["log", "info", "warn", "error", "debug", "table"] as const;
--- a/src/runner/browser-env/vite/browser-modules/mocha/index.ts
+++ b/src/runner/browser-env/vite/browser-modules/mocha/index.ts
@@ -1,4 +1,5 @@
 import { TestParser } from "./parser.js";
+import { wrapConsoleMethods } from "../utils/index.js";
 import { getErrorsOnPageLoad, getErrorsOnRunRunnable, BrowserError } from "../errors/index.js";
 import { BrowserEventNames, WorkerEventNames, type BrowserViteSocket, type RunnableFn } from "../types.js";
 
@@ -33,6 +34,8 @@ export class MochaWrapper {
         }
 
         this._socket.emit(BrowserEventNames.initialize, getErrorsOnPageLoad(error));
+
+        wrapConsoleMethods();
     }
 
     private _validate(): never | void {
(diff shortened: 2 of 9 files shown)


src/runner/browser-env/vite/browser-modules/mocha/index.ts
```diff
@@ -33,6 +34,8 @@ export class MochaWrapper {
         }
 
         this._socket.emit(BrowserEventNames.initialize, getErrorsOnPageLoad(error));
+
+        wrapConsoleMethods();
```
DudaGod, line 38: Do it before run any runnable. I do not call it earlier so that all sorts of browser system logs are not displayed

src/runner/browser-env/vite/socket.ts
```diff
@@ -104,6 +104,12 @@ function handleBrowserEvents(
         io.to(runUuid).except(socket.id).emit(BrowserEventNames.initialize, payload);
     });
 
+    socket.on(BrowserEventNames.callConsoleMethod, payload => {
+        const { runUuid } = socket.handshake.auth;
+
+        io.to(runUuid).except(socket.id).emit(BrowserEventNames.callConsoleMethod, payload);
```
DudaGod, line 110: I could call the console log right here (in master), but to my mind it's better to call it in the worker, that runs the test

src/runner/browser-env/vite/browser-modules/constants.ts
```diff
@@ -22,3 +22,5 @@ export const MAX_ARGS_LENGTH = 50;
 // used from - https://github.com/jestjs/jest/blob/726ca20752e38c18e20aa21740cec7aba7891946/packages/pretty-format/src/plugins/AsymmetricMatcher.ts#L11-L14
 export const ASYMMETRIC_MATCHER =
     typeof Symbol === "function" && Symbol.for ? Symbol.for("jest.asymmetricMatcher") : 0x13_57_a5;
+
+export const CONSOLE_METHODS = ["log", "info", "warn", "error", "debug"] as const;
```
KuznetsovRoman, line 26: What about `trace`?

`table` also could be usefull.

src/runner/browser-env/vite/browser-modules/utils/console.ts
```diff
@@ -0,0 +1,16 @@
...
+export const wrapConsoleMethods = (): void => {
+    const { socket } = window.__testplane__;
+
+    for (const method of CONSOLE_METHODS) {
+        const origCommand = console[method].bind(console);
+
+        console[method] = (...args: unknown[]): void => {
+            socket.emit(BrowserEventNames.callConsoleMethod, { method, args });
+
+            origCommand(...args);
```
KuznetsovRoman, line 13: nit: I suggest to use `return origCommand(...args)`, just in case if sombody would overwrite console method too

KuznetsovRoman (approved):
/ok

src/runner/browser-env/vite/browser-modules/constants.ts
DudaGod: > What about trace? Would it work?

It works, but stack trace is not quite helpful, because it called inside worker.

> table also could be usefull.

Ok, I add it.

src/runner/browser-env/vite/browser-modules/utils/console.ts
DudaGod: I have a synchronous event here. It means I can't return anything to browser. I don't see the point in such an implementation, since the real `console` methods is still running in the browser. Here I just show arguments in terminal.

PR #907, opened 2024-04-17
fix: correctly connect to vite server from clients (worker, browser)

DudaGod:
## What is done

If the user run browsers in the cloud, and the server is launched locally, then the user needs to launch some tunneler instance. In this case, the base url is usually substituted on tunneler url. Therefore, the worker must connect directly to the server (not use url from tunneler). And the browser should not explicitly specify the transport. Since the tunnel may not support a websocket connection.

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -5,9 +5,9 @@ name: Node.js CI
 
 on:
     push:
-        branches: [master, component_testing]
+        branches: [master]
     pull_request:
-        branches: [master, component_testing]
+        branches: [master]
 
 jobs:
     build:
--- a/src/runner/browser-env/index.ts
+++ b/src/runner/browser-env/index.ts
@@ -2,6 +2,7 @@ import { ViteServer } from "./vite/server";
 import { MainRunner as NodejsEnvRunner } from "..";
 import { TestCollection } from "../../test-collection";
 import { Config } from "../../config";
+import RuntimeConfig from "../../config/runtime-config";
 import { Interceptor } from "../../events";
 import type { Stats as RunnerStats } from "../../stats";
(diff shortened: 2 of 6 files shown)


.github/workflows/node.js.yml
```diff
@@ -5,9 +5,9 @@ name: Node.js CI
 
 on:
     push:
-        branches: [master, component_testing]
+        branches: [master]
     pull_request:
-        branches: [master, component_testing]
```
DudaGod, line 10: revert commit from master, which I merge accidentally in previous PR

src/runner/browser-env/index.ts
```diff
@@ -17,6 +18,7 @@ export class MainRunner extends NodejsEnvRunner {
     async run(testCollection: TestCollection, stats: RunnerStats): Promise<void> {
         try {
             await this._viteServer.start();
+            RuntimeConfig.getInstance().extend({ viteBaseUrl: this._viteServer.baseUrl });
```
DudaGod, line 21: Save default base url in runtime config

src/runner/browser-env/vite/browser-modules/globals.ts
```diff
@@ -9,7 +9,6 @@ const connectToSocket = (): BrowserViteSocket => {
             runUuid: window.__testplane__.runUuid,
             type: BROWSER_EVENT_PREFIX,
         },
-        transports: ["websocket"],
```
DudaGod, line 12: Do not specify transports explicitly because user tunneler may not support ws connection. More info about transfport in socketio - https://socket.io/docs/v4/engine-io-protocol/#transports

KuznetsovRoman (approved):

PR #909, opened 2024-04-17
feat: add ability to use dev server

KuznetsovRoman:
Added ability to specify dev server in config file:
Example
```js
const SERVER_PORT = 3000;
...
module.exports = {
    ...
    devServer: {
        command: "npm run server:dev",
        env: {PORT: SERVER_PORT},
        readinessProbe: {
            url: `http://localhost:${SERVER_PORT}/health`
        }
    }
}
```

Command could also be `node some/file.js` or something else

### DIFF
--- a/README.md
+++ b/README.md
@@ -1606,7 +1606,7 @@ The `browser` argument is a `WebdriverIO` session.
 Configuration data can be changed depending on extra conditions in the `prepareEnvironment` function.
 
 ### devServer
-Launch dev server on hermione init.
+Launch dev server on hermione initialize (INIT event).
 
 For example, this setup:
 ```js
--- a/README.md
+++ b/README.md
@@ -113,6 +113,7 @@ Hermione is a utility for integration testing of web pages using [WebdriverIO](h
     - [List of useful plugins](#list-of-useful-plugins)
   - [prepareBrowser](#preparebrowser)
   - [prepareEnvironment](#prepareenvironment)
+  - [devServer](#devserver)
 - [CLI](#cli)
   - [Reporters](#reporters)
   - [Require modules](#require-modules)
@@ -1603,6 +1604,40 @@ The `browser` argument is a `WebdriverIO` session.
 
 ### prepareEnvironment
 Configuration data can be changed depending on extra conditions in the `prepareEnvironment` function.
+
+### devServer
+Launch dev server on hermione init.
+
+For example, this setup:
+```js
+// .hermione.conf.js
+const SERVER_PORT = 3000;
+...
+module.exports = {
+    devServer: {
+        command: "npm run server:dev",
+        env: {PORT: SERVER_PORT},
+        readinessProbe: {
+            url: `http://localhost:${SERVER_PORT}/health`
+        }
+    }
+}
+```
+Will spawn child process "npm run server:dev", pass extra environment variable "PORT" with value "3000" and wait, until "http://localhost:3000/health" responds with 200-299 status code.
+
+Full list of parameters:
+ - command (optional) `String` – command to launch dev server. If null or not defined, dev server is disabled
+ - env (optional) `Record<string, string>` – extra environment variables to pass to child process, in addition to your `process.env` 
+ - args (optional)  `String[]` – arguments to pass to child process
+ - cwd (optional) `String` – current working directory of the child process. If not defined, hermione will try to find nearest "package.json", starting from the directory with hermione config
+ - logs (optional) `Boolean` – if enabled, shows dev server logs in the console with prefix "\[dev server\]". Enabled by default
+ - readinessProbe (optional) `Function | Object` - if function, ready check is completed when function is resolved. Receives child process object. Object by default
+   - url (optional) `String` – url to request ready check status. If not defined, ready check is disabled
+   - isReady (optional) `(fetchResponse => bool | Promise<bool>)` – predicate to check if server is ready based on `readinessProbe.url` fetch response. Returns `true` if statusCode is 2xx by default
+   - timeouts (optional) `Object` – server waiting timeouts
+     - waitServerTimeout (optional) `Number` - timeout to wait for server to be ready (ms). 60_000 by default
+     - probeRequestTimeout (optional) `Number` - one request timeout (ms), after which request will be aborted. 10_000 by default
+     - probeRequestInterval (optional) `Number` - interval between ready probe requests (ms). 1_000 by default
 
 ## CLI
 
--- a/src/config/defaults.js
+++ b/src/config/defaults.js
@@ -93,6 +93,22 @@ module.exports = {
     region: null,
     headless: null,
     isolation: null,
+    devServer: {
+        command: null,
+        cwd: null,
+        env: {},
+        args: [],
+        logs: true,
+        readinessProbe: {
+            url: null,
+            isReady: null,
+            timeouts: {
+                waitServerTimeout: 60000, // 60s
+                probeRequestTimeout: 10000, // 10s
+                probeRequestInterval: 1000, // 1s
+            },
+        },
+    },
 };
 
 module.exports.configPaths = [".hermione.conf.ts", ".hermione.conf.js"];
--- a/src/config/options.js
+++ b/src/config/options.js
@@ -116,11 +116,11 @@ const rootSection = section(
                     }
 
                     if (!_.isPlainObject(value)) {
-                        throw new Error("devServer.readinessProbe must be a function, object or null");
+                        throw new Error('"devServer.readinessProbe" must be a function, object or null');
                     }
 
                     if (!_.isUndefined(value.url) && typeof value.url !== "string" && value.url !== null) {
-                        throw new Error("devServer.readinessProbe.url must be a string or null");
+                        throw new Error('"devServer.readinessProbe.url" must be a string or null');
                     }
 
                     if (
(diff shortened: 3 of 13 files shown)


src/config/options.js
```diff
@@ -101,6 +101,61 @@ const rootSection = section(
...
+                    typeof value === "function"
+                        ? value
+                        : {
+                              ...defaults.devServer.readinessProbe,
+                              ...(value || {}),
+                              timeouts: {
+                                  ...defaults.devServer.readinessProbe.timeouts,
+                                  ...((value && value.timeouts) || {}),
+                              },
+                          },
```
KuznetsovRoman, line 156: applying deep defaults, so user can specify one timeout config and the rest would be used from default settings

src/config/defaults.js
```diff
@@ -93,6 +93,22 @@ module.exports = {
     region: null,
     headless: null,
     isolation: null,
+    devServer: {
+        command: null,
```
KuznetsovRoman, line 97: If not defined or null, dev server is disabled

src/config/defaults.js
```diff
@@ -93,6 +93,22 @@ module.exports = {
...
+        cwd: null,
+        env: {},
+        args: [],
+        logs: true,
+        readinessProbe: {
+            url: null,
```
KuznetsovRoman, line 103: if not defiend or null, ready check is disabled

src/dev-server/index.ts
```diff
@@ -0,0 +1,53 @@
...
+    if (!_.isEmpty(devServerConfig.args)) {
+        logger.log("Dev server args:", JSON.stringify(devServerConfig.args));
+    }
+
+    if (!_.isEmpty(devServerConfig.env)) {
+        logger.log("Dev server env:", JSON.stringify(devServerConfig.env, null, 4));
+    }
+
+    const devServer = spawn(devServerConfig.command, devServerConfig.args, {
+        env: { ...process.env, ...devServerConfig.env },
```
KuznetsovRoman, line 28: Applying env variables from the config on top of `process.env`

src/dev-server/index.ts
```diff
@@ -0,0 +1,53 @@
...
+        cwd: devServerConfig.cwd || findCwd(configPath),
```
KuznetsovRoman, line 29: If `cwd` is not specified, trying to find nearest package.json.

For example, if project directory looks like:
```
- configs
  - hermione
    - .hermione.conf.js
- package.json
```
it would use project directory root as cwd (so npm scripts would work and user could use relative to project root paths)

src/dev-server/index.ts
```diff
@@ -0,0 +1,53 @@
...
+                "An error occured while launching dev server",
+                `Dev server failed with code '${code}' (signal: ${signal})`,
+            ].join("\n");
+            hermione.halt(new Error(errorMessage), 5000);
+        }
+    });
+
+    process.once("exit", () => {
+        devServer.kill("SIGINT");
+    });
```
KuznetsovRoman, line 50: This would kill dev server when main process dies, so `html-reporter` gui would not have to reinit dev server on each test suite run.

src/dev-server/utils.ts
```diff
@@ -0,0 +1,141 @@
...
+export const findCwd = (configPath: string): string => {
+    let prev = configPath;
+
+    // eslint-disable-next-line no-constant-condition
+    while (true) {
+        const dir = path.dirname(prev);
+
+        if (dir === prev) {
+            return path.dirname(configPath);
+        }
```
KuznetsovRoman, line 18: Fallback in case we got up to the root

src/dev-server/utils.ts
```diff
@@ -0,0 +1,141 @@
...
+
+        if (foundPackageJson) {
+            return dir;
+        }
+
+        prev = dir;
+    }
+};
+
+class WithPrefixTransformer extends Transform {
```
KuznetsovRoman, line 30: https://github.com/gemini-testing/testplane-storybook/blob/master/src/storybook/dev-server/pipe-with-prefix/with-prefix-transformer.ts

src/dev-server/utils.ts
```diff
@@ -0,0 +1,141 @@
...
+        const resultData = resultPrefix + chunkRows.join("\n" + this.prefix) + resultSuffix;
+
+        this.push(resultData);
+        this.includePrefix = includeSuffix;
+
+        callback();
+    }
+}
+
+export const pipeLogsWithPrefix = (childProcess: ChildProcessWithoutNullStreams, prefix: string): void => {
```
KuznetsovRoman, line 58: https://github.com/gemini-testing/testplane-storybook/blob/master/src/storybook/dev-server/pipe-with-prefix/index.ts#L6

src/dev-server/utils.ts
```diff
@@ -0,0 +1,141 @@
...
+        return;
+    }
+
+    logger.log("Waiting for dev server to be ready");
+
+    if (typeof readinessProbe === "function") {
+        return Promise.resolve()
+            .then(() => readinessProbe(devServer))
+            .then(() => logger.log("Dev server is ready"));
+    }
```
KuznetsovRoman, line 87: If user specified its own `readinessProbe` function, use it (with "Dev server is ready" log)

src/dev-server/utils.ts
```diff
@@ -0,0 +1,141 @@
...
+                }
+            } catch (error) {
+                const err = error as { cause?: { code?: string } };
+
+                if (!isError && !isSuccess) {
+                    setTimeout(tryToFetch, readinessProbe.timeouts.probeRequestInterval).unref();
+
+                    const errorMessage = err && err.cause && (err.cause.code || err.cause);
+
+                    if (errorMessage && errorMessage !== "ECONNREFUSED") {
```
KuznetsovRoman, line 128: `ECONNREFUSED` is thrown if server is still not ready.
If it is no `ECONNREFUSED`, it could be usefull

src/hermione.ts
```diff
@@ -60,6 +61,16 @@ export class Hermione extends BaseHermione {
         this.emit(MasterEvents.CLI, parser);
     }
 
+    protected async _init(): Promise<void> {
+        await initDevServer({
```
KuznetsovRoman, line 65: Initing dev server before plugins, so plugins could fetch dev server

src/config/options.js
```diff
@@ -101,6 +101,61 @@ const rootSection = section(
...
+            logs: options.optionalBoolean("devServer.logs"),
+            readinessProbe: option({
+                defaultValue: defaults.devServer.readinessProbe,
+                validate: value => {
+                    if (typeof value === "function" || value === null) {
+                        return;
+                    }
+
+                    if (!_.isPlainObject(value)) {
+                        throw new Error("devServer.readinessProbe must be a function, object or null");
```
DudaGod, line 119: let's wrap the variable in quotation marks as we do above in this file.

`devServer.readinessProbe` -> `"devServer.readinessProbe"`

src/dev-server/index.ts
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    logger.log("Starting dev server with command", `"${devServerConfig.command}"`);
+
+    if (!_.isEmpty(devServerConfig.args)) {
+        logger.log("Dev server args:", JSON.stringify(devServerConfig.args));
+    }
+
+    if (!_.isEmpty(devServerConfig.env)) {
+        logger.log("Dev server env:", JSON.stringify(devServerConfig.env, null, 4));
```
DudaGod, line 24: do you think all this data needs to be specified? Maybe show them by some condition? For example use `debug` library

DudaGod (approved):
🔥

README.md
```diff
@@ -1604,6 +1605,40 @@ The `browser` argument is a `WebdriverIO` session.
...
+### devServer
+Launch dev server on hermione init.
+
+For example, this setup:
+```js
+// .hermione.conf.js
+const SERVER_PORT = 3000;
+...
+module.exports = {
+    devServer: {
```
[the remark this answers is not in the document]
KuznetsovRoman: I thought about it, but currently you can't do this with our version of `configparser`: section could only be a plain object or null.

README.md
```diff
@@ -1604,6 +1605,40 @@ The `browser` argument is a `WebdriverIO` session.
...
+        readinessProbe: {
+            url: `http://localhost:${SERVER_PORT}/health`
+        }
+    }
+}
+```
+Will spawn child process "npm run server:dev", pass extra environment variable "PORT" with value "3000" and wait, until "http://localhost:3000/health" responds with 200-299 status code.
+
+Full list of parameters:
+ - command (optional) `String` – command to launch dev server. If null or not defined, dev server is disabled
```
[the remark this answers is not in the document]
KuznetsovRoman: > To my mind it must be required

It cannot be done with our current implementation of `configparser`

> And I suggest using the * symbol to list the items

I found many examples where we use `-`. Moreover, we primary use `-` for defining config options
image

README.md
```diff
@@ -1604,6 +1605,40 @@ The `browser` argument is a `WebdriverIO` session.
...
+ - env (optional) `Record<string, string>` – extra environment variables to pass to child process, in addition to your `process.env` 
+ - args (optional)  `String[]` – arguments to pass to child process
+ - cwd (optional) `String` – current working directory of the child process. If not defined, hermione will try to find nearest "package.json", starting from the directory with hermione config
+ - logs (optional) `Boolean` – if enabled, shows dev server logs in the console with prefix "\[dev server\]". Enabled by default
+ - readinessProbe (optional) `Function | Object` - if function, ready check is completed when function is resolved. Receives child process object. Object by default
```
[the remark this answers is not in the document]
KuznetsovRoman: > Moreover why you support the ability to set option as a function

`readinessProbe` function replaces our "wait for server to be ready" mechanism completely. We just use the passed function, and if it resolves, we think server is ready.

`isReady` uses our readiness probing mechanism (request for 10s with interval of 1s up to 60s timeout total), but you can specify "i think server is ready, if response looks like this". For example, you could do this: 
```ts
{
    isReady: async (response) => {
        const res = await response.json();
        return res.success == "true"
    }
}
```

src/dev-server/index.ts
KuznetsovRoman: > do you think all this data needs to be specified?

Why not? I don't think there would be a lot of options, since it only logs **extra** env variables, which are passed directly from config file

PR #912, opened 2024-04-19
fix(ct): generate url with "runUuid" as pathname

DudaGod:
## What is done

Component tests use vite, which generates for each request index.html. But in the case of parallel launch of several browsers, a case may arise when the same output is returned to both browsers, although the urls are different by query parameter.

I create an issue to vite - https://github.com/vitejs/vite/issues/16467. As a result I found that vite cached response for each request by pathname (query parameters are not involved). So for now urls modified from `http://localhost:4444/?runUuid=12345` to `http://localhost:4444/12345`. After that, the problem does not reproduce anymore.


src/runner/browser-env/vite/plugins/generate-index-html.ts
```diff
@@ -1,6 +1,7 @@
 import path from "node:path";
 import url from "node:url";
 import { builtinModules } from "node:module";
+import _ from "lodash";
```
DudaGod, line 4: This plugin executes in nodejs env so there are no problems with import all methods from module.

src/worker/browser-env/runner/test-runner/index.ts
```diff
@@ -176,7 +176,7 @@ export class TestRunner extends NodejsEnvTestRunner {
         });
 
         const timeout = this._config.urlHttpTimeout || this._config.httpTimeout;
-        const uri = new URI(this._config.baseUrl).query({ runUuid: this._runUuid }).toString();
+        const uri = urljoin(this._config.baseUrl, this._runUuid);
```
KuznetsovRoman, line 179: It would be unclear, what that number means.

Perhaps, we should do something like this:
```ts
const uri = urljoin(this._config.baseUrl, `run-uuid-${this._runUuid}`);
```
So we get something like `http://foo.bar/run-uuid-12345`

src/runner/browser-env/vite/plugins/generate-index-html.ts
```diff
@@ -59,21 +60,17 @@ export const plugin = async (): Promise<Plugin[]> => {
                         }
 
                         const urlParsed = url.parse(req.originalUrl);
-                        const urlParamString = new URLSearchParams(urlParsed.query || "");
+                        const runUuid = _.compact(urlParsed.pathname?.split("/"))[0];
```
KuznetsovRoman, line 63: And here we could look for `run-uuid-\d+` path component, instead of just taking the first one

KuznetsovRoman (approved):
I would replace `this._runUuid` in pathname with something like `run-uuid-${this._runUuid}`

src/worker/browser-env/runner/test-runner/index.ts
DudaGod: we agreed to use an URL like this `/run-uuids/:uuid` in the chat

src/bundle/test-transformer.ts
```diff
@@ -35,7 +36,6 @@ export const setupTransformHook = (): VoidFunction => {
                     },
                 }),
             ],
-            require("@babel/plugin-transform-typescript"),
```
DudaGod, line 38: In some cases `@babel/plugin-transform-typescript` works incorrectly with my tsx files. But if I use @babel/preset-typescript then everything works fine. I know that this preset just use this plugin but it looks like preset run after plugins due to the docs. So I just rewrite code to use preset instead this plugin.

PR #915, opened 2024-04-23
fix(ct): correctly support custom element commands

DudaGod:


src/browser/browser.js
```diff
@@ -101,12 +102,18 @@ module.exports = class Browser {
     }
 
     _startCollectingCustomCommands() {
-        this._session.overwriteCommand("addCommand", (origCommand, name, ...rest) => {
-            if (!this._session[name]) {
```
DudaGod, line 105: There was a mistake here. It is not possible to check the presence of a command on an element in this way. also, if the user has a custom command with the same name on the browser and the element (our case with `assertView` for example), then it was added only for the browser instance.

So I rewrite it to use command list for browser and element from history.

src/browser/browser.js
```diff
@@ -101,12 +102,18 @@ module.exports = class Browser {
...
-                this._customCommands.add(name);
+        const browserCommands = getBrowserCommands();
+        const elementCommands = getElementCommands();
+
+        this._session.overwriteCommand("addCommand", (origCommand, name, wrapper, elementScope) => {
```
DudaGod, line 108: I need `elementScope` field in order to understand how exactly I should add command in browser environment.

src/browser/history/commands.ts
```diff
@@ -23,6 +24,7 @@ const wdioBrowserCommands = [
     "mockClearAll",
     "mockRestoreAll",
     "newWindow",
+    "overwriteCommand",
```
DudaGod, line 27: these commands were skipped for some reason

src/browser/history/index.js
```diff
@@ -88,7 +88,7 @@ const overwriteBrowserCommands = (session, callstack) =>
     overwriteCommands({
         session,
         callstack,
-        commands: cmds.getBrowserCommands(),
+        commands: cmds.getBrowserCommands().filter(cmd => !shouldNotWrapCommand(cmd)),
```
DudaGod, line 91: Here I should skip `addCommand` and `overwriteCommand`

src/runner/browser-env/vite/browser-modules/driver.ts
```diff
@@ -27,7 +27,13 @@ export default class ProxyDriver {
         commandWrapper: VoidFunction | undefined,
     ): unknown {
         const monad = webdriverMonad(params, modifier, getWdioPrototype(userPrototype));
-        return monad(window.__testplane__.sessionId, commandWrapper);
+        const browser = monad(window.__testplane__.sessionId, commandWrapper);
+
+        window.__testplane__.customCommands.forEach(({ name, elementScope }) => {
```
DudaGod, line 32: All custom commands are now added using the `addCommand` call

src/runner/browser-env/vite/browser-modules/mock/@wdio-logger.ts
```diff
@@ -0,0 +1,10 @@
+export default function getLogger(): typeof console {
```
DudaGod, line 1: I found that after using `addCommand` for custom commands wdio started adding its custom logs. To disable them, I implemented this stub

src/runner/browser-env/vite/browser-modules/mock/default-module.ts
```diff
@@ -0,0 +1 @@
+export default (): void => {};
```
DudaGod, line 1: Default mock for most libraries that only use `export default`

src/runner/browser-env/vite/plugins/generate-index-html.ts
```diff
@@ -46,6 +51,15 @@ export const plugin = async (): Promise<Plugin[]> => {
...
     const automationProtocolPath = `/@fs${driverModulePath}`;
 
+    const mockDefaultModulePath = path.resolve(browserModulesPath, "mock/default-module.js");
+    const mockImportMetaResolvePath = path.resolve(browserModulesPath, "mock/import-meta-resolve.js");
+    const mockWdioLoggerPath = path.resolve(browserModulesPath, "mock/@wdio-logger.js");
+
+    const modulesToMock = DEFAULT_MODULES_TO_MOCK.reduce((acc, val) => _.set(acc, val, mockDefaultModulePath), {
+        "@wdio/logger": mockWdioLoggerPath,
+        "import-meta-resolve": mockImportMetaResolvePath,
+    }) as Record<string, string>;
```
DudaGod, line 61: here, instead of using virtual modules, I started using mock files from the file system

src/worker/browser-env/runner/test-runner/index.ts
```diff
@@ -208,3 +215,20 @@ function transformExpectArg(arg: any): unknown {
...
 }
+
+async function getWdioInstance(
+    session: WebdriverIO.Browser,
+    element?: WebdriverIO.Element,
+): Promise<WebdriverIO.Browser | WebdriverIO.Element> {
+    const wdioInstance = element ? await session.$(element) : session;
+
+    if (isWdioElement(wdioInstance) && !wdioInstance.selector) {
+        wdioInstance.selector = element?.selector as Selector;
```
DudaGod, line 226: I need recover selector because it is undefined when element found by `elementId`

src/worker/browser-env/runner/test-runner/index.ts
```diff
@@ -208,3 +215,20 @@ function transformExpectArg(arg: any): unknown {
 
     return arg;
 }
+
+async function getWdioInstance(
+    session: WebdriverIO.Browser,
+    element?: WebdriverIO.Element,
+): Promise<WebdriverIO.Browser | WebdriverIO.Element> {
+    const wdioInstance = element ? await session.$(element) : session;
```
DudaGod, line 223: Here element is found by `elementId`

test/src/worker/browser-env/runner/test-runner/index.ts
```diff
@@ -372,41 +380,87 @@ describe("worker/browser-env/runner/test-runner", () => {
         });
 
         describe(`"${BrowserEventNames.runBrowserCommand}" event`, () => {
-            describe("should return error as first callback argument if", () => {
-                it("command does not exists in browser instance", done => {
-                    const browser = mkBrowser_();
-                    (BrowserAgent.prototype.getBrowser as SinonStub).resolves(browser);
-                    const socket = mkSocket_() as BrowserViteSocket;
-                    socketClientStub.returns(socket);
+            describe("call command on element instance", () => {
```
DudaGod, line 383: I tried to combine the tests for the element and for the browser, but the code turns out to be too complicated

test/src/worker/browser-env/runner/test-runner/index.ts
```diff
@@ -430,30 +485,90 @@ describe("worker/browser-env/runner/test-runner", () => {
                 });
             });
 
-            describe("should return result as second callback argument if", () => {
-                it("command executed successfully with string", done => {
-                    const result = "some_result";
-                    const browser = mkBrowser_();
-                    browser.publicAPI.execute = sandbox.stub().resolves(result);
-                    (BrowserAgent.prototype.getBrowser as SinonStub).resolves(browser);
+            describe("call command on browser instance", () => {
```
DudaGod, line 488: Here just add one additional `describe` without changes in tests

src/runner/browser-env/vite/browser-modules/driver.ts
```diff
@@ -67,25 +73,23 @@ function getAllProtocolCommands(): string[] {
...
+        return acc;
+    }, {} as PropertiesObject);
 }
 
 function mockCommand(commandName: string): ProtocolCommandFn {
-    return async (...args: unknown[]): Promise<unknown> => {
+    return async function (this: WebdriverIO.Browser | WebdriverIO.Element, ...args: unknown[]): Promise<unknown> {
         const { socket } = window.__testplane__;
         const timeout = getCommandTimeout(commandName);
+        const element = isWdioElement(this) ? this : undefined;
```
KuznetsovRoman, line 86: Using `undefined` is considered a bad practice.

I suggest to get rid of it. For example, we could use "null" or:
```ts
const opts = { name: commandName, args };

if (isWdioElement(this)) {
    opts.element = this;
}
```

test/src/worker/browser-env/runner/test-runner/index.ts
```diff
@@ -372,41 +380,87 @@ describe("worker/browser-env/runner/test-runner", () => {
...
+                                            { stack: sinon.match(expectedErrMsg), message: expectedErrMsg },
+                                        ]);
+                                        done();
+                                    } catch (err) {
+                                        done(err);
+                                    }
+                                },
+                            );
+                        });
+                    });
```
KuznetsovRoman, line 412: It would be more readable with `await socketEmit(socket, { name: "foo", args: [], element })`, as i mentioned in the previous PR

KuznetsovRoman (approved):

src/runner/browser-env/vite/browser-modules/driver.ts
DudaGod: I don't see a real problem with using `undefined` here. It just gives me ability to write code shorter, but result the same as in your example.

test/src/worker/browser-env/runner/test-runner/index.ts
DudaGod: Yeah, I know. I will rewrite them someday.

PR #919, opened 2024-04-25
feat: add error snippets

KuznetsovRoman:
## What is done
Add error snippets

### 1
image

### 2
image

### 3
Notice how stacktrace says example.hermione.tsx?import:**12:18**, but we are pointing to line 18 because of source map evaluation
image

### DIFF
--- a/src/error-snippets/frames.ts
+++ b/src/error-snippets/frames.ts
@@ -1,0 +1,74 @@
+import _ from "lodash";
+import ErrorStackParser from "error-stack-parser";
+import logger from "../utils/logger";
+import { softFileURLToPath } from "./utils";
+import type { ResolvedFrame, SufficientStackFrame } from "./types";
+
+/**
+ * @description
+ * Rank values:
+ *
+ * 0: Can't extract code snippet; useless
+ *
+ * 1: WebdriverIO internals: Better than nothing
+ *
+ * 2: Project internals: Better than WebdriverIO internals, but worse, than user code part
+ *
+ * 3: User code: Best choice
+ */
+const FRAME_REELVANCE: Record<string, { value: number; matcher: (fileName: string) => boolean }> = {
+    repl: { value: 0, matcher: fileName => /^REPL\d+$/.test(fileName) },
+    nodeInternals: { value: 0, matcher: fileName => /^node:[a-zA-Z\-_]/.test(fileName) },
+    wdioInternals: { value: 1, matcher: fileName => fileName.includes("/node_modules/webdriverio/") },
+    projectInternals: { value: 2, matcher: fileName => fileName.includes("/node_modules/") },
+    userCode: { value: 3, matcher: () => true },
+} as const;
+
+const getFrameRelevance = (frame: StackFrame): number => {
+    if ([frame.fileName, frame.lineNumber, frame.columnNumber].some(_.isUndefined)) {
+        return 0;
+    }
+
+    const fileName: string = softFileURLToPath(frame.fileName!);
+
+    for (const factor in FRAME_REELVANCE) {
+        if (FRAME_REELVANCE[factor].matcher(fileName)) {
+            return FRAME_REELVANCE[factor].value;
+        }
+    }
+
+    return 0;
+};
+
+export const findRelevantStackFrame = (error: Error): SufficientStackFrame | null => {
+    try {
+        const parsedStackFrames = ErrorStackParser.parse(error);
+
+        let relevantFrame: SufficientStackFrame | null = null;
+        let relevantFrameRank = 0;
+
+        for (const currentFrame of parsedStackFrames) {
+            const currentFrameRank = getFrameRelevance(currentFrame);
+
+            if (currentFrameRank > relevantFrameRank) {
+                relevantFrame = currentFrame as SufficientStackFrame;
+                relevantFrameRank = currentFrameRank;
+            }
+        }
+
+        return relevantFrame;
+    } catch (findError) {
+        logger.warn("Unable to find relevant stack frame:", findError);
+
+        return null;
+    }
+};
+
+export const resolveLocationWithStackFrame = (
+    stackFrame: SufficientStackFrame,
+    fileContents: string,
+): ResolvedFrame => ({
+    file: softFileURLToPath(stackFrame.fileName),
+    source: fileContents,
+    location: { line: stackFrame.lineNumber, column: stackFrame.columnNumber },
+});
--- a/src/error-snippets/index.ts
+++ b/src/error-snippets/index.ts
@@ -1,0 +1,38 @@
+import { findRelevantStackFrame, resolveLocationWithStackFrame } from "./frames";
+import { extractSourceMaps, resolveLocationWithSourceMap } from "./source-maps";
+import { getSourceCodeFile, formatErrorSnippet } from "./utils";
+import logger from "../utils/logger";
+import type { ResolvedFrame, SufficientStackFrame, WithSnippetError } from "./types";
+
+const stackFrameLocationResolver = async (stackFrame: SufficientStackFrame): Promise<ResolvedFrame> => {
+    const fileContents = await getSourceCodeFile(stackFrame.fileName);
+    const sourceMaps = await extractSourceMaps(fileContents, stackFrame.fileName);
+
+    return sourceMaps
+        ? resolveLocationWithSourceMap(stackFrame, sourceMaps)
+        : resolveLocationWithStackFrame(stackFrame, fileContents);
+};
+
+export const extendWithCodeSnippet = async (err: WithSnippetError): Promise<WithSnippetError> => {
+    if (!err) {
+        return err;
+    }
+
+    try {
+        const relevantStackFrame = findRelevantStackFrame(err);
+
+        if (!relevantStackFrame) {
+            return err;
+        }
+
+        const { file, source, location } = await stackFrameLocationResolver(relevantStackFrame);
+
+        err.snippet = formatErrorSnippet(err, { file, source, location });
+
+        return err;
+    } catch (snippetError) {
+        logger.warn("Unable to apply code snippet:", snippetError);
+
+        return err;
+    }
+};
--- a/src/error-snippets/source-maps.ts
+++ b/src/error-snippets/source-maps.ts
@@ -1,0 +1,45 @@
+import { SourceMapConsumer, type BasicSourceMapConsumer } from "source-map";
+import url from "url";
+import { SOURCE_MAP_URL_COMMENT } from "./constants";
+import { softFileURLToPath, getSourceCodeFile } from "./utils";
+import type { SufficientStackFrame, ResolvedFrame } from "./types";
+
+export const extractSourceMaps = async (
+    fileContents: string,
+    fileName: string,
+): Promise<BasicSourceMapConsumer | null> => {
+    const sourceMapsStartIndex = fileContents.indexOf(SOURCE_MAP_URL_COMMENT);
+    const sourceMapsEndIndex = fileContents.indexOf("\n", sourceMapsStartIndex);
+
+    if (sourceMapsStartIndex === -1) {
+        return null;
+    }
+
+    const sourceMapUrl =
+        sourceMapsEndIndex === -1
+            ? fileContents.slice(sourceMapsStartIndex + SOURCE_MAP_URL_COMMENT.length)
+            : fileContents.slice(sourceMapsStartIndex + SOURCE_MAP_URL_COMMENT.length, sourceMapsEndIndex);
+
+    const sourceMaps = await getSourceCodeFile(url.resolve(fileName, sourceMapUrl));
+
+    return new SourceMapConsumer(sourceMaps) as Promise<BasicSourceMapConsumer>;
+};
+
+export const resolveLocationWithSourceMap = (
+    stackFrame: SufficientStackFrame,
+    sourceMaps: BasicSourceMapConsumer,
+): ResolvedFrame => {
+    const positions = sourceMaps.originalPositionFor({ line: stackFrame.lineNumber, column: stackFrame.columnNumber });
+    const source = positions.source ? sourceMaps.sourceContentFor(positions.source) : null;
+    const location = { line: positions.line!, column: positions.column! };
+
+    if (!source) {
+        throw new Error("File source code could not be evaluated from the source map");
+    }
+
+    if (!location.line || !location.column) {
+        throw new Error("Line and column could not be evaluated from the source map");
+    }
+
+    return { file: softFileURLToPath(sourceMaps.file), source, location };
+};
--- a/src/error-snippets/utils.ts
+++ b/src/error-snippets/utils.ts
@@ -1,0 +1,85 @@
+import path from "path";
+import { fileURLToPath } from "url";
+import fs from "fs-extra";
+import { codeFrameColumns } from "@babel/code-frame";
+import { getErrorTitle } from "../browser/stacktrace/utils";
+import { SNIPPET_LINES_ABOVE, SNIPPET_LINES_BELOW, SOURCE_MAP_URL_COMMENT } from "./constants";
+
+interface FormatFileNameHeaderOpts {
+    line: number;
+    linesAbove: number;
+    linesBelow: number;
+}
+
+interface FormatErrorSnippetOpts {
+    file: string;
+    source: string;
+    location: { line: number; column: number };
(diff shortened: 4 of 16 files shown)


src/error-snippets/frames.ts
```diff
@@ -0,0 +1,61 @@
+import _ from "lodash";
+import ErrorStackParser from "error-stack-parser";
+import logger from "../utils/logger";
+import { softFileURLToPath } from "./utils";
+import type { ResolvedFrame, SufficientStackFrame } from "./types";
+
+// prettier-ignore
+const frameRelevance: Record<string, { value: number; matcher: (fileName: string) => boolean }> = {
```
KuznetsovRoman, line 8: https://nodejs.org/api/errors.html#errorstack

> The location information will be one of:

- `native`, if the frame represents a call internal to V8 (as in [].forEach).
- `plain-filename.js:line:column`, if the frame represents a call internal to Node.js.
- `/absolute/path/to/file.js:line:column`, if the frame represents a call in a user program (using CommonJS module system), or its dependencies.
- `<transport-protocol>:///url/to/module/file.mjs:line:column`, if the frame represents a call in a user program (using ES module system), or its dependencies.

KuznetsovRoman: Ranking them as:
- 0: can't extract code snippet; useless
- 1: better than nothing
- 2: better than wdio internals, but worse, than user code part (these could be testplane parts)
- 3: user code. If frame of user code appears, always should pick it

src/error-snippets/frames.ts
```diff
@@ -0,0 +1,61 @@
...
+                frameRelevance = currentFrameRelevance;
+            }
+        }
+
+        return relevantFrame;
+    } catch (findError) {
+        logger.warn("Unable to find relevant stack frame:", findError);
+
+        return null;
+    }
```
KuznetsovRoman, line 54: If error appers, dont throw any error, just returning `null` with console warn

src/error-snippets/index.ts
```diff
@@ -0,0 +1,54 @@
...
+import logger from "../utils/logger";
+import { SNIPPET_LINES_ABOVE, SNIPPET_LINES_BELOW } from "./constants";
+import type { ResolvedFrame, SufficientStackFrame, WithSnippetError } from "./types";
+
+const stackFrameResolver = async (stackFrame: SufficientStackFrame): Promise<ResolvedFrame> => {
+    const fileContents = await getSourceCodeFile(stackFrame.fileName);
+    const sourceMaps = await extractSourceMaps(fileContents, stackFrame.fileName);
+
+    return sourceMaps ? resolveWithSourceMap(stackFrame, sourceMaps) : resolveWithStackFrame(stackFrame, fileContents);
+};
```
KuznetsovRoman, line 14: `getSourceCodeFile` extracts source code from file and applies source maps, if it is in http headers: https://docs.google.com/document/d/1U1RGAehQwRypUTovF1KRlpiOFze0b-_2gc6fAH0KY0k/edit#heading=h.lmz475t4mvbx

So in other places we dont have to think "is the file from node:fs or browser:fetch?"

KuznetsovRoman: source maps will be evaluated, if exists

src/error-snippets/index.ts
```diff
@@ -0,0 +1,54 @@
...
+
+export const extendWithCodeSnippet = async (err: WithSnippetError): Promise<WithSnippetError> => {
```
KuznetsovRoman, line 16: this function does not throw exceptions either.
The worst it can do - log a warning into console.

src/error-snippets/source-maps.ts
```diff
@@ -0,0 +1,45 @@
...
+    if (sourceMapsStartIndex === -1) {
+        return null;
+    }
+
+    const sourceMapUrl =
+        sourceMapsEndIndex === -1
+            ? fileContents.slice(sourceMapsStartIndex + SOURCE_MAP_URL_COMMENT.length)
+            : fileContents.slice(sourceMapsStartIndex + SOURCE_MAP_URL_COMMENT.length, sourceMapsEndIndex);
+
+    const sourceMaps = await getSourceCodeFile(url.resolve(fileName, sourceMapUrl));
```
KuznetsovRoman, line 23: `sourceMapUrl` could either be relative, absolute or `data:application/json;base64,...`

> https://docs.google.com/document/d/1U1RGAehQwRypUTovF1KRlpiOFze0b-_2gc6fAH0KY0k/edit#heading=h.lmz475t4mvbx

After `url.resolve` we could get:
- absolute path (which we could fs.readFile)
- embedded (which would be handled by fetch)

src/error-snippets/utils.ts
```diff
@@ -0,0 +1,52 @@
...
+import fs from "fs-extra";
+import { SOURCE_MAP_URL_COMMENT } from "./constants";
+
+interface formatFileNameHeaderOpts {
+    line: number;
+    linesAbove: number;
+    linesBelow: number;
+}
+
+export const softFileURLToPath = (fileName: string): string => {
```
KuznetsovRoman, line 13: For `esm`, `fileName` starts with `file://`

src/error-snippets/utils.ts
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    try {
+        return fileURLToPath(fileName);
+    } catch (_) {
+        return fileName;
+    }
+};
+
+export const formatFileNameHeader = (fileName: string, opts: formatFileNameHeaderOpts): string => {
```
KuznetsovRoman, line 25: Formats this kind of header above code snippet:
```
    . | // /Users/foo/bar/project/tests-dir/test.js
    . |
    9 |     it.only('only test', async ({browser}) => {
```

src/error-snippets/utils.ts
```diff
@@ -0,0 +1,52 @@
...
+    const response = await fetch(fileName);
+    const responseText = await response.text();
+
+    if (responseText.includes(SOURCE_MAP_URL_COMMENT) || !response.headers.has("SourceMap")) {
+        return responseText;
+    }
+
+    const sourceMapUrl = response.headers.get("SourceMap");
+
+    return responseText + "\n" + SOURCE_MAP_URL_COMMENT + sourceMapUrl;
```
KuznetsovRoman, line 51: By specification, source maps could be in HTTP Header `SourceMap`.

If it is, we could just append our file contents and dont think about it elsewhere

src/reporters/flat.js
```diff
@@ -37,6 +37,11 @@ module.exports = class FlatReporter extends BaseReporter {
                 const icon = testCase.isFailed ? icons.FAIL : icons.RETRY;
 
                 this.informer.log(`   ${testCase.browserId}`);
+                if (testCase.errorSnippet) {
+                    this.informer.log("");
```
KuznetsovRoman, line 41: add "more space" to error snippet

src/reporters/flat.js
```diff
@@ -37,6 +37,11 @@ module.exports = class FlatReporter extends BaseReporter {
...
+                    testCase.errorSnippet.split("\n").forEach(line => this.informer.log(line));
```
KuznetsovRoman, line 42: align snippet with log prefix

src/reporters/utils/helpers.js
```diff
@@ -42,7 +43,11 @@ exports.getTestInfo = test => {
     };
 
     if (test.err) {
-        testInfo.error = getTestError(test);
+        testInfo.error = chalk.supportsColor ? getTestError(test) : stripAnsi(getTestError(test));
+    }
+
+    if (test.err && test.err.snippet) {
+        testInfo.errorSnippet = chalk.supportsColor ? test.err.snippet : stripAnsi(test.err.snippet);
```
KuznetsovRoman, line 50: This way we can use colorfull version for console, and remove control characters if user pipes logs into file

src/error-snippets/index.ts
```diff
@@ -0,0 +1,36 @@
...
+    if (!err) {
+        return err;
```
sipayRT, line 16: I totally understand what you're doing here, but it looks weird. why not explicitly return something like `null` or `undefined`?

src/error-snippets/source-maps.ts
```diff
@@ -0,0 +1,45 @@
...
+import { SOURCE_MAP_URL_COMMENT } from "./constants";
+import { softFileURLToPath, getSourceCodeFile } from "./utils";
+import type { SufficientStackFrame, ResolvedFrame } from "./types";
+
+export const extractSourceMaps = async (
+    fileContents: string,
+    fileName: string,
+): Promise<BasicSourceMapConsumer | null> => {
+    const sourceMapsStartIndex = fileContents.indexOf(SOURCE_MAP_URL_COMMENT);
+    const sourceMapsEndIndex = fileContents.indexOf("\n", sourceMapsStartIndex);
```
sipayRT, line 12: `sourceMapsStartIndex` can be -1 here. so we'll search `\n` in all content

src/error-snippets/source-maps.ts
```diff
@@ -0,0 +1,45 @@
...
+
+    if (sourceMapsStartIndex === -1) {
```
sipayRT, line 14: so this check needs to be moved higher

src/error-snippets/utils.ts
```diff
@@ -0,0 +1,85 @@
...
+    const offsetWidth = String(opts.line + opts.linesBelow).length;
+
+    const filePath = softFileURLToPath(fileName);
+    const relativeFileName = path.isAbsolute(filePath) ? path.relative(process.cwd(), filePath) : filePath;
+    const lineNumberOffset = ".".repeat(lineNumberWidth).padStart(offsetWidth);
+    const offset = `  ${lineNumberOffset} |`;
+    const grayBegin = "\x1B[90m";
+    const grayEnd = "\x1B[39m";
+
+    return [offset + ` // ${relativeFileName}`, offset].map(line => grayBegin + line + grayEnd + "\n").join("");
```
sipayRT, line 43: do I understand correctly that you are joing strings in this format in order to avoid gluing adjacent strings?

src/error-snippets/utils.ts
```diff
@@ -0,0 +1,85 @@
...
+                highlightCode: true,
+                forceColor: true,
+            },
+        );
+
+    return `\n${snippet}\n`;
+};
+
+export const getSourceCodeFile = async (fileName: string): Promise<string> => {
+    fileName = softFileURLToPath(fileName);
```
sipayRT, line 69: instead of rewriting argument it will be better to create new variable

test/src/error-snippets/frames.ts
```diff
@@ -0,0 +1,141 @@
...
+
+        it("should return first seen frame with equal relevance", () => {
+            const error = buildError_(frames.node.internal2, frames.node.repl, frames.node.wdio, frames.browser.wdio);
+
+            const stackFrame = findRelevantStackFrame(error);
+
+            assert.equal(stackFrame?.functionName, "Element.elementErrorHandlerCallbackFn");
+        });
+
+        it("should value other project deps frame more, than wdio frame", () => {
```
sipayRT, line 95: should rate...?

test/src/error-snippets/frames.ts
```diff
@@ -0,0 +1,141 @@
...
+            const error = buildError_(frames.browser.wdio, frames.browser.unknownWdio, frames.node.projectDeps);
+
+            const stackFrame = findRelevantStackFrame(error);
+
+            assert.equal(stackFrame?.functionName, "Object.tryCatcher");
```
sipayRT, line 100: it would be better to pass the value being checked directly in the test. now it is difficult to understand what this value is

test/src/error-snippets/frames.ts
```diff
@@ -0,0 +1,141 @@
...
+        it("should value user frame above all for node", () => {
+            const error = buildError_(frames.node.wdio, frames.node.unknownWdio, frames.node.repl, frames.node.user);
+
+            const stackFrame = findRelevantStackFrame(error);
+
+            assert.equal(stackFrame?.functionName, "Object.<anonymous>");
+        });
+    });
+
+    it("resolveWithStackFrame", () => {
```
sipayRT, line 125: not the most understandable test name :)

test/src/error-snippets/index.ts
```diff
@@ -0,0 +1,131 @@
...
+            fsReadFileStub.withArgs("/foo/file1").rejects(new Error("read file error"));
+
+            const result = await extendWithCodeSnippet(error);
+
+            assert.strictEqual(error, savedError);
+            assert.strictEqual(error, result);
+            assert.calledOnceWith(loggerStub.warn, "Unable to apply code snippet:");
+        });
+
+        it("should return the same Object.is object", async () => {
```
sipayRT, line 77: not sure that the name of the test needs to be specified about `Object.is`

test/src/error-snippets/utils.ts
```diff
@@ -0,0 +1,203 @@
...
+            const fileName = "myFile.txt";
+
+            const opts = {
+                line: 10,
+                linesAbove: 5,
+                linesBelow: 5,
+            };
+
+            const result = formatFileNameHeader(fileName, opts);
+            const expected = ["   . | // myFile.txt", "   . |"]
```
sipayRT, line 60: for `expected` value it would be better to use plain string instead of generation it

sipayRT (approved):

src/error-snippets/frames.ts
```diff
@@ -0,0 +1,61 @@
+import _ from "lodash";
+import ErrorStackParser from "error-stack-parser";
+import logger from "../utils/logger";
+import { softFileURLToPath } from "./utils";
+import type { ResolvedFrame, SufficientStackFrame } from "./types";
+
+// prettier-ignore
```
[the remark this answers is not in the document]
KuznetsovRoman: Prettier makes it ugly (nodeInternals would take extra lines)

src/error-snippets/index.ts
KuznetsovRoman: Why do you want to explicitly return `null` or `undefined` here?

I want it to return the same `Object.Is` equality object, and dont think about "what if it returned null" in other places. Its easier to think like "extendWithCodeSnippet added some code snippet if it could. If it didn't, anyway we won't try to do something with it"

src/error-snippets/source-maps.ts
KuznetsovRoman: Yes, so what? `indexOf` still can't throw the exception.

src/error-snippets/source-maps.ts
KuznetsovRoman: Yes, we could save some CPU cycles with it (and nothing more), but its more readable:
```ts
    const sourceMapsStartIndex = fileContents.indexOf(SOURCE_MAP_URL_COMMENT);
    const sourceMapsEndIndex = fileContents.indexOf("\n", sourceMapsStartIndex);

    if (sourceMapsStartIndex === -1) {
        return null;
    }
```
```ts
    const sourceMapsStartIndex = fileContents.indexOf(SOURCE_MAP_URL_COMMENT);

    if (sourceMapsStartIndex === -1) {
        return null;
    }

    const sourceMapsEndIndex = fileContents.indexOf("\n", sourceMapsStartIndex);
```

src/error-snippets/utils.ts
KuznetsovRoman: Yes.

`offset` is `   .. | ` part. So it doesn't look like this:
```
// file-name.ts

   17 |
 > 18 |         const elem = await browser.$('not-existing').waitForDisplayed();
      |                     ^ Error: element ("not-existing") still not displayed after 5000ms
```
but this:
```
   .. | // file-name.ts
   .. |
   17 |
 > 18 |         const elem = await browser.$('not-existing').waitForDisplayed();
      |                     ^ Error: element ("not-existing") still not displayed after 5000ms
```
Then we paint each line with `gray` (but each line separately, so our datetime prefix won't be painted in gray)

test/src/error-snippets/frames.ts
KuznetsovRoman: `stackFrameResolver` -> `stackFrameLocationResolver`
`resolveWithSourceMap` -> `resolveLocationWithSourceMap`
`resolveWithStackFrame` -> `resolveLocationWithStackFrame`

test/src/error-snippets/index.ts
KuznetsovRoman: Why not?
`{}` and `{}` are the same objects. But there is a nuance)

PR #921, opened 2024-05-03
fix: dont remove imports in nodejs env

KuznetsovRoman:
## What is done

Fixed removing extra imports in nodejs env.

### DIFF
--- a/src/bundle/test-transformer.ts
+++ b/src/bundle/test-transformer.ts
@@ -6,7 +6,7 @@ import { TRANSFORM_EXTENSIONS, JS_EXTENSION_RE } from "./constants";
 import type { NodePath, PluginObj, TransformOptions } from "@babel/core";
 import type { ImportDeclaration } from "@babel/types";
 
-export const setupTransformHook = (): VoidFunction => {
+export const setupTransformHook = (opts: { removeNonJsImports?: boolean } = {}): VoidFunction => {
     const transformOptions: TransformOptions = {
         browserslistConfigFile: false,
         babelrc: false,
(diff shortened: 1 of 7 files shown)


src/test-reader/test-parser.js
```diff
@@ -15,11 +15,13 @@ const clearRequire = require("clear-require");
 const path = require("path");
 
 class TestParser extends EventEmitter {
+    #opts;
     #buildInstructions;
 
-    constructor() {
+    constructor(opts = {}) {
```
shadowusr, line 21: I strongly suggest having explicit boolean option with a sensible name, like "removeJsImports" (if I got right what it does)

Now, when we pass testRunEnv, and get different imports behavior, it looks very unobvious and every time someone will use this class one'd need to investigate how opts are used to understand what behaviour will change.

Also, JSDoc comment with explicit opts type declaration would be really nice to have.

src/bundle/test-transformer.ts
```diff
@@ -22,23 +22,26 @@ export const setupTransformHook = (): VoidFunction => {
...
-                                path.remove();
-                            }
-                        },
-                    },
-                }),
-            ],
         ],
     };
 
+    const customIgnoreImportsPlugin = (): PluginObj => ({
```
shadowusr, line 28: Thinking out loud, I wonder if this was being applied to nodejs code by genuine oversight or for a reason... But since before having CT, everything worked fine without this code, it should work fine now, too

shadowusr (approved):
Looks good to me except for the opts naming in test parser.

src/bundle/test-transformer.ts
KuznetsovRoman: I think he just didn't thought ".tsx" files could be read in "nodejs" env, without explicit file endings.
As we in "nodejs" env don't know statically "which extension" the file have, i think it is better to just don't do anything, as it produces hard to debug problems.

I suggest to think about it later, then with CT contributor we could discuss this. For example, we could only remove unnamed imports (like `import "styles.css"`) and don't touch named imports (like `import myModule from "foo.css"`), or have specific blacklist of extensiong (for example, remove `import foo from "styles.css"`, but don't remove `import {foo} from "list.constants"`)

PR #928, opened 2024-05-15
fix: correctly screen element after moveTo

DudaGod:
fix - https://github.com/gemini-testing/testplane/issues/779


shadowusr (approved):
Is round safe to use here in terms of flakiness? Would floor be better or not?

DudaGod:
> Is round safe to use here in terms of flakiness? Would floor be better or not?

It is safe because under the hood `Math.floor` is used for `top` and `left` coords and `Math.ceil` for `right` and `bottom`. Here - https://github.com/gemini-testing/testplane/blob/master/src/browser/client-scripts/rect.js#L80-L87

PR #933, opened 2024-05-20
fix: "assertView", "moveCursorTo" and "currentTest" types

DudaGod:


src/browser/commands/types.ts
```diff
@@ -87,8 +88,9 @@ export type AssertViewCommand = (
     selectors: string | string[],
     opts?: AssertViewOpts,
 ) => Promise<void>;
+
 export type AssertViewElementCommand = (
-    this: WebdriverIO.Element,
+    this: WebdriverIO.Element | ChainablePromiseElement<WebdriverIO.Element>,
```
DudaGod, line 93: In order to be able write code like - `await browser.$("foo").assertView("plain")`. Without this I should write it like this - `await (await browser.$("foo")).assertView("plain");`

sipayRT (approved):

shadowusr (approved):

PR #936, opened 2024-05-23
feat(ct): support hot module replacement

DudaGod:
## What is done

Vite which is used inside component tests support hmr (hot module replacement) it means that it can reload page if something changed (for example if user modify test file). So in this PR I implement ability to rerun test runnables if vite reload page with test.

### DIFF
--- a/src/worker/browser-env/runner/test-runner/index.ts
+++ b/src/worker/browser-env/runner/test-runner/index.ts
@@ -7,10 +7,16 @@ import urljoin from "url-join";
 import NodejsEnvTestRunner from "../../../runner/test-runner";
 import { wrapExecutionThread } from "./execution-thread";
 import { WorkerEventNames } from "./types";
-import { WORKER_EVENT_PREFIX, SUPPORTED_ASYMMETRIC_MATCHER } from "./constants";
+import {
+    WORKER_EVENT_PREFIX,
+    SUPPORTED_ASYMMETRIC_MATCHER,
+    BRO_INIT_TIMEOUT_ON_RECONNECT,
+    BRO_INIT_INTERVAL_ON_RECONNECT,
+} from "./constants";
 import { VITE_RUN_UUID_ROUTE } from "../../../../runner/browser-env/vite/constants";
 import logger from "../../../../utils/logger";
 import RuntimeConfig from "../../../../config/runtime-config";
+import { AbortOnReconnectError } from "../../../../errors/abort-on-reconnect-error";
 
 import { BrowserEventNames } from "../../../../runner/browser-env/vite/types";
 import type { Selector } from "webdriverio";
@@ -35,6 +41,8 @@ export class TestRunner extends NodejsEnvTestRunner {
     private _socket: WorkerViteSocket;
     private _runUuid: string = crypto.randomUUID();
     private _runOpts!: WorkerTestRunnerRunOpts;
+    private _isReconnected: boolean = false;
+    private _broInitResOnReconnect: Error[] | null = null;
 
     constructor(opts: WorkerTestRunnerCtorOpts) {
         super(opts);
@@ -59,15 +67,118 @@ export class TestRunner extends NodejsEnvTestRunner {
 
     async run(opts: WorkerTestRunnerRunOpts): Promise<WorkerRunTestResult> {
         this._runOpts = opts;
+        let error: Error | undefined | null;
 
-        const results = await super.run({ ...opts, ExecutionThreadCls: wrapExecutionThread(this._socket) });
+        try {
+            await super.prepareToRun(this._runOpts);
 
+            error = await this._runWithAbort<Error | undefined>(throwIfAborted => {
+                const ExecutionThreadCls = wrapExecutionThread(this._socket, throwIfAborted);
+
+                return super.runRunnables(ExecutionThreadCls) as Promise<Error | undefined>;
+            });
+        } catch (err) {
+            error = err as Error;
+        }
+
+        while (error instanceof AbortOnReconnectError) {
+            error = null;
+
+            try {
+                await this._waitBroInitOnReconnect();
+
+                error = await this._runWithAbort<Error | undefined>(throwIfAborted => {
+                    const ExecutionThreadCls = wrapExecutionThread(this._socket, throwIfAborted);
+
+                    return super.runRunnables(ExecutionThreadCls) as Promise<Error | undefined>;
+                });
+            } catch (err) {
+                error = err as Error;
+            }
+        }
+
+        const results = await super.finishRun(error);
         this._socket.emit(WorkerEventNames.finalize);
+
+        if (error) {
+            throw error;
+        }
 
         return results;
     }
 
+    private _runWithAbort<T>(cb: (throwIfAborted: () => void) => Promise<T>): Promise<T> {
+        const controller = new AbortController();
+        const { signal } = controller;
+
+        let isAborted = signal.aborted;
+
+        signal.addEventListener("abort", () => {
+            isAborted = true;
+        });
+
+        const throwIfAborted = (): void => {
+            if (isAborted) {
+                throw new AbortOnReconnectError();
+            }
+        };
+
+        this._socket.once(BrowserEventNames.reconnect, () => {
+            this._broInitResOnReconnect = null;
+
+            this._socket.once(BrowserEventNames.initialize, errors => {
+                this._broInitResOnReconnect = errors;
+            });
+
+            if (this._browser.state.onReplMode) {
+                RuntimeConfig.getInstance().replServer.close();
+            }
+
+            this._isReconnected = true;
+            controller.abort();
+        });
+
+        return cb(throwIfAborted);
+    }
+
+    private _waitBroInitOnReconnect(): Promise<void> {
+        let intervalId: NodeJS.Timeout | null = null;
+
+        return new P((resolve, reject) => {
+            intervalId = setInterval(() => {
+                if (_.isNull(this._broInitResOnReconnect)) {
+                    return;
+                }
+
+                if (intervalId) {
+                    clearInterval(intervalId);
+                }
+
+                if (_.isEmpty(this._broInitResOnReconnect)) {
+                    resolve();
+                } else {
+                    reject(this._broInitResOnReconnect[0]);
+                }
+            }, BRO_INIT_INTERVAL_ON_RECONNECT).unref();
+        })
+            .timeout(
+                BRO_INIT_TIMEOUT_ON_RECONNECT,
+                `Browser didn't connect to the Vite server after reconnect in ${BRO_INIT_TIMEOUT_ON_RECONNECT}ms`,
+            )
+            .catch(err => {
+                if (intervalId) {
+                    clearInterval(intervalId);
+                }
+
+                throw err;
+            });
+    }
+
     _getPreparePageActions(browser: Browser, history: BrowserHistory): (() => Promise<void>)[] {
+        if (this._isReconnected) {
+            return super._getPreparePageActions(browser, history);
+        }
+
         return [
             async (): Promise<void> => {
                 const { default: expectMatchers } = await import("expect-webdriverio/lib/matchers");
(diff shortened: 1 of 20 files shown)


src/worker/browser-env/runner/test-runner/index.ts
```diff
@@ -59,15 +67,114 @@ export class TestRunner extends NodejsEnvTestRunner {
...
+        };
+
+        this._socket.once(BrowserEventNames.reconnect, () => {
+            this._broInitResOnReconnect = undefined;
+
+            this._socket.once(BrowserEventNames.initialize, errors => {
+                this._broInitResOnReconnect = errors;
+            });
+
+            if (this._browser.state.onReplMode) {
```
DudaGod, line 137: If repl mode is used I need close previous repl server in order to not conflict with new one, which opens on rerun runnables

src/worker/browser-env/runner/test-runner/index.ts
```diff
@@ -59,15 +67,114 @@ export class TestRunner extends NodejsEnvTestRunner {
...
+                    return;
+                }
+
+                intervalId && clearInterval(intervalId);
+
+                _.isEmpty(this._broInitResOnReconnect) ? resolve() : reject(this._broInitResOnReconnect[0]);
+            }, BRO_INIT_INTERVAL_TIMEOUT_ON_RECONNECT).unref();
+        })
+            .timeout(
+                BRO_INIT_TIMEOUT_ON_RECONNECT,
```
DudaGod, line 163: Here I wait only 5 seconds because it should works fast. New url is not open and just refreshed in the same browser tab

src/worker/browser-env/runner/test-runner/index.ts
```diff
@@ -59,15 +67,114 @@ export class TestRunner extends NodejsEnvTestRunner {
...
+            .catch(err => {
+                intervalId && clearInterval(intervalId);
+
+                throw err;
+            });
+    }
+
     _getPreparePageActions(browser: Browser, history: BrowserHistory): (() => Promise<void>)[] {
+        if (this._isReconnected) {
+            return super._getPreparePageActions(browser, history);
```
DudaGod, line 175: Do not open page again and subscribe on events like `callConsoleMethod` or `runBrowserCommand` if browser was reconnected

src/worker/runner/test-runner/index.js
```diff
@@ -21,40 +22,107 @@ module.exports = class TestRunner extends Runner {
...
 
         this._file = file;
         this._config = config;
         this._browserAgent = browserAgent;
     }
 
-    async run({ sessionId, sessionCaps, sessionOpts, state, ExecutionThreadCls = ExecutionThread }) {
-        const test = this._test;
-        const testplaneCtx = test.testplaneCtx || {};
+    async run({ sessionId, sessionCaps, sessionOpts, state }) {
```
DudaGod, line 32: Here I just split this method on three: `prepareToRun`, `runRunnables` and `finishRun`. Code inside them remained the same.

src/worker/runner/test-runner/index.js
```diff
@@ -21,40 +22,107 @@ module.exports = class TestRunner extends Runner {
...
+        await this.prepareToRun({ sessionId, sessionCaps, sessionOpts, state });
+
+        const error = await this.runRunnables(ExecutionThread);
 
-        let browser;
+        return this.finishRun(error);
+    }
+
+    // TODO: make it protected
+    async prepareToRun({ sessionId, sessionCaps, sessionOpts, state }) {
```
DudaGod, line 41: I want to use these methods in `BrowserTestRunner` and stub them. So I want make these methods protected but it is on javascript. Therefore, I left todo here for now

test/src/worker/browser-env/runner/test-runner/index.ts
```diff
@@ -357,20 +375,119 @@ describe("worker/browser-env/runner/test-runner", () => {
...
+
+            afterEach(() => clock.restore());
+
+            describe("should throw error if", () => {
+                it("browser not inited after reconnect", async () => {
+                    const socket = mkSocket_() as BrowserViteSocket;
+                    socketClientStub.returns(socket);
+
+                    const promise = run_();
+                    await clock.tickAsync(1);
```
DudaGod, line 407: used in order to give event loop ability to run source code (subscribe on `initialize` event)

src/worker/browser-env/runner/test-runner/index.ts
```diff
@@ -59,15 +67,114 @@ export class TestRunner extends NodejsEnvTestRunner {
...
+
+        signal.addEventListener("abort", () => {
+            isAborted = true;
+        });
+
+        const throwIfAborted = (): void => {
+            if (isAborted) {
+                throw new AbortError();
+            }
+        };
```
KuznetsovRoman, line 128: Wouldn't `isAborted` be closured by value?

I know we could closure `signal`, since it is passed by reference:
`signal.aborted = true`, `if (signal.aborted)`
But im not sure about the value part

src/worker/browser-env/runner/test-runner/index.ts
```diff
@@ -59,15 +67,114 @@ export class TestRunner extends NodejsEnvTestRunner {
...
+
+        this._socket.once(BrowserEventNames.reconnect, () => {
+            this._broInitResOnReconnect = undefined;
```
KuznetsovRoman, line 131: Could we not use `undefined` as the value? There are reasons: https://eslint.org/docs/latest/rules/no-undefined

src/worker/browser-env/runner/test-runner/index.ts
```diff
@@ -59,15 +67,114 @@ export class TestRunner extends NodejsEnvTestRunner {
 
     async run(opts: WorkerTestRunnerRunOpts): Promise<WorkerRunTestResult> {
         this._runOpts = opts;
+        let error: Error | undefined = undefined;
```
KuznetsovRoman, line 70: What's the point of explicitly setting `undefined` here? Can't we just do:
`let error: Error` or `let error?: Error` or `let error: Error | undefiend`?

src/worker/browser-env/runner/test-runner/index.ts
```diff
@@ -59,15 +67,114 @@ export class TestRunner extends NodejsEnvTestRunner {
...
+                const ExecutionThreadCls = wrapExecutionThread(this._socket, throwIfAborted);
+
+                return super.runRunnables(ExecutionThreadCls) as Promise<Error | undefined>;
+            });
+        } catch (err) {
+            error = err as Error;
+        }
+
+        while (error instanceof AbortError) {
+            error = undefined;
```
KuznetsovRoman, line 85: Lets not use explicit `undefined` as value. There are reasons: https://eslint.org/docs/latest/rules/no-undefined
We can substitute it with `null`

src/worker/browser-env/runner/test-runner/index.ts
```diff
@@ -59,15 +67,114 @@ export class TestRunner extends NodejsEnvTestRunner {
...
+
+                error = await this._runWithAbort<Error | undefined>(throwIfAborted => {
+                    const ExecutionThreadCls = wrapExecutionThread(this._socket, throwIfAborted);
+
+                    return new Promise((resolve, reject) => {
+                        (super.runRunnables(ExecutionThreadCls) as Promise<Error | undefined>)
+                            .then(resolve)
+                            .catch(reject);
+                    });
+                });
```
KuznetsovRoman, line 98: Whats the point of returning a manual promise here with `then(resolve).catch(reject)`? 

Can't we just do `return super.runRunnables(ExecutionThreadCls) as Promise<Error | undefined>`

src/worker/browser-env/runner/test-runner/index.ts
KuznetsovRoman, line 128: Im not sure if it will work, since closure might caprute `isAborted` by value.
Im sure we can use `signal.aborted` directly, since `signal` is passed by reference, but not sure about this one.

src/worker/browser-env/runner/test-runner/index.ts
KuznetsovRoman, line 131: I suggest to not use explicit undefined. Can't we use `null` here?

src/worker/browser-env/runner/test-runner/index.ts
```diff
@@ -59,15 +67,114 @@ export class TestRunner extends NodejsEnvTestRunner {
...
+            controller.abort();
+        });
+
+        return cb(throwIfAborted);
+    }
+
+    private _waitBroInitOnReconnect(): Promise<void> {
+        let intervalId: NodeJS.Timeout | null = null;
+
+        return new P((resolve, reject) => {
```
KuznetsovRoman, line 151: Why bluebird promise?
Lets not use that legacy library in the new code.

src/worker/browser-env/runner/test-runner/index.ts
```diff
@@ -59,15 +67,114 @@ export class TestRunner extends NodejsEnvTestRunner {
...
+            intervalId = setInterval(() => {
+                if (_.isUndefined(this._broInitResOnReconnect)) {
+                    return;
+                }
+
+                intervalId && clearInterval(intervalId);
+
+                _.isEmpty(this._broInitResOnReconnect) ? resolve() : reject(this._broInitResOnReconnect[0]);
```
KuznetsovRoman, line 159: We don't usually use ternary expressions if we don't need the value, so i suggest to use `if ... else` here, for consistency

src/worker/browser-env/runner/test-runner/index.ts
```diff
@@ -59,15 +67,114 @@ export class TestRunner extends NodejsEnvTestRunner {
...
+    private _waitBroInitOnReconnect(): Promise<void> {
+        let intervalId: NodeJS.Timeout | null = null;
+
+        return new P((resolve, reject) => {
+            intervalId = setInterval(() => {
+                if (_.isUndefined(this._broInitResOnReconnect)) {
+                    return;
+                }
+
+                intervalId && clearInterval(intervalId);
```
KuznetsovRoman, line 157: Don't we usually use `if (...)`?

src/worker/browser-env/runner/test-runner/index.ts
```diff
@@ -59,15 +67,114 @@ export class TestRunner extends NodejsEnvTestRunner {
...
+
+                _.isEmpty(this._broInitResOnReconnect) ? resolve() : reject(this._broInitResOnReconnect[0]);
+            }, BRO_INIT_INTERVAL_TIMEOUT_ON_RECONNECT).unref();
+        })
+            .timeout(
+                BRO_INIT_TIMEOUT_ON_RECONNECT,
+                `Browser didn't connect to the Vite server after reconnect in ${BRO_INIT_TIMEOUT_ON_RECONNECT}ms`,
```
KuznetsovRoman, line 164: Why could that happen? Can we tell the user "what is wrong" information and/or "how to fix it"? 
That message itself only could be usefull to Testplane developers
P.S.: if you don't know any scenarios where it can happen, its ok to leave it be as it is.

src/worker/runner/test-runner/index.js
```diff
@@ -21,40 +22,107 @@ module.exports = class TestRunner extends Runner {
...
+                throw new InternalError("OneTimeScreenshooter instance must be initialized before finish test run");
+            }
+
+            if (this._screenshooter.getScreenshot()) {
+                error.screenshot = this._screenshooter.getScreenshot();
+            }
+        }
+
+        // we need to check session twice:
+        // 1. before afterEach hook to prevent work with broken sessions
```
KuznetsovRoman, line 70: nit:

> prevent work

"prevent working"

KuznetsovRoman (approved):
/ok

sipayRT (approved):

src/worker/browser-env/runner/test-runner/index.ts
DudaGod: It can happen if something stuck in the browser and I didn't get request from it. I think this is a rare case and in most cases users will not see it at all. But unfortunately I can't tell the user exactly what caused the error.

PR #944, opened 2024-06-05
docs: add info how to run component tests

DudaGod:
## What is done

I wrote the documentation about the new option `testRunEnv` and about new feature - component testing.
I also added a folder with ready-made examples to run component tests.


README.md
```diff
@@ -29,6 +29,8 @@ Testplane (ex-Hermione) is a battle-hardened framework for testing web apps at a
 
 ⚡ **Infinite Extensibility:** Testplane offers a versatile plugin system with dozens of open-source plugins on GitHub, along with custom reporters, commands, and execution logic.
 
+📦 **Various Test Environments:** With Testplane you can run tests not only in Node.js environment but also in the browser. It means you can run e2e/integration tests in Node.js and unit/component tests in browser.
```
DudaGod, line 32: here I tried to describe the feature in the style of how Nikolai does it, but I'm not sure that I do it right.

examples/component-testing/create-react-app/.gitignore
```diff
@@ -0,0 +1,23 @@
+# See https://help.github.com/articles/ignoring-files/ for more about ignoring files.
```
DudaGod, line 1: create-react-app folder generated using `npx create-react-app create-react-app --template typescript`. I just install few modules and configure testplane and vite

.npmignore
```diff
@@ -2,3 +2,4 @@
 .npmignore
 .sublime-project*
 test/
+examples/
```
KuznetsovRoman, line 5: Shouldn't we add here "hermione-report" just like in ".gitignore"?

README.md
KuznetsovRoman: > You can run tests not only in Node.js environment but also in the browser

"So what?" the user might ask.

I suggest to write it with component testing title and "because you can run your tests not only in node.js environment, but also in the browser" so the accent is on the CT. Average user wouldn't know the difference between those two and especially couldn't think about "what it is for". He probably will just look at "various test environments" and think about various browsers, or node versions, etc...

docs/component-testing.md
```diff
@@ -0,0 +1,138 @@
+## Testplane Component Testing (experimental)
+
+Almost all modern web interfaces are written using frameworks (React, Vue, Svelte, ...) to simplify the creation, reuse and composition of web components. It is important to test such components in isolation from each other to be sure that each component is doing its job correctly. Just like we write unit tests separately from integration tests. Testplane already supports testing components using [Storybook](https://storybook.js.org/) (via [@testplane/storybook](https://github.com/gemini-testing/testplane-storybook) plugin), but this tool is not relevant for all projects. Therefore, we have developed another component testing option that does not require the use of Storybook.
```
KuznetsovRoman, line 3: 👍 for "testplane-storybook" advertising

> Almost all modern web interfaces are written

I think more english would be "Almost every modern web interface is written..."

KuznetsovRoman: > to be sure that each component

"that" seems to be extra

docs/component-testing.md
```diff
@@ -0,0 +1,138 @@
...
+
+### Implementation options for component testing
+
+Component testing is a type of testing in which the logic of a web component is tested in isolation from the web page in which it is used. In order to perform such a test, you need to be able to render the component correctly. [JSDom](https://github.com/jsdom/jsdom) is often used for this task (also used inside Jest), which renders web components using the Node.js virtual renderer, i.e. without using a real browser. On the one hand, it works faster (the browser does not launch), and on the other hand, it is less stable, since the checks are not performed in a real browser. The second popular solution is to use a very fast dev server [Vite](https://vitejs.dev/), which supports many frameworks (React, Vue, Svelte, ...) and is responsible for rendering components in isolation.
```
KuznetsovRoman, line 7: > "in which"

"where"

> (also used inside Jest)

"it is also used inside of Jest"

> using the Node.js virtual renderer, i.e. without using a real browser

I dont get what that "i.e." (in example) means here. I think, we should use "that is" here instead.

> the browser does not launch

What does the browser not launch? We should use passive voice here: "the browser is not being launched"

docs/component-testing.md
```diff
@@ -0,0 +1,138 @@
...
+
+We chose the option using Vite, as this approach ensures that the page is tested more closely to reality (as if the user had opened it). At the same time, the tests themselves run a little longer than in the JSDom. But the most important thing for us is the stability and reproducibility of test results, so the choice was obvious.
```
KuznetsovRoman, line 9: > approach ensures that the page

extra "that"

> `of test results`

There might be missing "the" article

docs/component-testing.md
```diff
@@ -0,0 +1,138 @@
...
+### How to use it?
+
+We will set up testing of react components written in Typescript. Therefore, first of all, we will install the necessary dependencies:
+
+```bash
+npm i testplane vite @vitejs/plugin-react @testing-library/react --save-dev
+npm i react --save
+```
+
+Now let's create a Vite config in which we will connect the plugin to support React. Example:
```
KuznetsovRoman, line 20: > in which

"where"

docs/component-testing.md
```diff
@@ -0,0 +1,138 @@
...
+            ],
+            browsers: [
+                'chrome'
+            ]
+        },
+    },
+}
+```
+
+And in the end, we can write a test in which we simply output the `document` value to the console without using the `browser.execute` command:
```
KuznetsovRoman, line 57: > in which

"where"

> browser.execute

Maybe we should make it hyperlink to wdio command? Not everyone knows what it does

docs/component-testing.md
```diff
@@ -0,0 +1,138 @@
...
+> - there is no access to `currentTest` from `it`, `beforeEach` and `afterEach`. It will appear in the near future;
+> - the [@testplane/global-hook](https://github.com/gemini-testing/testplane-global-hook) plugin is temporarily not supported.
+
+
+### What additional features are supported?
+
+#### Hot Module Replacement (HMR)
+
+[HMR](https://vitejs.dev/guide/api-hmr.html) is supported in Vite. This means that if you change the loaded file, either the component will be remounted, or the page will be completely preloaded. If the component is described in a separate file (i.e. not in the same file as the test), a remount will be performed, but the test will not be restarted. And if you change the test file, the page will be reloaded, which will cause Testplane to interrupt the execution of the current test and start it again. Due to this feature, you can quickly develop components in Vite and write tests for them. It is recommended to use it together with the [REPL mode](./cli.md#repl-mode).
```
KuznetsovRoman, line 130: > This means that

"It means" (without "that")

docs/component-testing.md
```diff
@@ -0,0 +1,138 @@
...
+
+#### Using the browser and expect instances directly in the browser DevTools
+
+Instances of the `browser` and `expect` are available inside the browser in which the test is being performed. It is quite convenient to use it when debugging the test.
```
KuznetsovRoman, line 134: I would add "are available inside of the browser's global scope".

docs/component-testing.md
```diff
@@ -0,0 +1,138 @@
...
+
+#### Logs from the browser console in the terminal
+
+Calling the `log`, `info`, `warn`, `error`, `debug` and `table` commands on the `console` object in the browser causes information to be displayed not only in the browser's DevTools, but also in the terminal from which Testplane was launched. I.e., you can call `console.log` in the test/component and then see the result of it execution in the terminal. This is also handy when debugging the test.
```
KuznetsovRoman, line 138: > and then see the result

"And you will be able to see the result"

> This is also handy

I suggest "This is especially handy"

docs/config.md
```diff
@@ -551,6 +552,38 @@ By default, Testplane will run all browsers simultaneously. Sometimes (i.e. when
 #### fileExtensions
 Ability to set file extensions, which Testplane will search on the file system. Default value is `[".js", ".mjs", ".ts", ".mts", ".jsx", ".tsx"]`.
 
+#### testRunEnv
+Ability to specify in which environment the tests should be run. There are two available environments:
+
+* `nodejs` – Testplane will run tests in nodejs process. Default value.
+* `browser` – Testplane will run tests inside a browser.
```
KuznetsovRoman, line 559: > inside a browser

"inside of the browser"

docs/config.md
```diff
@@ -551,6 +552,38 @@ By default, Testplane will run all browsers simultaneously. Sometimes (i.e. when
...
+
+The `browser` environment has additional optiosn:
```
KuznetsovRoman, line 561: typo: "optiosn"

examples/component-testing/create-react-app/.testplane.conf.ts
```diff
@@ -0,0 +1,36 @@
+export default {
+    baseUrl: "http://localhost",
+    automationProtocol: "devtools",
+    sessionsPerBrowser: 1,
+    testsPerSession: 10,
+    windowSize: "1280x1024",
```
KuznetsovRoman, line 6: Why 1280x1024? Is it 4/3 ratio?
We could make it HD 16/9: "1280x720"

KuznetsovRoman (approved):

docs/component-testing.md
```diff
@@ -0,0 +1,138 @@
...
+// src/tests/test.testplane.tsx
+it('should log document', async () => {
+    console.log(document);
+});
+```
+
+If such a test were performed in a Node.js environment, then it would have fallen with the error: `ReferenceError: document is not defined`. But in our case, it will be executed directly in the browser, where the global variable `document` is available. Therefore, in the browser and terminal log (we will tell you about this feature below) we will see the following:
+
+```
+{
```
sipayRT, line 69: is this the real output? `document` object has much more fields

sipayRT (approved):

docs/component-testing.md
DudaGod: > I think more english would be "Almost every modern web interface is written..."

Hm, to my mind, my version is better

docs/component-testing.md
DudaGod: > "where"

Why I should use `where` here? It seems that the meaning is the same

> I dont get what that "i.e." (in example) means here. I think, we should use "that is" here instead.

About i.e. - https://en.wiktionary.org/wiki/i.e.

docs/component-testing.md
DudaGod: > where

It seems that the meaning is the same

docs/component-testing.md
DudaGod: You can use `JSON.stringify(document)` in [ya.ru](https://ya.ru/) and you will see the same output.

PR #946, opened 2024-06-06
fix(ct): do not fail on log html elements

DudaGod:
## What is done

The problem occurs when you try to log an html element in test which use browser runner. Because html element has cyclic links and test crashes on call `JSON.stringify` when transferring such an element between processes.

So I use `outerHTML` if it exists.

### DIFF
--- a/src/runner/browser-env/vite/browser-modules/utils/console.ts
+++ b/src/runner/browser-env/vite/browser-modules/utils/console.ts
@@ -1,6 +1,8 @@
 import { CONSOLE_METHODS } from "../constants.js";
 import { BrowserEventNames } from "../types.js";
 
+const isDOMElement = (data: unknown): data is Element => data instanceof Element;
+
 export const wrapConsoleMethods = (): void => {
     const { socket } = window.__testplane__;
 
@@ -8,7 +10,9 @@ export const wrapConsoleMethods = (): void => {
         const origCommand = console[method].bind(console);
 
         console[method] = (...args: unknown[]): void => {
-            socket.emit(BrowserEventNames.callConsoleMethod, { method, args });
+            const preparedArgs = args.map(arg => (isDOMElement(arg) ? arg.outerHTML : arg));
+
+            socket.emit(BrowserEventNames.callConsoleMethod, { method, args: preparedArgs });
 
             origCommand(...args);
         };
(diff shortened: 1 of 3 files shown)


src/runner/browser-env/vite/browser-modules/utils/console.ts
```diff
@@ -1,14 +1,21 @@
 import { CONSOLE_METHODS } from "../constants.js";
 import { BrowserEventNames } from "../types.js";
 
+const isHtmlTag = (data: unknown): data is HTMLElement =>
+    typeof data === "object" && data !== null && "outerHTML" in data;
```
DudaGod, line 5: it's a pretty dumb check, because I didn't find better solution

KuznetsovRoman: Maybe we should add more checks then? e.g. typeof outerHtml, or 2-3 html-element-special properties?

KuznetsovRoman: Or we can check (with weakmap) if object has cycling links, and if it is, then we handle it.
Because "not null object with outerHtml" is pretty weak check

KuznetsovRoman (approved):
Other than that, /ok

src/runner/browser-env/vite/browser-modules/utils/console.ts
DudaGod: Problem reproduced when I found element using `screen` from `@testing-library/react`, like this:
```
const linkElement = screen.getByText(/Counter: 0/i);
console.log(linkElement);
```

In this case I got an error:
```
[02:20:25 +0300] > 16 |     console.log('linkElement:', linkElement);
[02:20:25 +0300]      |            ^ TypeError: Converting circular structure to JSON
[02:20:25 +0300]     --> starting at object with constructor 'HTMLParagraphElement'
[02:20:25 +0300]     |     property '__reactFiber$uk3aqfeobc' -> object with constructor 'FiberNode'
[02:20:25 +0300]     --- property 'stateNode' closes the circle
```

I can use `JSON.stringify` in order to check if object has cycle reference and in this case use `outerHTML`, but there is another problem, for example, not all elements have cyclic reference and code like:
```
JSON.stringify(document.querySelector('body')) // -> '{}'
```

This response `{}` looks useless for user. So I found another solution. I will use the following check:
```
data instanceof Element
```

And if it's return `true` I will use property [outerHTML](https://developer.mozilla.org/en-US/docs/Web/API/Element/outerHTML)

PR #947, opened 2024-06-07
Update transform request doc

rostik404:
TransformRequest example is not working without return statement


DudaGod:
Hello. Thank you for your contribution. I just fix commit message following the rules - https://www.conventionalcommits.org

PR #949, opened 2024-06-14
fix: resolve file field correctly for exports mocha interface

shadowusr:


sipayRT (approved):

shadowusr:
Related to https://github.com/mochajs/mocha/issues/5146

PR #954, opened 2024-06-19
feat: add relativePath to refImg

KuznetsovRoman:
## What is done
`refImg` now has `relativePath` property

### DIFF
--- a/src/types/index.ts
+++ b/src/types/index.ts
@@ -62,6 +62,10 @@ export interface ImageInfo {
     size: ImageSize;
 }
 
+export interface RefImageInfo extends ImageInfo {
+    relativePath: string;
+}
+
 export interface DiffOptions extends LooksSameOptions {
     current: string;
     reference: string;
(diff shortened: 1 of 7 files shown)


src/types/index.ts
```diff
@@ -211,7 +215,7 @@ export type MasterEventHandler<T extends BaseTestplane> = {
     (event: Events["WARNING"], callback: () => void): T;
     (event: Events["ERROR"], callback: (err: Error) => void): T;
 
-    (event: Events["UPDATE_REFERENCE"], callback: (data: { state: string; refImg: ImageInfo }) => void): T;
+    (event: Events["UPDATE_REFERENCE"], callback: (data: { state: string; refImg: RefImageInfo }) => void): T;
```
KuznetsovRoman, line 218: Actually, `refImg` here still could be  `ImageInfo` type, but only if its emitted from the outside (from the html-reporter, for example)

DudaGod (approved):

PR #955, opened 2024-06-24
feat: add traceparent header to all requests

Kabir-Ivan:
A randomly generated traceparent header unique for each test is added to all requests.

### DIFF
--- a/src/browser/browser.js
+++ b/src/browser/browser.js
@@ -95,6 +95,9 @@ module.exports = class Browser {
                             this.state.testXReqId
                         }${X_REQUEST_ID_DELIMITER}${crypto.randomUUID()}`;
                     }
+                    if (!req.headers["traceparent"] && this.state.traceparent) {
+                        req.headers["traceparent"] = this.state.traceparent;
+                    }
 
                     return req;
                 };
--- a/src/runner/test-runner/regular-test-runner.js
+++ b/src/runner/test-runner/regular-test-runner.js
@@ -81,14 +81,29 @@ module.exports = class RegularTestRunner extends Runner {
         this._test.duration = Date.now() - this._test.startTime;
     }
 
+    _getTraceparent() {
+        const version = "00";
+        const traceId = crypto.randomBytes(16).toString("hex");
+        const parentId = "00" + crypto.randomBytes(7).toString("hex");
+        const traceFlag = "01";
+
+        return `${version}-${traceId}-${parentId}-${traceFlag}`;
+    }
+
     async _getBrowser() {
         try {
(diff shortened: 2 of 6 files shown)


src/runner/test-runner/regular-test-runner.js
```diff
@@ -83,7 +83,12 @@ module.exports = class RegularTestRunner extends Runner {
 
     async _getBrowser() {
         try {
-            const state = { testXReqId: crypto.randomUUID() };
+            const state = {
```
KuznetsovRoman, line 86: nit: it would be nice, if you could get from the code "how the traceparent is being built"
Right now i can't get "what that means", "what are those 00 at the beginning" and such.

I would write it like this and pull it out to another private method `_getTraceparent`:
```ts
_getTraceparent() {
  const version = "00";
  const traceId = crypto.randomBytes(16).toString("hex");
  const parentId = "00" + crypto.randomBytes(7).toString("hex");
  const traceFlag = "01"  

  return `${version}-${traceId}-${parentId}-${traceFlag}`
}
```

```ts
try {
  const state = {
    traceparent: this._getTraceparent(),
    ...
  }
```
That way you can get "00" at the beginning is version, actual traceId is the middle part, and such.

test/src/runner/test-runner/regular-test-runner.js
```diff
@@ -122,7 +128,10 @@ describe("runner/test-runner/regular-test-runner", () => {
         });
 
         it("should modify state if 'testXReqId' is not actual", async () => {
-            const state = { testXReqId: "12345" };
+            const state = {
+                testXReqId: "12345",
+                traceparent: "00-11111111111111111111111111111111-0011111111111111-01",
+            };
```
KuznetsovRoman, line 134: Test is called "should modify state if 'testXReqId' is not actual", but there is `traceparent` check also.

I would write separate test for "traceparent", because its unobvious that this test will fall if `traceparent` is broken.

KuznetsovRoman (commented):
Don't forget to rebase. You could do:
```bash
git checkout master
git pull origin master
git checkout add_traceparent
git rebase master
git push -f origin add_traceparent
```

KuznetsovRoman (approved):

src/browser/browser.js
```diff
@@ -95,6 +95,9 @@ module.exports = class Browser {
                             this.state.testXReqId
                         }${X_REQUEST_ID_DELIMITER}${crypto.randomUUID()}`;
                     }
+                    if (!req.headers["traceparent"] && this.state.traceparent) {
```
DudaGod, line 98: Looks like we should write something about this new header field in documentation.

PR #960, opened 2024-06-26
fix: declare err property on Test type

KuznetsovRoman:


test/src/runner/test-runner/regular-test-runner.js
```diff
@@ -305,13 +305,14 @@ describe("runner/test-runner/regular-test-runner", () => {
...
+                const err = new Error();
 
                 const workers = mkWorkers_();
-                workers.runTest.rejects(new Error());
+                workers.runTest.rejects(err);
 
                 await run_({ runner, workers });
 
-                assert.calledOnceWith(onFail, sinon.match(test));
+                assert.calledOnceWith(onFail, sinon.match(Object.assign(test, { err })));
```
KuznetsovRoman, line 315: `onFail` as before is called with test object, containing "err" property
Now `test` is `undefined`, while before it was not defined (yeah, these things are different for sinon.match).
I.E:
`sinon.match({ foo: undefined })` and `sinon.match({ })` are different things.

src/test-reader/test-object/test.ts
```diff
@@ -8,6 +8,7 @@ type TestOpts = TestObjectData &
 
 export class Test extends ConfigurableTestObject {
     public fn: TestFunction<TestFunctionCtx>;
+    public err?: Error;
```
DudaGod, line 11: Where it is initialized? Does this field exist only in the test instance, or can be in hooks too?

DudaGod (approved):
otherwise ok

src/test-reader/test-object/test.ts
KuznetsovRoman: Here: https://github.com/gemini-testing/testplane/blob/f9239c9c1985f38b01d7d5fd0dd387fab24fc3c8/src/runner/test-runner/regular-test-runner.js#L40

It can not be in hooks: https://github.com/gemini-testing/testplane/blob/master/src/worker/runner/test-runner/index.js#L153

PR #961, opened 2024-06-27
fix: don't add error snippets on screenshot error

KuznetsovRoman:


src/error-snippets/index.ts
```diff
@@ -14,7 +14,7 @@ const stackFrameLocationResolver = async (stackFrame: SufficientStackFrame): Pro
 };
 
 export const extendWithCodeSnippet = async (err: WithSnippetError): Promise<WithSnippetError> => {
-    if (!err) {
+    if (!err || shouldNotAddCodeSnippet(err)) {
```
sipayRT, line 17: we can pass error to the `shouldNotAddCodeSnippet` function and then check all cases inside of the function

sipayRT (approved):

PR #968, opened 2024-07-08
feat: implement ability to specify browser as passive

DudaGod:
### What is done:

- implement ability to specify browser as passive using option "passive: true" in browser config;
- make "browserId" as string for runnables, because I don't see case in which it can be `undefined`;
- add "enable" method for all runnables (use it inside `testplane.also.in`);
- add "also" controller in `global.testplane` in order to be able to run test in passive browser.

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -56,6 +56,7 @@
   - [region](#region)
   - [headless](#headless)
   - [isolation](#isolation)
+  - [passive](#passive)
 - [system](#system)
   - [debug](#debug)
   - [mochaOpts](#mochaopts)
@@ -197,6 +198,7 @@ Option name               | Description
 `region`                  | Ability to choose different datacenters for run in cloud service. Default value is `null`.
 `headless`                | Ability to run headless browser in cloud service. Default value is `null`.
 `isolation`               | Ability to execute tests in isolated clean-state environment ([incognito browser context](https://chromedevtools.github.io/devtools-protocol/tot/Target/#method-createBrowserContext)). Default value is `false`, but `true` for chrome@93 and higher.
+`passive`                 | Ability to make browser passive. Tests are not run in passive browsers by default. Default value is `false`.
 
 #### desiredCapabilities
 **Required.** Used WebDriver [DesiredCapabilities](https://github.com/SeleniumHQ/selenium/wiki/DesiredCapabilities). For example,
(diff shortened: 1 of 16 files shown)


src/test-reader/controllers/also-controller.ts
```diff
@@ -0,0 +1,41 @@
...
+                    obj.enable();
+                }
+            });
+        });
+    }
+
+    #match(browserId: string, matchers: string | RegExp | Array<string | RegExp>): boolean {
+        return ([] as Array<string | RegExp>).concat(matchers).some(m => {
+            return m instanceof RegExp ? m.test(browserId) : m === browserId;
+        });
```
KuznetsovRoman, line 39: Can we throw an error, if specified browser does not exist?
Why are you comparing strings as `===` instead of `browserId.includes(m)`? This way we could write `testplane.also.in("chrome-phone")` and run the test in all of the chrome phones (e.g. chrone-phone-121, chrome-phone-112, chrome-phone-dark...)

src/test-reader/test-object/configurable-test-object.ts
```diff
@@ -86,7 +91,7 @@ export class ConfigurableTestObject extends TestObject {
     }
 
     get browserId(): ConfigurableTestObjectData["browserId"] {
-        return this.#getInheritedProperty<ConfigurableTestObjectData["browserId"]>("browserId", undefined);
+        return this.#getInheritedProperty<ConfigurableTestObjectData["browserId"]>("browserId", "");
```
KuznetsovRoman, line 94: 👍

docs/config.md
```diff
@@ -197,6 +198,7 @@ Option name               | Description
 `region`                  | Ability to choose different datacenters for run in cloud service. Default value is `null`.
 `headless`                | Ability to run headless browser in cloud service. Default value is `null`.
 `isolation`               | Ability to execute tests in isolated clean-state environment ([incognito browser context](https://chromedevtools.github.io/devtools-protocol/tot/Target/#method-createBrowserContext)). Default value is `false`, but `true` for chrome@93 and higher.
+`passive`                 | Ability to make browser passive in which tests do not run by default. Default value is `false`.
```
KuznetsovRoman, line 201: I would split the first sentence into two.
- Ability to make browser passive
- Tests are not run in passive browsers by default

Currently its hard to read, because it could be read as [passive in which tests do not run by default], like there could be another types of "passive" or something like that

docs/config.md
```diff
@@ -491,6 +493,9 @@ Ability to run headless browser in cloud service. Default value is `null`. Can b
 ####  isolation
 Ability to execute tests in isolated clean-state environment ([incognito browser context](https://chromedevtools.github.io/devtools-protocol/tot/Target/#method-createBrowserContext)). It means that `testsPerSession` can be set to `Infinity` in order to speed up tests execution and save browser resources. Currently works only in chrome@93 and higher. Default value is `null`, but `true` for chrome@93 and higher.
 
+#### passive
+Ability to make browser passive in which tests do not run by default. Using [testplane.also.in](./writing-tests.md#also) makes it possible to run test or suite before which it is specified.
```
KuznetsovRoman, line 497: The same as previous

KuznetsovRoman (approved):
Otherwise, /ok

src/test-reader/controllers/also-controller.ts
DudaGod: > Can we throw an error, if specified browser does not exist?

We can do it, but it should be done in another PR in which we will also change logic in `testplane.skip` and `testplane.only`. To my mind they should work the same way.

> Why are you comparing strings as === instead of browserId.includes(m)?

For this case user should use regexp. Other controllers (`skip` and `only`) work in the same way. It means if we want match not strictly then it must be changed in all controllers.

PR #972, opened 2024-07-15
fix: support error snippets nested browser commands

KuznetsovRoman:
Examples: https://nda.ya.ru/t/1XWaQ6YG76y5LA

### DIFF
--- a/src/browser/stacktrace/index.ts
+++ b/src/browser/stacktrace/index.ts
@@ -1,4 +1,4 @@
-import { ShallowStackFrames, applyStackFrames, captureRawStackFrames } from "./utils";
+import { ShallowStackFrames, applyStackTraceIfBetter, captureRawStackFrames } from "./utils";
 import { getBrowserCommands, getElementCommands } from "../history/commands";
 import { runWithHooks } from "../history/utils";
 
@@ -15,7 +15,7 @@ export const runWithStacktraceHooks = ({
 }): ReturnType<typeof fn> => {
     const frames = captureRawStackFrames(stackFilterFunc || runWithStacktraceHooks);
 
-    if (stackFrames.isNested(frames)) {
+    if (stackFrames.areInternal(frames)) {
         return fn();
     }
 
@@ -25,7 +25,7 @@ export const runWithStacktraceHooks = ({
         before: () => stackFrames.enter(key, frames),
         fn,
         after: () => stackFrames.leave(key),
-        error: (err: Error) => applyStackFrames(err, frames),
+        error: (err: Error) => applyStackTraceIfBetter(err, frames),
     });
 };
 
--- a/src/browser/stacktrace/utils.ts
+++ b/src/browser/stacktrace/utils.ts
@@ -1,6 +1,8 @@
+import _ from "lodash";
 import ErrorStackParser from "error-stack-parser";
 import type { SetRequired } from "type-fest";
 import logger from "../../utils/logger";
+import { softFileURLToPath } from "../../utils/fs";
 import { WDIO_IGNORED_STACK_FUNCTIONS, WDIO_STACK_TRACE_LIMIT } from "./constants";
 
 export type RawStackFrames = string;
@@ -48,7 +50,7 @@ const getErrorRawStackFrames = (e: ErrorWithStack): RawStackFrames => {
     return e.stack.slice(errorMessageEndsStackIndex + 1);
 };
 
-export function captureRawStackFrames(filterFunc?: (...args: unknown[]) => unknown): RawStackFrames {
+export const captureRawStackFrames = (filterFunc?: (...args: unknown[]) => unknown): RawStackFrames => {
     const savedStackTraceLimit = Error.stackTraceLimit;
     const targetObj = {} as { stack: RawStackFrames };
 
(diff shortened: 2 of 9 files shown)


src/browser/stacktrace/index.ts
```diff
@@ -15,7 +15,7 @@ export const runWithStacktraceHooks = ({
 }): ReturnType<typeof fn> => {
     const frames = captureRawStackFrames(stackFilterFunc || runWithStacktraceHooks);
 
-    if (stackFrames.isNested(frames)) {
+    if (stackFrames.areInternal(frames)) {
```
KuznetsovRoman, line 18: Now some nested frames are saved.
Only internal ones are completely ignored (`node_modules`, `node:internal`)

src/browser/stacktrace/index.ts
```diff
@@ -25,7 +25,7 @@ export const runWithStacktraceHooks = ({
         before: () => stackFrames.enter(key, frames),
         fn,
         after: () => stackFrames.leave(key),
-        error: (err: Error) => applyStackFrames(err, frames),
+        error: (err: Error) => applyStackTraceIfBetter(err, frames),
```
KuznetsovRoman, line 28: From multiple saved stacktraces, we apply the best one

src/browser/stacktrace/utils.ts
```diff
@@ -59,19 +61,90 @@ export function captureRawStackFrames(filterFunc?: (...args: unknown[]) => unkno
...
+    const fileName: string = softFileURLToPath(frame.fileName!);
+
+    for (const factor in FRAME_REELVANCE) {
+        if (FRAME_REELVANCE[factor].matcher(fileName)) {
+            return FRAME_REELVANCE[factor].value;
+        }
+    }
+
+    return 0;
+};
```
KuznetsovRoman, line 100: Pulled out from "error-snippets"

src/browser/stacktrace/utils.ts
```diff
@@ -59,19 +61,90 @@ export function captureRawStackFrames(filterFunc?: (...args: unknown[]) => unkno
...
 
-export function applyStackFrames(error: Error, frames: RawStackFrames): Error {
+const getStackTraceRelevance = (error: Error): number => {
+    const framesParsed = ErrorStackParser.parse(error);
+
+    return framesParsed.reduce((maxRelevance, frame) => {
+        return Math.max(maxRelevance, getFrameRelevance(frame));
+    }, 0);
+};
```
KuznetsovRoman, line 108: Calculates relevance of each frame and returns stacktrace relevance, based on relevance of the most relevant frame

src/browser/stacktrace/utils.ts
```diff
@@ -59,19 +61,90 @@ export function captureRawStackFrames(filterFunc?: (...args: unknown[]) => unkno
...
+
+const createErrorWithStack = (stack: RawStackFrames, errorMessage = ""): Error => {
+    const newError = new Error(errorMessage);
+
+    newError.stack = getErrorTitle(newError) + "\n" + stack;
+
+    return newError;
+};
```
KuznetsovRoman, line 116: Helper function. Its only needed to parse stack frames.

src/browser/stacktrace/utils.ts
```diff
@@ -59,19 +61,90 @@ export function captureRawStackFrames(filterFunc?: (...args: unknown[]) => unkno
...
     }
 
-    error.stack = getErrorTitle(error) + "\n" + frames;
+    error.stack = getErrorTitle(error) + "\n" + stack;
 
     return error;
-}
+};
+
+export const applyStackTraceIfBetter = (error: Error, stack: RawStackFrames): Error => {
```
KuznetsovRoman, line 128: Calculates current stacktrace relevance and stacktrace relevance of the next nested level call.
Apples the new one if it suites better

src/browser/stacktrace/utils.ts
```diff
@@ -132,13 +205,29 @@ export class ShallowStackFrames {
...
+
+    areInternal(childFrames: RawStackFrames): boolean {
+        const parentStackFrame = this._getParentStackFrame(childFrames);
+
+        if (!parentStackFrame) {
+            return false;
+        }
+
+        const isNodeModulesFrame = (frame: string): boolean => frame.includes("/node_modules/");
+        const isNodeInternalFrame = (frame: string): boolean => frame.includes(" (node:");
```
KuznetsovRoman, line 226: Here it is too expensive to parse frames, because "areInternal" method will be called a lot of times on every browser command. So we are checking node_modules frames and internal frames by these functions

src/error-snippets/utils.ts
```diff
@@ -29,23 +29,16 @@ export const shouldNotAddCodeSnippet = (err: Error): boolean => {
...
-    if (!fileName.startsWith("file://")) {
-        return fileName;
-    }
-
-    try {
-        return fileURLToPath(fileName);
-    } catch (_) {
-        return fileName;
-    }
+const trimAsyncPrefix = (fileName: string): string => {
```
KuznetsovRoman, line 32: Also noticed "error-stack-parser" does not trim "async" prefix on files.

src/utils/fs.ts
```diff
@@ -8,3 +9,15 @@ export const exists = async (path: string): Promise<boolean> => {
...
+    if (!fileName.startsWith("file://")) {
+        return fileName;
+    }
+
+    try {
+        return fileURLToPath(fileName);
+    } catch (_) {
+        return fileName;
+    }
+};
```
KuznetsovRoman, line 23: Pulled out of "error-snippets" because it is used inside "browser/stacktrace" too

DudaGod (approved):

src/browser/stacktrace/utils.ts
KuznetsovRoman: > Why not just leave the old name - `applyStackTrace`?

Lets pretend you don't know implementation details. You call 
```ts
const error = ...
const myNewStack = ...

applyStackTrace(error, myNewStack);
```
What do you think `error.stack` is now equal to? According to `applyStackTrace` function name, of course it is "myNewStack". But turns out, it is not, and its even worse. It is "myNewStack" sometimes, but also sometimes original "error.stack"

So that's why it is called "applyStackTraceIfBetter". That way developer would not be ensured "error.stack" to be "myNewStack". He would get from this function name that "it only apples stacktrace if it is better"

src/browser/stacktrace/utils.ts
```diff
@@ -59,19 +61,90 @@ export function captureRawStackFrames(filterFunc?: (...args: unknown[]) => unkno
...
+
+    newError.stack = getErrorTitle(newError) + "\n" + stack;
+
+    return newError;
+};
+
+const applyStackTrace = (error: Error, stack: RawStackFrames): Error => {
     if (!error || !error.message) {
         return error;
     }
```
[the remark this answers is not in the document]
KuznetsovRoman: In this scenarios, yes. In general, it is critical for the whole "stacktrace" module not to throw errors. Because if it does, the entire run will fail due to unhandled error.

As the check is cheep, i would like to keep it here.

PR #973, opened 2024-07-16
feat: add --run-failed option

Kabir-Ivan:
Make it possible to run only failed tests by adding --run-failed option in CLI.

### DIFF
--- a/src/cli/index.js
+++ b/src/cli/index.js
@@ -58,7 +58,6 @@ exports.run = (opts = {}) => {
             "--update-refs",
             'update screenshot references or gather them if they do not exist ("assertView" command)',
         )
-        .option("--run-failed", "only run tests that failed during the last run")
         .option("--inspect [inspect]", "nodejs inspector on [=[host:]port]")
         .option("--inspect-brk [inspect-brk]", "nodejs inspector with break at the start")
         .option(
@@ -79,7 +78,6 @@ exports.run = (opts = {}) => {
                     set: sets,
                     grep,
                     updateRefs,
-                    runFailed,
                     require: requireModules,
                     inspect,
                     inspectBrk,
@@ -97,7 +95,6 @@ exports.run = (opts = {}) => {
                     sets,
                     grep,
                     updateRefs,
-                    runFailed,
                     requireModules,
                     inspectMode: (inspect || inspectBrk) && { inspect, inspectBrk },
                     replMode: {
--- a/src/cli/index.js
+++ b/src/cli/index.js
@@ -58,6 +58,7 @@ exports.run = (opts = {}) => {
             "--update-refs",
             'update screenshot references or gather them if they do not exist ("assertView" command)',
         )
+        .option("--run-failed", "only run tests that failed during the last run")
         .option("--inspect [inspect]", "nodejs inspector on [=[host:]port]")
         .option("--inspect-brk [inspect-brk]", "nodejs inspector with break at the start")
         .option(
@@ -78,6 +79,7 @@ exports.run = (opts = {}) => {
                     set: sets,
                     grep,
                     updateRefs,
+                    runFailed,
                     require: requireModules,
                     inspect,
                     inspectBrk,
@@ -95,6 +97,7 @@ exports.run = (opts = {}) => {
                     sets,
                     grep,
                     updateRefs,
+                    runFailed,
                     requireModules,
                     inspectMode: (inspect || inspectBrk) && { inspect, inspectBrk },
                     replMode: {
--- a/src/test-reader/test-parser.js
+++ b/src/test-reader/test-parser.js
@@ -16,6 +16,7 @@ const clearRequire = require("clear-require");
 const path = require("path");
 const fs = require("fs-extra");
 const logger = require("../utils/logger");
+const { getShortMD5 } = require("../utils/crypto");
 
 class TestParser extends EventEmitter {
     #opts;
@@ -30,7 +31,7 @@ class TestParser extends EventEmitter {
         super();
 
         this.#opts = opts;
-        this.#failedTests = [];
+        this.#failedTests = new Set();
         this.#buildInstructions = new InstructionsList();
     }
 
@@ -72,7 +73,11 @@ class TestParser extends EventEmitter {
 
         if (config.lastFailed.only) {
             try {
-                this.#failedTests = await fs.readJSON(config.lastFailed.input);
+                this.#failedTests = new Set();
+
+                for (const test of await fs.readJSON(config.lastFailed.input)) {
+                    this.#failedTests.add(getShortMD5(`${test.fullTitle}${test.browserId}${test.browserVersion}`));
+                }
             } catch {
                 logger.warn(
                     `Could not read failed tests data at ${this._config.lastFailed.input}. Running all tests instead`,
@@ -128,15 +133,9 @@ class TestParser extends EventEmitter {
         }
 
         if (config.lastFailed && config.lastFailed.only) {
-            const failedStringified = this.#failedTests.map(data => JSON.stringify(data));
-            treeBuilder.addTestFilter(test =>
-                failedStringified.includes(
-                    JSON.stringify({
-                        fullTitle: test.fullTitle(),
-                        browserId: test.browserId,
-                    }),
-                ),
-            );
+            treeBuilder.addTestFilter(test => {
+                return this.#failedTests.has(getShortMD5(`${test.fullTitle()}${test.browserId}${test.browserVersion}`));
+            });
         }
 
         const rootSuite = treeBuilder.applyFilters().getRootSuite();
--- a/src/test-reader/test-parser.js
+++ b/src/test-reader/test-parser.js
@@ -70,12 +70,14 @@ class TestParser extends EventEmitter {
         const esmDecorator = f => f + `?rand=${rand}`;
         await readFiles(files, { esmDecorator, config: mochaOpts, eventBus });
 
-        try {
-            this.#failedTests = await fs.readJSON(config.lastFailed.input);
-        } catch {
-            logger.warn(
-                `Could not read failed tests data at ${this._config.lastFailed.input}. Running all tests instead`,
-            );
+        if (config.lastFailed.only) {
+            try {
+                this.#failedTests = await fs.readJSON(config.lastFailed.input);
+            } catch {
+                logger.warn(
+                    `Could not read failed tests data at ${this._config.lastFailed.input}. Running all tests instead`,
+                );
+            }
         }
 
         revertTransformHook();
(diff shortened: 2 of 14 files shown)


src/testplane.ts
```diff
@@ -154,21 +154,19 @@ export class Testplane extends BaseTestplane {
...
     protected async _saveFailed(): Promise<void> {
         const dirname = path.dirname(this._config.failedTestsPath);
-        try {
-            await fs.access(dirname);
-        } catch {
-            await fs.mkdir(dirname, { recursive: true });
-        }
-        await fs.writeFile(this._config.failedTestsPath, JSON.stringify(this.failedList));
+        await fs.ensureDir(dirname);
+        await fs.writeJSON(this._config.failedTestsPath, this.failedList, { spaces: 4 });
```
KuznetsovRoman, line 169: Just found out "fs-extra" has "outputJSON" method. It does the same thing as "writeJSON", but also creates the directory if it does not exist: https://github.com/jprichardson/node-fs-extra/blob/master/docs/outputJson.md

So in "_saveFailed" you could just call 
```ts
await fs.outputJSON(this._config.failedTestsPath, this.failedList, { spaces: 4 })
```
Without "path" library methods

src/cli/index.js
```diff
@@ -58,6 +58,7 @@ exports.run = (opts = {}) => {
             "--update-refs",
             'update screenshot references or gather them if they do not exist ("assertView" command)',
         )
+        .option("--run-failed", "only run tests that failed during the last run")
```
shadowusr, line 61: I think it may be named something like `--run-last-failed`, that would sound more clear

src/cli/index.js
```diff
@@ -78,6 +79,7 @@ exports.run = (opts = {}) => {
                     set: sets,
                     grep,
                     updateRefs,
+                    runFailed,
```
shadowusr, line 82: Also, perhaps there should be an option in config and env variable for that as well? I'm not sure though. Thinking about using this in CI, it would be convenient to have separate config for CI to run only failed tests by default or specify it by using env variable. Discussble

src/testplane.ts
```diff
@@ -119,6 +152,25 @@ export class Testplane extends BaseTestplane {
...
+        try {
+            const data = await fs.readFile(this._config.failedTestsPath, "utf8");
+            return JSON.parse(data);
+        } catch {
+            return [];
+        }
+    }
+
+    protected async _saveFailed(): Promise<void> {
+        const dirname = path.dirname(this._config.failedTestsPath);
```
shadowusr: Thinking about using this in CI, I think it would be convenient to be able to fetch this file from remote URL...

But that'd require to have two separate options for input and output. Generally, let's discuss with the team if we need run options like grep and files to be in config?
This would make possible doing so:

```js
export default {
    grep: /some-test/,
    lastFailed: {
        only: true,
        output: '.testplane/file.json',
        input: 'https://ci-artifacts.com/.../file.json',
    }, // or just true/false if default input/output are sufficient

    // other config options
}
```

src/testplane.ts
```diff
@@ -101,7 +112,29 @@ export class Testplane extends BaseTestplane {
...
+        this.on(MasterEvents.ERROR, (err: Error) => this.halt(err));
+
+        if (runFailed) {
+            let previousRunFailedTests: FailedListItem[];
+            this.on(MasterEvents.INIT, async () => {
+                previousRunFailedTests = await this._readFailed();
+            });
+
+            this.on(MasterEvents.AFTER_TESTS_READ, testCollection => {
+                if (previousRunFailedTests.length) {
```
shadowusr, line 125: I'm not sure if this is the best way to do this. In `src/test-reader/test-parser` we have this code:
```js
if (grep) {
    treeBuilder.addTestFilter(test => grep.test(test.fullTitle()));
}
```

We probably can implement this the same way, by adding a filter function.

src/testplane.ts
```diff
@@ -101,7 +112,29 @@ export class Testplane extends BaseTestplane {
...
+                    testCollection.disableAll();
+                    previousRunFailedTests.forEach(({ fullTitle, browserId }) => {
+                        testCollection.enableTest(fullTitle, browserId);
+                    });
+                }
```
shadowusr, line 130: Also, if run-failed option was passed, but no failed tests file was found, should we print a warning about it, what do you think?

shadowusr (changes requested):

src/test-reader/test-parser.js
```diff
@@ -113,6 +125,18 @@ class TestParser extends EventEmitter {
             treeBuilder.addTestFilter(test => grep.test(test.fullTitle()));
         }
 
+        if (config.lastFailed && config.lastFailed.only) {
+            const failedStringified = this.#failedTests.map(data => JSON.stringify(data));
+            treeBuilder.addTestFilter(test =>
+                failedStringified.includes(
+                    JSON.stringify({
```
DudaGod, line 132: Why you stringify it twice? To my mind we don't need `JSON.stringify` here and 3 lines above at all.

src/testplane.ts
```diff
@@ -101,7 +110,13 @@ export class Testplane extends BaseTestplane {
         );
         this.runner = runner;
 
-        this.on(MasterEvents.TEST_FAIL, () => this._fail()).on(MasterEvents.ERROR, (err: Error) => this.halt(err));
+        this.on(MasterEvents.TEST_FAIL, () => this._fail());
+        this.on(MasterEvents.TEST_FAIL, res => this._addFailedTest(res));
+        this.on(MasterEvents.ERROR, (err: Error) => this.halt(err));
+
+        this.on(MasterEvents.RUNNER_END, async () => {
+            await this._saveFailed();
```
DudaGod, line 118: it can be write to one string

test/src/testplane.js
```diff
@@ -467,6 +498,7 @@ describe("testplane", () => {
             it("all runner events with passed event data", () => {
                 const runner = mkNodejsEnvRunner_();
                 const testplane = mkTestplane_();
+                sandbox.stub(testplane, "_addFailedTest");
```
DudaGod, line 501: we never use private methods in tests. You should rewrite it.

test/src/test-reader/test-parser.js
```diff
@@ -32,12 +33,16 @@ describe("test-reader/test-parser", () => {
...
         setupTransformHook = sandbox.stub().named("setupTransformHook").returns(sinon.stub());
+        fsReadJSON = sandbox.stub().resolves([]);
 
         TestParser = proxyquire("src/test-reader/test-parser", {
             "clear-require": clearRequire,
             "./mocha-reader": { readFiles },
             "./controllers/browser-version-controller": browserVersionController,
             "./test-transformer": { setupTransformHook },
+            "fs-extra": {
+                readJSON: fsReadJSON,
```
DudaGod, line 44: Why you don't just stub it using `sandbox.stub(fs, 'readJSON')`?

test/src/testplane.js
```diff
@@ -52,10 +52,16 @@ describe("testplane", () => {
 
         initReporters = sandbox.stub().resolves();
         signalHandler = new AsyncEmitter();
+        fsReadJSON = sandbox.stub().resolves([]);
+        fsOutputJSON = sandbox.stub().resolves();
 
-        Testplane = proxyquire("src/testplane", {
+        Testplane = proxyquire.noPreserveCache().noCallThru()("src/testplane", {
```
DudaGod, line 58: For what these options (`noPreserveCache `, `noCallThru `) used here?

test/src/testplane.js
```diff
@@ -52,10 +52,16 @@ describe("testplane", () => {
...
             "./reporters": { initReporters },
             "./signal-handler": signalHandler,
+            "fs-extra": {
+                readJSON: fsReadJSON,
+                outputJSON: fsOutputJSON,
```
DudaGod, line 63: You can stub it using `sandbox.stub(fs, 'readJSON')`, where fs - it is imported `fs-extra`

DudaGod (changes requested):
There are some questions

src/test-reader/test-parser.js
DudaGod: I don't like solution with iterate through all tests in order to find desired one. I described the above solution with `Map`, which should be much faster. What do you think?

It will also allow us to get rid of `json.strigify` here.

DudaGod (changes requested):
Everything is cool, but I suggest using Map for matching and add `browserVersion` to `failedTests`, because we can use few versions of one browser.

test/src/test-reader/test-parser.js
```diff
@@ -14,6 +14,7 @@ const proxyquire = require("proxyquire").noCallThru();
 const path = require("path");
```
DudaGod, line 14: Where are the tests for this module? There should be quite a lot of them here since the main code is in this file.

In `loadFiles` (public) we can check that:
- `config.lastFailed.only` not set -> do not read json files
- `config.lastFailed.only` not set -> do not call addTestFilter (`parse` should be called after `loadFiles`)
- `config.lastFailed.only` set -> read json from one file (string)
- `config.lastFailed.only` set -> read json from one file (string with commas, case with env)
- `config.lastFailed.only` set -> read json from few files (array of strings)
- and other tests

Method `parse` must be tested as well.

test/src/testplane.js
```diff
@@ -384,18 +398,34 @@ describe("testplane", () => {
...
+                const results = {
+                    fullTitle: () => "Title",
+                    browserId: "chrome",
+                };
+                mkNodejsEnvRunner_(runner => runner.emit(RunnerEvents.TEST_FAIL, results));
 
                 return runTestplane().then(success => assert.isFalse(success));
             });
 
-            it("should halt if there were some errors", () => {
```
DudaGod, line 392: Why you remove this test?

test/src/testplane.js
```diff
@@ -467,16 +497,20 @@ describe("testplane", () => {
             it("all runner events with passed event data", () => {
                 const runner = mkNodejsEnvRunner_();
                 const testplane = mkTestplane_();
+                const results = {
+                    fullTitle: () => "Title",
+                    browserId: "chrome",
```
DudaGod, line 502: then maybe add `browserVersion` as well?

test/src/testplane.js
```diff
@@ -761,8 +795,13 @@ describe("testplane", () => {
         it('should return "true" after some test fail', () => {
             const testplane = mkTestplane_();
 
+            const results = {
+                fullTitle: () => "Title",
+                browserId: "chrome",
```
DudaGod, line 800: then maybe add `browserVersion` as well?

DudaGod (changes requested):
The source is ok, but I don't see a lot of tests on this new functionality. Let's add them.

PR #984, opened 2024-08-01
chore: remove modern-node-polyfills

Kabir-Ivan:


src/runner/browser-env/vite/utils.ts
```diff
@@ -23,6 +26,33 @@ export const getNodeModulePath = async ({
     return resolve(moduleName, path.join(rootFileUrl, parent));
 };
 
+// Just copy-pasted from modern-node-polyfills to avoid potential errors
+export const polyfillPath = async (moduleName: string): Promise<string> => {
```
shadowusr, line 30: If this was taken from a MIT-licensed repo, you are required to include the corresponding license. For instance, you can create a separate file where in the beginning you copy-paste modern-node-polyfills' license to comment and then this code.

For our own ease of use, I'd also include a permalink to source file on github.

shadowusr (approved):
If you tested it well and everything works on https://github.com/gemini-testing/testplane/tree/master/examples/component-testing, looks good to me (besides including the license)

PR #986, opened 2024-08-05
feat: implement new cli command to read tests

DudaGod:
### What is done

Implement ability to read tests using cli command `npx testplane list-tests` which support options like for run tests:
- --config;
- --browser;
- --set;
- --require;
- --grep;
and new options:
- --ignore;
- --silent;
- --output-file;
- --formatter.

The formatter supports 2 values: list (default) and tree. When using `list` user get plain list of unique tests, example:
```json
[
    {
        "id": "5a105e8",
        "titlePath": [
            "test1"
        ],
        "browserIds": [
            "yandex",
            "chrome"
        ],
        "file": "tests/second.hermione.js",
        "pending": false,
        "skipReason": ""
    },
    {
        "id": "d2b3179",
        "titlePath": [
            "suite",
            "test2"
        ],
        "browserIds": [
            "yandex",
            "chrome"
        ],
        "file": "tests/second.hermione.js",
        "pending": false,
        "skipReason": ""
    }
]
```

The same data in `tree` formatter looks like:
```json
[
    {
        "id": "36749990",
        "title": "suite",
        "line": 3,
        "column": 1,
        "file": "example.hermione.js",
        "suites": [],
        "tests": [
            {
                "id": "d2b3179",
                "title": "test2",
                "line": 4,
                "column": 5,
                "browserIds": [
                    "yandex"
                ],
                "pending": false,
                "skipReason": ""
            }
        ],
        "pending": false,
        "skipReason": ""
    },
    {
        "id": "5a105e8",
        "title": "test1",
        "line": 1,
        "column": 1,
        "browserIds": [
            "yandex"
        ],
        "file": "example.hermione.js",
        "pending": false,
        "skipReason": ""
    }
]
```

the important difference is that the `tree` formatter supports `line` and `column` fields. They can be used to find out exactly where the suite or test is declared in the file. It will be used in vscode extension.

----

The same output is supported via API - `testCollection.format(<type>)`


src/cli/commands/list-tests/formatters/tree.ts
```diff
@@ -0,0 +1,129 @@
...
+
+type TreeTest = Omit<TreeSuite, "suites" | "tests"> & {
+    browserIds: string[];
+};
+
+type MainTreeRunnable = (TreeSuite | TreeTest) & {
+    file: string;
+};
+
+export const format = (testCollection: TestCollection): MainTreeRunnable[] => {
```
DudaGod, line 24: here, in order to create the right tree, I move from the tests (which can be disabled) up the parent suites.

src/cli/commands/list-tests/formatters/tree.ts
```diff
@@ -0,0 +1,129 @@
...
+    testCollection.eachTest((test, browserId) => {
+        if ((test as TestDisabled).disabled) {
+            return;
+        }
+
+        if (allTestsById.has(test.id)) {
+            const treeTest = allTestsById.get(test.id)!;
+
+            if (!treeTest.browserIds.includes(browserId)) {
+                treeTest.browserIds.push(browserId);
```
DudaGod, line 37: Like for list tests I just collect all browser ids for the same test

src/cli/commands/list-tests/formatters/tree.ts
```diff
@@ -0,0 +1,129 @@
...
+
+function addSuite(treeSuite: TreeSuite, child: TreeSuite): void {
+    treeSuite.suites.push(child);
+}
+
+function getMainRunanbleFields(runanble: Suite | Test): Partial<Pick<MainTreeRunnable, "file">> {
+    const isMain = runanble.parent && runanble.parent.root;
+
+    return {
+        ...(isMain ? { file: path.relative(process.cwd(), runanble.file) } : {}),
```
DudaGod, line 118: file field exists only in topmost runnable (suite or test in which parent suite is root, which doesn't have any data)

src/cli/commands/list-tests/index.ts
```diff
@@ -0,0 +1,72 @@
...
+    ignore?: Array<string>;
+    silent?: boolean;
+    outputFile?: string;
+    formatter: ValueOf<typeof Formatters>;
+};
+
+export type ListTestsCmd = typeof commander & CommonCmdOpts;
+
+export const registerCmd = (cliTool: ListTestsCmd, testplane: Testplane): void => {
+    withCommonCliOptions({ cmd: cliTool.command(`${commandName}`), actionName: "list" })
```
DudaGod, line 23: Here I get options common options from main command, like: `config`, `browsers`, `sets`, `grep`, ...

src/cli/commands/list-tests/index.ts
```diff
@@ -0,0 +1,72 @@
...
+                    },
+                });
+
+                // eslint-disable-next-line @typescript-eslint/no-var-requires
+                const { format } = require(path.resolve(__dirname, "./formatters", formatter));
+                const result = format(testCollection);
+
+                if (outputFile) {
+                    await fs.ensureDir(path.dirname(outputFile));
+                    await fs.writeJson(outputFile, result);
```
DudaGod, line 55: it is possible to save the result to a file

src/test-reader/index.js
```diff
@@ -20,7 +20,7 @@ module.exports = class TestReader extends EventEmitter {
     }
 
     async read(options = {}) {
-        const { paths, browsers, ignore, sets, grep } = options;
+        const { paths, browsers, ignore, sets, grep, runnableOpts } = options;
```
DudaGod, line 23: Add new option `runnableOpts` which has only one nested option `saveLocations` in order to collect all `line` and column` values for each suite and test.

src/test-reader/mocha-reader/index.js
```diff
@@ -95,6 +99,94 @@ function applyOnly(rootSuite, eventBus) {
     });
 }
 
+function addLocationToRunnables(inBus, config, runnableOpts) {
+    if (!runnableOpts || !runnableOpts.saveLocations) {
```
DudaGod, line 103: Add locations only if user ask it (enable for `tree` formatter)

src/test-reader/mocha-reader/index.js
```diff
@@ -95,6 +99,94 @@ function applyOnly(rootSuite, eventBus) {
...
+        return;
+    }
+
+    const sourceMapSupport = tryToRequireSourceMapSupport();
```
DudaGod, line 107: Check if `source-map-support` from `ts-node` exists in user node_modules. It is used to correctly determine the position of the declaration suite and test (really helpful for tests on typescript)

src/test-reader/mocha-reader/index.js
```diff
@@ -95,6 +99,94 @@ function applyOnly(rootSuite, eventBus) {
...
+    const { suiteMethods, testMethods } = getMethodsByInterface(config.ui);
```
DudaGod, line 108: different ui options support different method names

src/test-reader/mocha-reader/index.js
```diff
@@ -95,6 +99,94 @@ function applyOnly(rootSuite, eventBus) {
...
+
+    inBus.on(MochaEventBus.events.EVENT_FILE_PRE_REQUIRE, ctx => {
```
DudaGod, line 110: Here ctx - it is global value, which has `describe`, `it` and others for `bdd` interface for example

src/test-reader/mocha-reader/index.js
```diff
@@ -95,6 +99,94 @@ function applyOnly(rootSuite, eventBus) {
...
+                methods: suiteMethods,
+                eventName: MochaEventBus.events.EVENT_SUITE_ADD_SUITE,
+            },
+            {
+                methods: testMethods,
+                eventName: MochaEventBus.events.EVENT_SUITE_ADD_TEST,
+            },
+        ].forEach(({ methods, eventName }) => {
+            methods.forEach(methodName => {
+                ctx[methodName] = withLocation(ctx[methodName], { inBus, eventName, sourceMapSupport });
```
DudaGod, line 122: I need wrap methods in order to determine the position of the call

src/test-reader/mocha-reader/index.js
```diff
@@ -95,6 +99,94 @@ function applyOnly(rootSuite, eventBus) {
...
+
+                if (ctx[methodName]) {
+                    ctx[methodName].only = withLocation(ctx[methodName].only, { inBus, eventName, sourceMapSupport });
+                    ctx[methodName].skip = withLocation(ctx[methodName].skip, { inBus, eventName, sourceMapSupport });
+                }
+
+                if (!config.ui || config.ui === "bdd") {
+                    const pendingMethodName = `x${methodName}`;
```
DudaGod, line 130: In `bdd` interface support `xdescribe` and `xit` to skip suites/tests

src/test-reader/mocha-reader/index.js
```diff
@@ -95,6 +99,94 @@ function applyOnly(rootSuite, eventBus) {
...
+function withLocation(origFn, { inBus, eventName, sourceMapSupport }) {
+    if (!_.isFunction(origFn)) {
+        return origFn;
+    }
+
+    const wrappedFn = (...args) => {
+        const origStackTraceLimit = Error.stackTraceLimit;
+        const origPrepareStackTrace = Error.prepareStackTrace;
+
+        Error.stackTraceLimit = 2;
```
DudaGod, line 151: Use only 2 last messages from stack trace. First message it is this file. And second test file.

src/test-reader/mocha-reader/index.js
```diff
@@ -95,6 +99,94 @@ function applyOnly(rootSuite, eventBus) {
...
+                column: frame.getColumnNumber(),
+            };
+        };
+
+        const obj = {};
+        Error.captureStackTrace(obj);
+
+        const location = obj.stack;
+        Error.stackTraceLimit = origStackTraceLimit;
+        Error.prepareStackTrace = origPrepareStackTrace;
```
DudaGod, line 166: return previous values

src/test-reader/mocha-reader/index.js
```diff
@@ -95,6 +99,94 @@ function applyOnly(rootSuite, eventBus) {
...
+
+        inBus.once(eventName, runnable => {
```
DudaGod, line 168: subscribe one on add suite or test in order to save location to it

src/utils/cli.ts
```diff
@@ -0,0 +1,51 @@
+import type { Command } from "@gemini-testing/commander";
+import logger from "./logger";
+import { requireModule } from "./module";
+
+// used from https://github.com/sindresorhus/escape-string-regexp/blob/main/index.js
+const escapeRe = (str: string): string => {
```
DudaGod, line 6: Function which is used from escape-string-regexp. Because there are no typings for it (for version with support commonjs).

docs/programmatic-api.md
```diff
@@ -255,6 +255,9 @@ await testplane.readTests(testPaths, options);
     * **ignore** (optional) `String|Glob|Array<String|Glob>` - patterns to exclude paths from the test search.
     * **sets** (optional) `String[]`– Sets to run tests in.
     * **grep** (optional) `RegExp` – Pattern that defines which tests to run.
+    * **replMode** (optional) `{enabled: boolean; beforeTest: boolean; onFail: boolean;}` - [Test development mode using REPL](./cli.md#--repl). When reading the tests, it checks that only one test is running in one browser.
+    * **runnableOpts** (optional):
```
DudaGod, line 259: new option which gives me ability to save test locations in file (`line` and `column`)

docs/programmatic-api.md
```diff
@@ -317,6 +320,8 @@ TestCollection API:
 
 * `eachRootSuite((root, browserId) => ...)` - iterates over all root suites in collection which have some tests.
 
+* `format(formatterType)` - formats the tests in one of the available formatting types (`list` or `tree`). You can read more about the available formatting types [here](./cli.md#--formattername).
```
DudaGod, line 323: User can format test not only via cli command but also using `testCollection.format` method

src/test-collection/formatters/list.ts
```diff
@@ -0,0 +1,33 @@
+import path from "node:path";
+import type { TestCollection, TestDisabled, FormatterListTest } from "..";
+
+export const format = (testCollection: TestCollection): FormatterListTest[] => {
+    const allTestsById = new Map<string, FormatterListTest>();
+
+    testCollection.eachTest((test, browserId) => {
+        if ((test as TestDisabled).disabled) {
```
DudaGod, line 8: should do nothing with disabled tests (for example for passive browsers)

src/test-collection/formatters/list.ts
```diff
@@ -0,0 +1,33 @@
...
+            return;
+        }
+
+        if (allTestsById.has(test.id)) {
+            const foundTest = allTestsById.get(test.id)!;
```
DudaGod, line 13: for the same test in different browsers, I specify only the field `browserIds`. In order not to store too much of the same data (except for the browser name)

src/test-collection/formatters/tree.ts
```diff
@@ -0,0 +1,125 @@
...
+                treeTest.browserIds.push(browserId);
+            }
+
+            return;
+        }
+
+        const treeTest = createTreeTest(test, browserId);
+        allTestsById.set(treeTest.id, treeTest);
+
+        collectSuites(test.parent!, treeTest, allSuitesById);
```
DudaGod, line 34: here, in order to create the right tree, I move from the tests (which can be disabled) up the parent suites.

src/test-collection/formatters/tree.ts
```diff
@@ -0,0 +1,125 @@
...
+    testCollection.eachTest((test, browserId) => {
+        if ((test as TestDisabled).disabled) {
+            return;
+        }
+
+        if (allTestsById.has(test.id)) {
+            const treeTest = allTestsById.get(test.id)!;
+
+            if (!treeTest.browserIds.includes(browserId)) {
+                treeTest.browserIds.push(browserId);
```
DudaGod, line 25: Like for list tests I just collect all browser ids for the same test

src/test-collection/formatters/tree.ts
```diff
@@ -0,0 +1,125 @@
...
+
+function addSuite(treeSuite: FormatterTreeSuite, child: FormatterTreeSuite): void {
+    treeSuite.suites.push(child);
+}
+
+function getMainRunanbleFields(runanble: Suite | Test): Partial<Pick<FormatterTreeMainRunnable, "file">> {
+    const isMain = runanble.parent && runanble.parent.root;
+
+    return {
+        ...(isMain ? { file: path.relative(process.cwd(), runanble.file) } : {}),
```
DudaGod, line 114: file field exists only in topmost runnable (suite or test in which parent suite is root, which doesn't have any data)

docs/component-testing.md
```diff
@@ -1,3 +1,17 @@
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
```
shadowusr, line 1: This file should be deleted too, iirc you moved it to the site as well.

src/cli/commands/list-tests/index.ts
```diff
@@ -0,0 +1,64 @@
+import path from "node:path";
+import fs from "fs-extra";
+import { ValueOf } from "type-fest";
```
shadowusr, line 3: type-fest is currently a dev dependency. we already experienced issues because of it (when you install testplane, but do not have type-fest which is needed to correctly resolve all typings). I suggest to either inline ValueOf type or move type-fest to dependencies.

src/test-collection/formatters/list.ts
```diff
@@ -0,0 +1,33 @@
+import path from "node:path";
+import type { TestCollection, TestDisabled, FormatterListTest } from "..";
+
+export const format = (testCollection: TestCollection): FormatterListTest[] => {
```
shadowusr, line 4: Here I once again dislike the fact that formatters are tightly coupled with TestCollection, even though conceptually a formatter should only know how to format a test and type-wise can be as simple as `(tests: Test[]) => FormattedTest[]`.

IMO it makes code simpler and more robust, reduces amount of types needed and reduces coupling

But that's my perspective and I do not insist on doing it this way, this may be done in the far future

src/test-collection/index.ts
```diff
@@ -172,4 +206,22 @@ export class TestCollection {
 
         return this;
     }
+
+    format(formatterType: ValueOf<typeof Formatters>): (FormatterListTest | FormatterTreeMainRunnable)[] {
```
shadowusr, line 210: Here I see no benefit in using formatters via TestCollection as opposed to using them directly. IMO it only introduces additional logic, constants, checks, etc. I still think it would be nice to get rid of the whole TestCollection class in the future =)

Not critical, up to you to decide

src/utils/cli.ts
shadowusr: Hmm is there a reason we can't use `_.escapeRegExp` ?

src/test-reader/mocha-reader/utils.js
```diff
@@ -59,6 +59,18 @@ const computeFile = mochaSuite => {
     return null;
 };
 
+const getMethodsByInterface = (mochaInterface = "bdd") => {
+    switch (mochaInterface) {
+        case "tdd":
```
shadowusr, line 64: So there's no way to make this universal? =(

What if user defines its own interface? And I know for sure that we have users that use `exports` interface that are actively using Testplane

It's ok for MVP, but do we have a plan here?

src/test-collection/formatters/tree.ts
shadowusr: Maybe we should leave this comment in code

shadowusr (approved):
Great addition overall! 🔥 

Looks good to me besides documentation bits. Didn't look at unit tests.

src/test-collection/formatters/list.ts
DudaGod: At the moment, the test reading method returns exactly the `TestCollection` structure and I don't want to force the user to convert `TestCollection` to tests array and then call a special method `format`. I mean it's not convenient right now. But I agree that `TestCollection` should be removed in next major in order to simplify the work with tests.

PR #988, opened 2024-08-06
feat: screenshot viewport with assertView by default

Kabir-Ivan:


src/browser/commands/types.ts
```diff
@@ -85,7 +85,7 @@ export interface AssertViewOpts extends Partial<AssertViewOptsConfig> {
 export type AssertViewCommand = (
     this: WebdriverIO.Browser,
     state: string,
-    selectors: string | string[],
+    selectors?: string | string[],
```
shadowusr, line 88: I doubt this will allow doing `assertView('some-state', null, {...opts})`. null needs to be explicitly stated here.

shadowusr (approved):
Looks good to me, if you tested it well, including scrolling the page first and only then doing viewport screenshot.

Typings need to be fixed as well.

Another thing is we need to discuss what's the better choice for API: `assertView('some-state', null, opts)` vs `assertView('some-state', opts)`.

shadowusr (changes requested):
Let's switch to the `assertView('some-state', opts)` API implemented via overloads, so that user didn't have to write null explicitly

docs/writing-tests.md
```diff
@@ -281,6 +281,20 @@ it('some test', async ({ browser }) => {
...
+
+```js
+it('some test', async ({ browser }) => {
+    await browser.url('some/url');
+    await browser.assertView('plain'); // Screen the viewport
+
+    await browser.click('.button');
+    await browser.assertView('clicked', {
+        ignoreDiffPixelCount: 5
+    }); // Screen the viewport with custom options
```
DudaGod, line 294: Why not in one line? And I would write a comment above `assertView`

docs/writing-tests.md
```diff
@@ -300,7 +314,13 @@ it('some test', async ({ browser }) => {
 Parameters:
 
 - state (required) `String` – state name; should be unique within one test
-- selector (required) `String|String[]` – DOM-node selector that you need to capture
+- selector (optional) `String|String[]` – DOM-node selector that you need to capture. If not specified or skipped, will be set to `body` and the following options will be automatically added to opts: 
+```{
+    allowViewportOverflow: true,
```
DudaGod, line 319: Why it is `true` by default? To my mind it should be set the same as for `assertView` by user selector

docs/writing-tests.md
```diff
@@ -300,7 +314,13 @@ it('some test', async ({ browser }) => {
...
+    compositeImage: false,
+    captureElementFromTop: false
+}``` 
+So, assertView without `selector` parameter will take a screenshot of the current viewport.
```
DudaGod, line 323: I suggest adding a line break above.

docs/writing-tests.md
```diff
@@ -300,7 +314,13 @@ it('some test', async ({ browser }) => {
...
 - opts (optional) `Object`:
```
DudaGod, line 324: For assert viewport it is second argument. This is not clear in the current description.

src/browser/commands/assert-view/index.js
```diff
@@ -155,7 +155,23 @@ module.exports.default = browser => {
         testplaneCtx.assertViewResults.add({ stateName: state, refImg: refImg });
     };
 
-    session.addCommand("assertView", async function (state, selectors, opts = {}) {
+    session.addCommand("assertView", async function (state, selectorsOrOpts, opts = {}) {
+        let selectors;
```
DudaGod, line 159: as for me, the code turned out to be a little be dirty here. Why not just create two functions: `assertViewBySelector` and `assertViewByViewport`?

The code would look something like this:
```
const waitSelectorsForExist = (...) => {
  await Promise.map([].concat(selectors), async selector => {...});
};

const assertViewBySelector = (...) => {
  await waitSelectorsForExist(...);

  return assertView(state, selectors, opts);
};

const assertViewByViewport = (...) => {
  // prepare opts
  
  return assertView(state, selectors, opts);
};

const shouldAssertViewport = (...) => {
  return typeof selectorsOrOpts === "string" || _.isArray(selectorsOrOpts);
};

session.addCommand("assertView", async function (state, selectorsOrOpts, opts = {}) {
  return shouldAssertViewport ? assertViewByViewport(...) : assertViewBySelector(...);
});
```

What do you think? To my mind this code is much easier to read.

src/browser/commands/assert-view/index.js
```diff
@@ -155,7 +155,23 @@ module.exports.default = browser => {
...
+        if (selectorsOrOpts) {
+            if (typeof selectorsOrOpts === "string" || _.isArray(selectorsOrOpts)) {
+                selectors = selectorsOrOpts;
+            } else if (typeof selectorsOrOpts === "object") {
+                opts = selectorsOrOpts;
+            }
+        }
+        if (!selectors) {
+            opts = _.defaults(opts, {
+                allowViewportOverflow: true,
```
DudaGod, line 169: Why it is specified as `true` by default?

src/browser/commands/assert-view/index.js
```diff
@@ -155,7 +155,23 @@ module.exports.default = browser => {
...
+                compositeImage: false,
+                captureElementFromTop: false,
+            });
+            selectors = "body";
+        }
         await Promise.map([].concat(selectors), async selector => {
```
DudaGod, line 175: Should we wait `body` in case when user assert viewport?

DudaGod (changes requested):
There are some questions, which I would like to discuss before merge it.

src/browser/commands/assert-view/index.js
Kabir-Ivan: Because without it we will get an error if we try to screen the viewport on a page with height of more than 1 screen.

src/browser/commands/assert-view/index.js
Kabir-Ivan: Not sure as body should probably always exist

docs/writing-tests.md
```diff
@@ -281,6 +281,25 @@ it('some test', async ({ browser }) => {
...
+    await browser.url('some/url');
+    // Scroll 1000 px down
+    await browser.execute(() => {
+        window.scrollTo(0, 1000);
+    });
+    // Screen the viewport
+    await browser.assertView('plain'); 
+
+    // Click the button
+    await browser.click('.button');
```
DudaGod, line 297: I suggest calling it through an explicit element search, like this:
```
await browser.$('.button').click();
```

docs/writing-tests.md
DudaGod: I don't see the answer. To my mind we don't need to specify it.

src/browser/commands/assert-view/index.js
```diff
@@ -155,18 +155,43 @@ module.exports.default = browser => {
...
+
+    const assertViewBySelector = async (browser, state, selectors, opts) => {
+        await waitSelectorsForExist(browser, selectors);
 
         return assertView(state, selectors, opts);
+    };
+
+    const assertViewByViewport = async (state, opts) => {
+        opts = _.defaults(opts, {
+            allowViewportOverflow: true,
```
DudaGod, line 179: if this option is not set, then an error is thrown when taking a screenshot of the viewport?

docs/writing-tests.md
```diff
@@ -300,7 +314,13 @@ it('some test', async ({ browser }) => {
 Parameters:
 
 - state (required) `String` – state name; should be unique within one test
-- selector (required) `String|String[]` – DOM-node selector that you need to capture
+- selector (optional) `String|String[]` – DOM-node selector that you need to capture. If not specified or skipped, will be set to `body` and the following options will be automatically added to opts: 
+```{
```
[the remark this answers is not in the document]
DudaGod: I don't see a fix here. This code should be displayed inside the top item, but for this we need use indent (usually a tab is used for this)

src/browser/commands/assert-view/index.js
Kabir-Ivan: yes

shadowusr (approved):

PR #989, opened 2024-08-06
chore: update webdriverio version

Kabir-Ivan:


src/browser/existing-browser.js
```diff
@@ -187,12 +187,29 @@ module.exports = class ExistingBrowser extends Browser {
...
         this._addMetaAccessCommands(this._session);
         this._decorateUrlMethod(this._session);
+        this._overrideGetElementsList(this._session);
 
         commandsList.forEach(command => require(`./commands/${command}`).default(this));
 
         super._addCommands();
     }
 
+    _overrideGetElementsList(session) {
```
shadowusr, line 197: I think we definitely need a comment with rationale for this

package.json
```diff
@@ -54,10 +54,9 @@
     "@babel/code-frame": "7.24.2",
     "@gemini-testing/commander": "2.15.3",
     "@jspm/core": "2.0.1",
-    "@types/mocha": "10.0.1",
     "@wdio/globals": "8.21.0",
     "@wdio/protocols": "8.22.0",
-    "@wdio/types": "8.21.0",
+    "@wdio/types": "8.39.0",
```
shadowusr, line 59: I see mismatching versions of wdio packages (under the `@wdio` scope). I am 99% sure they need to be the same version (8.39.1 if you chose it for the main webdriverio package), even if everything works now, because it would lead to duplicating packages when user installs testplane

src/worker/browser-env/runner/test-runner/index.ts
```diff
@@ -255,11 +254,11 @@ export class TestRunner extends NodejsEnvTestRunner {
         const { publicAPI: session } = browser;
 
         return async (payload, cb): Promise<void> => {
-            if (!global.expect) {
+            if (typeof expect === "undefined") {
```
shadowusr, line 257: Hmmm, out of curiosity, what's the difference, why does it matter?

shadowusr (changes requested):
Looks good to me if tested on large projects, but we need to have the same version of wdio packages.

package.json
```diff
@@ -92,7 +92,7 @@
     "urijs": "1.19.11",
     "url-join": "4.0.1",
     "vite": "5.1.6",
-    "webdriverio": "8.39.1",
+    "webdriverio": "8.39.0",
```
Kabir-Ivan, line 95: 8.39.0 because there in no @wdio/types@8.39.1

shadowusr (changes requested):
I meant all packages in `@wdio` scope need to use a single version to reduce deps duplication. Can you please check how those are resolved after installing packages using `npm list @wdio/protocols` for example, so that there will be only one version installed.

shadowusr (approved):
If everything works on a large project and there are no duplicate deps, looks good to me

PR #990, opened 2024-08-07
docs: add example on how to use testing-library with typescript

shadowusr:


examples/testing-library/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "m-t-l-e",
```
DudaGod, line 2: What does it mean? Why not just use folder name?

examples/testing-library/package.json
```diff
@@ -0,0 +1,20 @@
...
+  "version": "1.0.0",
+  "description": "",
+  "main": "index.js",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
```
DudaGod, line 7: maybe we should add a testplane launch here?

examples/testing-library/testplane-tests/example.testplane.ts
```diff
@@ -0,0 +1,16 @@
+import type {} from "testplane";
```
DudaGod, line 1: What is it for? Mistake or some hack?

examples/testing-library/testplane-tests/example.testplane.ts
```diff
@@ -0,0 +1,16 @@
...
+
+it("example", async ({browser}) => {
+    // testing library matcher
+    await expect(browser.$("some elem")).toContainHTML("<span>some html</span>");
```
DudaGod, line 5: Hm, I don't understand where code with open some url?

DudaGod (commented):
Does this example really work? Because I don't see testplane config and open url in test example. Looks like it should not work. Can we fix it?

shadowusr:
we have an article on our site covering this already: https://testplane.io/docs/v8/guides/how-to-add-testing-library/

PR #991, opened 2024-08-08
fix: fix ConfigInput type

shadowusr:


src/config/types.ts
```diff
@@ -171,8 +170,8 @@ export interface SetsConfig {
 }
 
 // Only browsers desiredCapabilities are required in input config
-export type ConfigInput = {
-    browsers: Record<string, SetRequired<Partial<CommonConfig>, "desiredCapabilities">>;
+export type ConfigInput = Partial<CommonConfig> & {
+    browsers: Record<string, Partial<CommonConfig> & { desiredCapabilities: WebdriverIO.Capabilities }>;
```
KuznetsovRoman, line 174: > WebdriverIO.Capabilities

It has required `browserName` and optional everything else?

KuznetsovRoman (approved):

src/config/types.ts
shadowusr: Yeah, checked that

PR #992, opened 2024-08-08
fix: error on assertView when viewport has fractional values

DudaGod:
Error looks like: Image to composite must have same dimensions or smaller

### DIFF
--- a/src/browser/client-scripts/index.js
+++ b/src/browser/client-scripts/index.js
@@ -139,6 +139,7 @@ function prepareScreenshotUnsafe(areas, opts) {
             width: viewportWidth,
             height: viewportHeight
         })
+            .round()
             .scale(pixelRatio)
             .serialize(),
         documentHeight: Math.ceil(documentHeight * pixelRatio),


src/browser/client-scripts/index.js
```diff
@@ -139,6 +139,7 @@ function prepareScreenshotUnsafe(areas, opts) {
             width: viewportWidth,
             height: viewportHeight
         })
+            .round()
```
DudaGod, line 142: Here I need to round viewport values like on 83 line in this file (https://github.com/gemini-testing/testplane/pull/992/files#diff-91aa2fe211abdfca85d88f33d0d930b33d5d80e47e37a255375c41ac41b78487R83).

The problem is reproduced if viewport has fractional values and after first round (on 83 line) captured element is located within the viewport. But if it is located in the area just along the border, for example rounded viewport and rounded capture area has the same `bottom` coordinate then in the process of taking screenshots bottom coordinate starts to differ, because here viewport is not rounded. So as a result testplane try to scroll on 1 px and screen and then composite two images. And here it falls. But in fact, there should have been only one screenshot, without any compositing.

This `round` is fixed it.

sipayRT (approved):

DudaGod:
I can't merge it right now, because it leads to differences in screenshots (when viewport has fractional dimensions). It will be implemented in next major version. The problem with `Image to composite must have same dimensions or smaller` solved here - https://github.com/gemini-testing/testplane/pull/997.

PR #993, opened 2024-08-08
fix: fix swc warning

Kabir-Ivan:


src/utils/typescript.ts
```diff
@@ -22,6 +22,7 @@ export const tryToRegisterTsNode = (): void => {
 
         if (JSON.parse(swcRaw)) {
             try {
+                require("@swc/core");
```
Kabir-Ivan, line 25: Require @swc/core to get an error with code "MODULE_NOT_FOUND" if it is not installed

Kabir-Ivan: This way `testplane: you may install @swc/core for significantly faster reading of typescript tests.` will be shown

shadowusr (approved):
I'd rename the `isWorker` param to `isSilent`. Other than that, looks good to me.

PR #1008, opened 2024-09-04
fix: remove extra stack frames

KuznetsovRoman:
Example:
Before (watch 1-st retry): https://nda.ya.ru/t/G_mhJakm782mKy
After (watch 1-st retry): https://nda.ya.ru/t/pt39trMd782mNQ

### DIFF
--- a/src/browser/stacktrace/constants.ts
+++ b/src/browser/stacktrace/constants.ts
@@ -6,3 +6,7 @@ export const WDIO_IGNORED_STACK_FUNCTIONS = [
     "Element.newCommand",
     "Element.elementErrorHandlerCallbackFn",
 ];
+// Other frames are being filtered out by error-stack-parser
+// Declared at https://github.com/stacktracejs/error-stack-parser/blob/v2.1.4/error-stack-parser.js#L17
+// Used at https://github.com/stacktracejs/error-stack-parser/blob/v2.1.4/error-stack-parser.js#L52-L54
+export const STACK_FRAME_REG_EXP = /^\s*at .*(\S+:\d+|\(native\))/m;
--- a/src/browser/stacktrace/utils.ts
+++ b/src/browser/stacktrace/utils.ts
@@ -47,7 +47,13 @@ const getErrorRawStackFrames = (e: ErrorWithStack): RawStackFrames => {
     const errorMessageStackIndex = e.stack.indexOf(e.message);
     const errorMessageEndsStackIndex = e.stack.indexOf("\n", errorMessageStackIndex + e.message.length);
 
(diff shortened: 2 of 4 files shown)


src/browser/stacktrace/constants.ts
```diff
@@ -6,3 +6,7 @@ export const WDIO_IGNORED_STACK_FUNCTIONS = [
     "Element.newCommand",
     "Element.elementErrorHandlerCallbackFn",
 ];
+// Other frames are being filtered out by error-stack-parser
+// Declared at https://github.com/stacktracejs/error-stack-parser/blob/v2.1.4/error-stack-parser.js#L17
+// Used at https://github.com/stacktracejs/error-stack-parser/blob/v2.1.4/error-stack-parser.js#L52-L54
+export const STACK_FRAME_REG_EXP = /^\s*at .*(\S+:\d+|\(native\))/m;
```
KuznetsovRoman, line 12: Example - 
`    at Reflect.get (<anonymous>)`
is a valid frame, but it is not supported by error-stack-parser, though
`    at Object.proxyGetter [as get] (...)`
which is nearby, is supported.
By filtering these in `filterExtraStackFrames` we ensure [rawFramesArr.length === framesParsed.length](https://github.com/gemini-testing/testplane/pull/1008/files#diff-f659cdb6ee50d63a229978b33f5d3d2c982c3acff969c479a49416a1201b8c9aR157) (previously we could see missmatch in frames length and refuse to do anything in order to not make it worse
Since those are useless frames (for example, you will still be able `Object.proxyGetter [as get]` with filepath nearby), its better to remove them, so we can filter out other extra stack frames consistently

src/browser/stacktrace/utils.ts
```diff
@@ -158,21 +158,48 @@ export const filterExtraWdioFrames = (error: Error): Error => {
...
+                "/node_modules/testplane/src/browser/stacktrace/",
+            ];
+
+            return Boolean(
+                frame.fileName && testplaneExtraInternalFramePaths.some(path => frame.fileName?.includes(path)),
+            );
+        };
+
+        const shouldIncludeFrame = (frame: StackFrame): boolean => {
+            if (isIgnoredWebdriverioFrame(frame) || isWdioUtilsFrame(frame)) {
```
sipayRT, line 199: why should we separate these checks?

KuznetsovRoman: One is webdriverio frame, and other is wdioUtils frame.
For webdriverio frames we only ignore some but not all frames
I think its just easier to read when we have multiple structured conditions instead of one function with lots of "||"

PR #1009, opened 2024-09-05
chore(examples): add example with DnD

sipayRT:


examples/drag-and-drop/.testplane.conf.ts
```diff
@@ -0,0 +1,41 @@
...
+            // https://github.com/gemini-testing/html-reporter
+            enabled: true,
+            path: "testplane-report",
+            defaultView: "all",
+            diffMode: "3-up-scaled"
+        }
+    },
+    devServer: {
+        command: "npm run server:dev"
+    },
```
KuznetsovRoman, line 40: I suggest to pass SERVER_PORT as environment variable here:
```js
    devServer: {
        command: "npm run server:dev",
        env: { PORT: SERVER_PORT },
    },
```
instead of passing it via npm script

KuznetsovRoman: And i think you could use `npx http-server ./static/` as command

sipayRT: in this case you will not be able to run the dev-server separately with the same port (8080 by default)

PR #1019, opened 2024-09-19
fix: stop REPL servers on call "halt" method

DudaGod:
## What is done

In the case when the user switches to REPL mode and then someone calls `halt` method, then Testplane does not end correctly. In this case, it is necessary to explicitly close the repl server in worker. So I did it.

### DIFF
--- a/src/worker/index.js
+++ b/src/worker/index.js
@@ -8,3 +8,7 @@ testplaneFacade.init();
 exports.runTest = (fullTitle, options) => {
     return testplaneFacade.runTest(fullTitle, options);
 };
+
+exports.cancel = () => {
+    return testplaneFacade.cancel();
+};
--- a/src/worker/testplane-facade.ts
+++ b/src/worker/testplane-facade.ts
@@ -31,6 +31,10 @@ module.exports = class TestplaneFacade {
         return this.promise;
     }
 
+    cancel(): void {
+        RuntimeConfig.getInstance().replServer?.close();
+    }
+
     syncConfig(): Promise<void> {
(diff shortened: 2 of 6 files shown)


src/worker/index.js
```diff
@@ -8,3 +8,7 @@ testplaneFacade.init();
 exports.runTest = (fullTitle, options) => {
     return testplaneFacade.runTest(fullTitle, options);
 };
+
+exports.cancel = () => {
```
DudaGod, line 12: I can't stop repl server from master process. So I need to send command to all workers.

src/worker/testplane-facade.ts
```diff
@@ -31,6 +31,10 @@ module.exports = class TestplaneFacade {
         return this.promise;
     }
 
+    cancel(): void {
+        RuntimeConfig.getInstance().replServer?.close();
```
DudaGod, line 35: If `replServer` doesn't exist (user do not use `--repl` option) then nothing will not fail here.

test/src/runner/index.js
```diff
@@ -694,5 +695,16 @@ describe("NodejsEnvRunner", () => {
             assert.notCalled(BrowserRunner.prototype.run);
             assert.notCalled(BrowserRunner.prototype.cancel);
         });
+
+        it.only("should cancel all executing workers", async () => {
```
sipayRT, line 699: only

sipayRT (approved):

test/src/runner/index.js
DudaGod: wow, looks like a commit from my cat. I will remove it!

PR #1021, opened 2024-10-03
chore: add contribution notice

sipayRT:


KuznetsovRoman (approved):
Contributor would have to read this contributing guide. Would be nice if we could somehow automate this, so contributor would only have to check the checkbox "I have read cla(link)"

PR #1023, opened 2024-10-07
fix(component-testing): correctly import components with css from third party libraries

DudaGod:
### What is done

Some libraries can load css files when you import them in component test file. For example:
```
// test.testplane.tsx
import {Button} from "some-component-library";

describe("suite", {...});
```

This file executes not only in browser, but also in nodejs environment. Where I get an error like: `Unexpected token ...` when try to import it. To solve this problem we use pirates + babel, they remove unnecessary imports. But when user import some module (without file extension) we can't figure out if it needs to be ignored in node env or not.

So in order to fix this problem we first require this module and if we get `Unexpected token ...` error from css file then we remove it.

### DIFF
--- a/src/bundle/test-transformer.ts
+++ b/src/bundle/test-transformer.ts
@@ -2,9 +2,13 @@ import * as nodePath from "node:path";
 import * as babel from "@babel/core";
 import { addHook } from "pirates";
 import { TRANSFORM_EXTENSIONS, JS_EXTENSION_RE } from "./constants";
+import { requireModuleSync } from "../utils/module";
 
 import type { NodePath, PluginObj, TransformOptions } from "@babel/core";
 import type { ImportDeclaration } from "@babel/types";
+
+const STYLE_EXTESTION_RE = /\.(css|less|scss|sass|styl|stylus|pcss)$/;
+const IGNORE_STYLE_ERRORS = ["Unexpected token"];
 
 export const setupTransformHook = (opts: { removeNonJsImports?: boolean } = {}): VoidFunction => {
     const transformOptions: TransformOptions = {
@@ -34,6 +38,15 @@ export const setupTransformHook = (opts: { removeNonJsImports?: boolean } = {}):
 
                 if (extname && !extname.match(JS_EXTENSION_RE)) {
                     path.remove();
+                    return;
+                }
+
+                try {
+                    requireModuleSync(path.node.source.value);
+                } catch (err) {
+                    if (shouldIgnoreImportError(err as Error)) {
+                        path.remove();
+                    }
                 }
             },
         },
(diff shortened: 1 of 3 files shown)


src/bundle/test-transformer.ts
```diff
@@ -2,10 +2,14 @@ import * as nodePath from "node:path";
 import * as babel from "@babel/core";
 import { addHook } from "pirates";
 import { TRANSFORM_EXTENSIONS, JS_EXTENSION_RE } from "./constants";
+import { requireModuleSync } from "../utils/module";
```
DudaGod, line 5: Helper in order to be able to test `require`

src/bundle/test-transformer.ts
```diff
@@ -2,10 +2,14 @@ import * as nodePath from "node:path";
...
 
 import type { NodePath, PluginObj, TransformOptions } from "@babel/core";
 import type { ImportDeclaration } from "@babel/types";
 
+const STYLE_EXTESTION_RE = /\.(css|less|scss)$/;
+const IGNORE_IMPORT_ERRORS = ["Unexpected token"];
```
DudaGod: When style files imported in node js environment I get an error like: `Unexpected token .` (when first line in css is `.some-class {...}` or `Unexpected token {` (when first line is `body {...}`)

src/bundle/test-transformer.ts
```diff
@@ -52,3 +65,19 @@ export const setupTransformHook = (opts: { removeNonJsImports?: boolean } = {}):
...
+        return (err as Error).message.startsWith(ignoreImportErr);
+    });
+
+    if (!shouldIgnoreImport) {
+        return false;
+    }
+
+    const firstStackFrame = (err as Error).stack?.split("\n")[0] || "";
+    const filePath = firstStackFrame.split(":")[0];
+    const isStyleFilePath = STYLE_EXTESTION_RE.test(filePath);
```
DudaGod, line 80: When I get an error from import css first frame in stack looks like: `/Users/.../some.css:100500`.

src/bundle/test-transformer.ts
```diff
@@ -34,6 +38,15 @@ export const setupTransformHook = (opts: { removeNonJsImports?: boolean } = {}):
...
                     path.remove();
+                    return;
+                }
+
+                try {
+                    requireModuleSync(path.node.source.value);
+                } catch (err) {
+                    if (shouldIgnoreImportFromStyleFile(err as Error)) {
+                        path.remove();
+                    }
```
KuznetsovRoman, line 49: I think function should be named "shouldIgnoreImportFromStyleFile" only if we are sure this is style file.

"shouldIgnoreImportFileError" or "shouldIgnoreImportError" is more suitable, in my opinion, because we check "is this style file" inside of the function

PR #1024, opened 2024-10-09
fix(component-testing): ability to import esm modules

DudaGod:
## What is done

In component testing user can import esm modules in the test file, because it executes in the browser, which support them. But test files also import in nodejs environment which supports only cjs. Because of this all esm only modules throw `ERR_REQUIRE_ESM` error in nodejs env.

So i implement replacing import esm module to variable with proxy. Like this:
```
// From:
import esmPkg from "esm-module";
// To:
const esmPkg = new Proxy({}, {
    get: function (target, prop) {
        return prop in target ? target[prop] : new Proxy(() => {}, this);
    },
    apply: function () {
        return new Proxy(() => {}, this);
    }
});
```

This proxy gives ability to call every method or getter that user want.

Need review only second commit.

### DIFF
--- a/src/bundle/test-transformer.ts
+++ b/src/bundle/test-transformer.ts
@@ -5,7 +5,6 @@ import { TRANSFORM_EXTENSIONS, JS_EXTENSION_RE } from "./constants";
 import { requireModuleSync } from "../utils/module";
 
 import type { NodePath, PluginObj, TransformOptions } from "@babel/core";
-import type { ImportDeclaration } from "@babel/types";
 
 const STYLE_EXTESTION_RE = /\.(css|less|scss|sass|styl|stylus|pcss)$/;
 const IGNORE_STYLE_ERRORS = ["Unexpected token"];
@@ -30,10 +29,10 @@ export const setupTransformHook = (opts: { removeNonJsImports?: boolean } = {}):
         ],
     };
 
-    const customIgnoreImportsPlugin = (): PluginObj => ({
+    const customIgnoreImportsPlugin = ({ types: t }: { types: typeof babel.types }): PluginObj => ({
         name: "ignore-imports",
         visitor: {
-            ImportDeclaration(path: NodePath<ImportDeclaration>): void {
+            ImportDeclaration(path: NodePath<babel.types.ImportDeclaration>): void {
                 const extname = nodePath.extname(path.node.source.value);
 
                 if (extname && !extname.match(JS_EXTENSION_RE)) {
@@ -46,6 +45,12 @@ export const setupTransformHook = (opts: { removeNonJsImports?: boolean } = {}):
                 } catch (err) {
                     if (shouldIgnoreImportError(err as Error)) {
                         path.remove();
+                        return;
+                    }
+
+                    if ((err as NodeJS.ErrnoException).code === "ERR_REQUIRE_ESM") {
+                        mockEsmModuleImport(t, path);
+                        return;
                     }
                 }
             },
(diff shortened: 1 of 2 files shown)


src/bundle/test-transformer.ts
```diff
@@ -52,3 +70,127 @@ export const setupTransformHook = (opts: { removeNonJsImports?: boolean } = {}):
...
+    return isStyleFilePath;
+}
+
+/*
+Replace esm module import with a Proxy.
+Examples
+1) `import pkg from "package"` -> `const pkg = new Proxy({}, {get: ..., apply: ...})`
+2) `import {a, b as c} from "package"` -> `const {a, c} = new Proxy({}, {get: ..., apply: ...})`
+*/
+function mockEsmModuleImport(t: typeof babel.types, path: NodePath<babel.types.ImportDeclaration>): void {
```
DudaGod, line 96: Used method from here - https://babeljs.io/docs/babel-types.
In order to implement it I did use ast explorer - https://astexplorer.net/.

src/bundle/test-transformer.ts
```diff
@@ -52,3 +70,127 @@ export const setupTransformHook = (opts: { removeNonJsImports?: boolean } = {}):
...
+Examples:
+1) `import pkg from "package"` -> `pkg`
+2) `import {a, b as c} from "package"` -> `const {a, с} `
+*/
+function genVarDeclKey(
+    t: typeof babel.types,
+    node: NodePath<babel.types.ImportDeclaration>["node"],
+): babel.types.Identifier | babel.types.ObjectPattern {
+    if (node.specifiers.length === 1) {
+        if (["ImportDefaultSpecifier", "ImportNamespaceSpecifier"].includes(node.specifiers[0].type)) {
```
DudaGod, line 119: `ImportDefaultSpecifier` -> `import pkg from '...'`,
`ImportNamespaceSpecifier` -> `import * as pkg from '...'`

src/bundle/test-transformer.ts
```diff
@@ -52,3 +70,127 @@ export const setupTransformHook = (opts: { removeNonJsImports?: boolean } = {}):
...
+            return t.identifier(node.specifiers[0].local.name);
+        }
+
+        return t.objectPattern([
```
DudaGod, line 123: Here support case like:
```
import {a, b, c as d} from "..."
```

src/bundle/test-transformer.ts
```diff
@@ -52,3 +70,127 @@ export const setupTransformHook = (opts: { removeNonJsImports?: boolean } = {}):
...
+            t.objectProperty(
+                t.identifier(node.specifiers[0].local.name),
+                t.identifier(node.specifiers[0].local.name),
+                false,
+                true,
```
DudaGod, line 128: `shorthand`, it means use `{a}` instead of `{a: a}`

test/src/test-reader/test-transformer.ts
```diff
@@ -59,29 +60,165 @@ describe("test-transformer", () => {
...
+                (pirates.addHook as SinonStub).callsFake(cb => {
+                    transformedCode = cb(`import "${moduleName}"`, moduleName);
+                });
+
+                setupTransformHook({ removeNonJsImports: true });
+
+                assert.match(transformedCode, new RegExp(`require\\("${moduleName}"\\)`));
+            });
+
+            describe.only("replace import of esm module with a proxy", () => {
```
DudaGod, line 140: todo: remove it

src/bundle/test-transformer.ts
```diff
@@ -46,6 +45,12 @@ export const setupTransformHook = (opts: { removeNonJsImports?: boolean } = {}):
                 } catch (err) {
                     if (shouldIgnoreImportError(err as Error)) {
                         path.remove();
+                        return;
+                    }
+
+                    if ((err as NodeJS.ErrnoException).code === "ERR_REQUIRE_ESM") {
+                        mockEsmModuleImport(t, path);
+                        return;
```
KuznetsovRoman, line 53: Should you return here?
Doesn't it make it work only once?

test/src/test-reader/test-transformer.ts
```diff
@@ -136,6 +136,90 @@ describe("test-transformer", () => {
...
+                const expectedProxyValue =
+                    "" +
+                    `new Proxy({}, {\n` +
+                    `  get: function (target, prop) {\n` +
+                    `    return prop in target ? target[prop] : new Proxy(() => {}, this);\n` +
+                    `  },\n` +
+                    `  apply: function () {\n` +
+                    `    return new Proxy(() => {}, this);\n` +
+                    `  }\n` +
+                    `});\n`;
```
KuznetsovRoman, line 152: nit: I would write it like this:

```js
const expectedProxyValue = [
    "",
    "new Proxy({}, {\",
    "    get: function (target, prop) {",
    ...
].join("\n")
```
just so there would be no need for manual "\n" at the end of each line

KuznetsovRoman (approved):

PR #1027, opened 2024-10-25
feat(component-testing): implement mocks

DudaGod:
### What is done:

Implement ability to mock modules and use spies in component testing. The same as in jest.

### How it works

I'll explain it with a simple example:

```
// tests/test.testplane.tsx
import { render } from '@testing-library/react';
import App from '../App';
// testplane/mock - new file which is exports methods from `@vitest/spy` and custom methods: `mock`, `unmock`. 
// These methods works only in browser env;
import {fn, mock} from "testplane/mock"; 
import { handleClick } from './utils';

mock('./utils', () => ({
    handleClick: fn().mockImplementationOnce(() => {
        console.log('Handle click by mock');
    })
}));

it('should render react button', async ({browser}) => {
    render(
        <App />
    );

    await browser.$('input').click();

    expect(handleClick).toHaveBeenCalledTimes(1);
});
```

```
// App.tsx
import {handleClick} from "./tests/utils";

export default function App() {
    return (
        <div id="root">
            <input
                type="button"
                value="Some button"
                onClick={handleClick}
            />
        </div >
    );
}
```

```
// utils.ts
export const handleClick = (): void => {
    console.log('Handle click by user code');
}
```

In order to correctly mock `handleClick` method from `utils.ts` source code of test file and dependencies which used mocked module modified in runtime. Test file will looks like:
```
// tests/test.testplane.tsx
import * as __testplane_import_1__ from './utils';
import {fn, mock} from "testplane/mock";

await mock('./utils', () => ({
    handleClick: fn().mockImplementationOnce(() => {
        console.log('Handle click by mock');
    })
}), __testplane_import_1__);

const { render } = await import('@testing-library/react');
const App = await import('../App');
const { handleClick } = importWithMock('./utils', __testplane_import_1__);

// ...
```

What happens:
- import of original module `utils` moved to the top and saves to generated variable - `__testplane_import_1__`. It is used in order to correctly mock module (for example if user mock only on method);
- import of `fn` and `mock` also moved to the top in order to call `mock` before import all other modules;
- call `mock` with `await` and send original module as third argument. Moreover move it to the top. Here used `mock` from `src/runner/browser-env/vite/browser-modules/mock.ts` which is async. 
- import all other modules using dynamic imports. It is necessary in order to correctly mock module and then import all other deps in which mocked module can be imported. Without dynamic import modules will imported first and then `await mock` will be executed;
- for mocked module use helper `importWithMock` which gets mocked implementation of module instead of original module.

Component file will looks like (after modify in runtime):
```
import * as __testplane_import_2__ from './tests/utils';
const {handleClick} = importWithMock("./tests/utils", __testplane_import_2__);

export default function App() {
    // ...
}
```

What happens:
- like in test file we save original module to special variable - `__testplane_import_2__`;
- import mocked module using helper `importWithMock`

### Discuss:
- Should rename new exported file `mock` for something more understandable? This file not only export `mock` and `unmock` methods but also spied: `fn`, `spyOn`. Moreover this module works only in browser environment. And can't be called in node env.


package.json
```diff
@@ -34,6 +33,20 @@
     "type": "git",
     "url": "git://github.com/gemini-testing/testplane.git"
   },
+  "exports": {
+    ".": "./build/src/index.js",
+    "./mock": "./build/src/mock/index.js"
+  },
+  "typesVersions": {
```
DudaGod, line 40: In order to correctly set paths to types

src/mock/vitest-spy.ts
```diff
@@ -0,0 +1,580 @@
+// TODO: use @vitest/spy when migrate to esm
```
DudaGod, line 1: This file copied as is from @vitest/spy and used in nodejs environment. In browser environment this file replaced to special file which works only in browser env.

Copied because module is esm and I can't use it correctly in cjs world.

src/runner/browser-env/vite/utils.ts
```diff
@@ -27,3 +39,39 @@ export const getNodeModulePath = async ({
 export const prepareError = (error: Error): Error => {
     return JSON.parse(JSON.stringify(error, Object.getOwnPropertyNames(error)));
 };
+
+export const getTestInfoFromViteRequest = (req: Connect.IncomingMessage): TestInfoFromViteRequest | null => {
+    if (!req.url?.endsWith("index.html") || !req.originalUrl) {
+        return null;
+    }
```
KuznetsovRoman, line 46: How it behaves with query(`"?"`) or hash location(`"#"`) parameters?
And what is the difference between "url" and "originalUrl"?

src/runner/browser-env/vite/utils.ts
```diff
@@ -27,3 +39,39 @@ export const getNodeModulePath = async ({
...
+};
+
+export const getPathWithoutExtName = (fsPath: string): string => {
+    const extname = path.extname(fsPath);
+
+    if (!extname) {
+        return fsPath;
+    }
+
+    return fsPath.slice(0, extname.length * -1);
```
KuznetsovRoman, line 76: Why `extname.length * -1` and not `-extname.length`? `extname.length * -1` looks weird

src/utils/browser.ts
```diff
@@ -5,3 +5,7 @@ export const isSupportIsolation = (browserName: string, browserVersion = ""): bo
 
     return browserName === "chrome" && Number(browserVersionMajor) >= MIN_CHROME_VERSION_SUPPORT_ISOLATION;
 };
+
+export const isRunInBrowserEnv = (): boolean => {
+    return typeof window !== "undefined" && typeof window.document !== "undefined";
+};
```
KuznetsovRoman, line 11: What if i mocked "globalThis.window" in node.js environment?
There are more reliable checks. I like this one, with `global` and `window` equality to `this` (i think, we better should compare with `globalThis`): https://stackoverflow.com/a/31090240

src/runner/browser-env/vite/utils.ts
DudaGod: > How it behaves with query("?") or hash location("#") parameters?

It works correctly with all of these parameters. Because in one of my plugin I return only `index.html`. It means url inside `req.url` always has `/index.html`. Url inside `originalUrl` has url which is open in test. Looks like - `/run-uuids/1d344b36-a09a-4cc2-9ff9-158e91fa70a1`

src/utils/browser.ts
DudaGod: I realized today that it doesn't work correctly. It means I can't realize is `mock` method called in node env or browser if it is used outside of hook/suite/test. Example:
```
import {mock} from "testplane";

mock('axios');

it('foo bar', () => {...});
```

This file will be read in master and worker process, before it will be run in browser env. It means that `mock` will be called in test parsing stage. So I need knowledge from config in which env it starts. But it means I need add this logic to the global.

We already had `global.testplane`, but I don't want to pollute to this variable. So right now I remove this checker and `mock` and `unmock` calls to nothing in node env.

I will create ticket about this problem.

PR #1029, opened 2024-11-13
feat: auto browser download

KuznetsovRoman:
## What is done
Added ability to automatically download browsers and/or webdrivers for these browsers.

Supported browsers to download:
- chrome > 73
- firefox

Supported drivers to download:
- chromedriver (for chrome)
- geckodriver (for firefox)
- edgedriver (for edge)

With "--local" (or gridUrl: "local") and webdriver protocol, webdiver instances are automatically run for these browsers:
- chrome (>73)
- firefox
- edge (if the browser itself is installed. Driver will be installed automatically)
- safari (mac only)

When launching tests with "--local" (and optionally webdriver protocol) all supported browsers/drivers will be installed automatically, if necessary. With "--local" we at all cost are trying to avoid network requests to google/mozilla/... servers.

## Example

- Set up a testplane config, using webdriver protocol and describe some chrome/firefox browsers with defined versions
- Launch eigher using `npx testplane --local`, either by setting "gridUrl": "local"

In order to just download browser, `npx testplane install-browsers` could be used. This way we also check if our browsers are up-to-date, so if we have 114.0.5696.0 downloaded and user wants chrome@114.0, we check for newest chrome@114.0 version and see it is "114.0.5735.133". Then this new version is downloaded
- If used without arguments, all of the config browsers will be installed
- If argument is browserId from config, the corresponding browser will be installed
- If argument is <browserName>@<browserVersion>, it will be installed.
These can be combined: `npx testplane install-browsers my-chrome chrome@115`

Directory to save testplane browsers and drivers can be set with `TESTPLANE_BROWSERS_PATH` env variable. By default, `~/.testplane` is used.


package.json
```diff
@@ -54,6 +54,7 @@
     "@babel/code-frame": "7.24.2",
     "@gemini-testing/commander": "2.15.4",
     "@jspm/core": "2.0.1",
+    "@puppeteer/browsers": "2.4.0",
```
KuznetsovRoman, line 57: - download chrome
- download chromium (replaces chrome for versions >73, <115)
- download firefox

package.json
```diff
@@ -64,12 +65,16 @@
     "bluebird": "3.5.1",
     "chalk": "2.4.2",
     "clear-require": "1.0.1",
+    "cli-progress": "3.12.0",
```
KuznetsovRoman, line 68: multiple progress bar
image

package.json
```diff
@@ -64,12 +65,16 @@
...
     "debug": "2.6.9",
     "devtools": "8.39.0",
+    "edgedriver": "5.6.1",
```
KuznetsovRoman, line 71: - download edge driver

package.json
```diff
@@ -64,12 +65,16 @@
...
     "error-stack-parser": "2.1.4",
     "expect-webdriverio": "3.6.0",
+    "extract-zip": "2.0.1",
```
KuznetsovRoman, line 74: - extract zip archieve with chromedriver, which is installed manually (>73, <115)

package.json
```diff
@@ -64,12 +65,16 @@
...
     "fastq": "1.13.0",
     "fs-extra": "5.0.0",
+    "geckodriver": "4.5.0",
```
KuznetsovRoman, line 77: - download and run geckodriver

package.json
```diff
@@ -93,6 +98,7 @@
     "urijs": "1.19.11",
     "url-join": "4.0.1",
     "vite": "5.1.6",
+    "wait-port": "1.1.0",
```
KuznetsovRoman, line 101: - wait for port to be ready to receive requests when launching webdriver on this port

src/browser-installer/chromium/revisions/linux.ts
```diff
@@ -0,0 +1,42 @@
...
+    104: 1012822,
+    105: 1027072,
+    106: 1036920,
+    107: 1047812,
+    108: 1059082,
+    109: 1070158,
+    110: 1084167,
+    111: 1097778,
+    112: 1107206,
+} as Record<number, number>;
```
KuznetsovRoman, line 42: I add `as Record<number, number>` so `.d.ts` would see this file as "just map" and don't save each value as constant

src/browser-installer/chromium/utils.ts
```diff
@@ -0,0 +1,36 @@
...
+        mac_arm: "mac64_m1", // eslint-disable-line camelcase
+        win32: "win32",
+        win64: "win32",
+    };
+
+    const milestone = getMilestone(version);
+    const platform = getChromePlatform(version);
+    const isNewMacArm =
+        platform === BrowserPlatform.MAC_ARM && Number(milestone) >= MIN_CHROMEDRIVER_MAC_ARM_NEW_ARCHIVE_NAME;
+    const archiveName = isNewMacArm ? "mac_arm64" : chromeDriverArchiveName[platform];
```
KuznetsovRoman, line 26: Starting at chromium@106, chromium driver started shipping mac arm builds with "mac_arm64" prefix instead of old "mac64_m1"

src/browser-installer/edge/driver.ts
```diff
@@ -0,0 +1,51 @@
...
+    const fullVersion = await retryFetch(`${MSEDGEDRIVER_API}/LATEST_RELEASE_${milestone}`).then(res => res.text());
+
+    if (!fullVersion) {
+        throw new Error(`Couldn't resolve latest edgedriver version for ${milestone}`);
+    }
+
+    const versionNormalized = fullVersion
+        .split("")
+        .filter(char => /\.|\d/.test(char))
+        .join("");
```
KuznetsovRoman, line 23: This API resolves string with null characters for each odd character

src/browser-installer/firefox/utils.ts
```diff
@@ -0,0 +1,13 @@
+const firefoxChannels = ["stable", "nightly"];
+
+export const normalizeFirefoxVersion = (version: string): string => {
+    return version.includes(".") ? version : `${version}.0`;
+};
```
KuznetsovRoman, line 5: `@puppeteer/browsers` can't handle browser versions in format "115", so we add ".0" suffix if it isn't exist. That way it works

src/browser-installer/install.ts
```diff
@@ -0,0 +1,112 @@
+import _ from "lodash";
+
+/**
+ * @returns path to browser binary
+ */
+export const installBrowser = async (
```
KuznetsovRoman, line 6: Install browser (maybe with webdriver), return browser binary path

src/browser-installer/registry/index.ts
```diff
@@ -0,0 +1,215 @@
...
+    type SupportedBrowser,
+    type SupportedDriver,
+    type DownloadProgressCallback,
+} from "../utils";
+import { getFirefoxBuildId } from "../firefox/utils";
+import logger from "../../utils/logger";
+import type { createBrowserDownloadProgressBar } from "./cli-progress-bar";
+
+type VersionToPathMap = Record<string, string | Promise<string>>;
+type BinaryName = Exclude<SupportedBrowser | SupportedDriver, SupportedBrowser & SupportedDriver>;
```
KuznetsovRoman, line 21: That way if `SupportedBrowser` and `SupportedDriver` would have intersection, we wouldn't be able to use binary name neither as browser name, nor as driver name

src/browser-installer/registry/index.ts
```diff
@@ -0,0 +1,215 @@
...
+type RegistryKey = `${BinaryName}_${BrowserPlatform}`;
+type Registry = Record<RegistryKey, VersionToPathMap>;
+
+const registryPath = getRegistryPath();
+const registry: Registry = existsSync(registryPath) ? readJsonSync(registryPath) : {};
```
KuznetsovRoman, line 26: Registry is:
- Singletone
- Sync

So its state is valid and up-to-date all the time.
It acts like synchronization point  between multiple browsers. So when multiple browsers (possibly with different browserId) with same browserName and browserVersion would want to install the browser, `install` function only being called one time, and other browsers (NewBrowser instance) receive same promise, which will be resolved to binary path when download is complete

src/browser-installer/registry/index.ts
```diff
@@ -0,0 +1,215 @@
...
+
+        return suitableBuildIdsSorted[suitableBuildIdsSorted.length - 1];
+    }
+
+    if (driverName === Driver.GECKODRIVER) {
+        const buildIds = Object.keys(registry[registryKey]);
+        const buildIdsSorted = buildIds.sort(semverVersionsComparator);
+
+        return buildIdsSorted.length ? buildIdsSorted[buildIdsSorted.length - 1] : null;
+    }
```
KuznetsovRoman, line 118: geckodriver is compatible with all browser versions, so we just pick the most recent version for firefox

src/browser-installer/registry/index.ts
```diff
@@ -0,0 +1,215 @@
...
+
+    if (!suitableBuildIds.length) {
+        return null;
+    }
+
+    const firefoxVersionComparator = (a: string, b: string): number => {
+        a = a.slice(a.indexOf("_") + 1);
+        b = b.slice(b.indexOf("_") + 1);
+
+        return parseInt(a.replace(".", ""), 16) - parseInt(b.replace(".", ""), 16);
```
KuznetsovRoman, line 164: firefox has versions like "stable_131.0a1", "stable_129.0b9" and such. Parsing them without channel prefix ("stable_" or "nightly_") as hex number works

src/browser-installer/utils.ts
```diff
@@ -0,0 +1,187 @@
...
+    return platform;
+};
+
+export const getChromePlatform = (version: string): BrowserPlatform => {
+    const milestone = getMilestone(version);
+    const platform = getBrowserPlatform();
+
+    if (platform === BrowserPlatform.MAC_ARM && Number(milestone) < MIN_CHROMIUM_MAC_ARM_VERSION) {
+        return BrowserPlatform.MAC;
+    }
```
KuznetsovRoman, line 113: for chrome, use `mac64` platform, if chromium version is old and `mac_arm` version does not exist

src/browser-installer/utils.ts
```diff
@@ -0,0 +1,187 @@
...
+export const getEdgeDriverDir = (driverVersion: string): string =>
+    getDriverDir("edgedriver", getBrowserPlatform() + "-" + driverVersion);
+export const getChromiumDriverDir = (driverVersion: string): string =>
+    getDriverDir("chromedriver", getBrowserPlatform() + "-" + driverVersion);
+export const getChromeDriverDir = (): string => getDriversDir(); // path is set by @puppeteer/browsers.install
+
+export const getFirefoxBrowserDir = (): string => getBrowsersDir(); // path is set by @puppeteer/browsers.install
+export const getChromeBrowserDir = (): string => getBrowsersDir(); // path is set by @puppeteer/browsers.install
+
+export const retryFetch = async (
```
KuznetsovRoman, line 144: As we fetch some data from github, retries wont hurt

src/browser-pool/basic-pool.ts
```diff
@@ -26,10 +28,11 @@ export class BasicPool implements Pool {
 
         this._activeSessions = {};
         this._cancelled = false;
+        this._wdPool = new WebdriverPool();
```
KuznetsovRoman, line 31: all browsers of the same browser-pool share WebdriverPool

src/browser-pool/webdriver-pool.ts
```diff
@@ -0,0 +1,99 @@
+import type { ChildProcess } from "child_process";
+import { runBrowserDriver, getDriverNameForBrowserName } from "../browser-installer";
+import type { SupportedBrowser, SupportedDriver } from "../browser-installer";
+
+type DriverVersion = string;
+type Port = string;
+type ChildProcessWithStatus = { process: ChildProcess; gridUrl: string; isBusy: boolean };
+export type WdProcess = { gridUrl: string; free: () => void; kill: () => void };
+
+export class WebdriverPool {
```
KuznetsovRoman, line 10: We have N browsers and K webdriverds.
This class handles webdriver processes:
- when we can reuse existing webdriver process, which is free
- when we need to launch new webdriver process

src/browser/new-browser.ts
```diff
@@ -77,13 +83,17 @@ export class NewBrowser extends Browser {
         try {
             this.setHttpTimeout(this._config.sessionQuitTimeout);
             await this._session!.deleteSession();
+            this._wdProcess?.free();
         } catch (e) {
             warn(`WARNING: Can not close session: ${(e as Error).message}`);
+            this._wdProcess?.kill();
```
KuznetsovRoman, line 89: If session can't be closed, we kill webdriver process, so later we could launch the new one

src/browser/new-browser.ts
```diff
@@ -108,10 +118,26 @@ export class NewBrowser extends Browser {
...
+
+        let gridUrl;
+
+        if (this._isLocalGridUrl() && config.automationProtocol === "webdriver") {
+            gridUrl = await this._getLocalWebdriverGridUrl();
+        } else {
+            // if automationProtocol is not "webdriver", fallback to default grid url from "local"
+            // because in "devtools" protocol we dont need gridUrl, but it still has to be valid URL
+            gridUrl = config.gridUrl === LOCAL_GRID_URL ? DEFAULT_GRID_URL : config.gridUrl;
+        }
```
KuznetsovRoman, line 136: "isLocalGridUrl" === either "gridUrl" is "local", either `testplane` is launched with `--local`
- If isLocalGridUrl == true and automationProtocol === "webdriver", we get webdriver process and use its gridUrl
- if isLocalGridUrl == true and automationProtocol is "devtools", gridUrl won't be used either way, but we need to pass any valid url

src/browser/new-browser.ts
```diff
@@ -133,15 +159,18 @@ export class NewBrowser extends Browser {
...
             ? this._extendCapabilitiesByVersion()
             : config.desiredCapabilities;
         const capabilitiesWithAddedHeadless = this._addHeadlessCapability(
             config.headless,
             capabilitiesExtendedByVersion!,
         );
-        return capabilitiesWithAddedHeadless;
+
+        return this._isLocalGridUrl()
+            ? this._addExecutablePath(config, capabilitiesWithAddedHeadless)
```
KuznetsovRoman, line 172: Here we don't check for automationProtocol, so we could use our binaries for "devtools" automation protocol

src/browser/new-browser.ts
```diff
@@ -175,6 +204,43 @@ export class NewBrowser extends Browser {
...
+
+    protected async _addExecutablePath(
+        config: BrowserConfig,
+        capabilities: WebdriverIO.Capabilities,
+    ): Promise<WebdriverIO.Capabilities> {
+        const browserNameLowerCase = config.desiredCapabilities?.browserName?.toLowerCase() as string;
+        const executablePath = await installBrowser(
+            this._config.desiredCapabilities?.browserName as SupportedBrowser,
+            this._config.desiredCapabilities?.browserVersion as string,
+        );
```
KuznetsovRoman, line 233: Installs browser (if necessary) and returns its executable path
For safari and edge `null` is returned (we won't set binary path in capabilities, but wdio will be able to find them, if they are installed)

src/browser-installer/registry/index.ts
KuznetsovRoman: It looks like this:
```json
{
    "chrome_mac_arm": {
        "114.0.5735.133": "../browsers/chrome/mac_arm-114.0.5735.133/chrome-mac-arm64/Google Chrome for Testing.app/Contents/MacOS/Google Chrome for Testing",
        "115.0.5790.170": "../browsers/chrome/mac_arm-115.0.5790.170/chrome-mac-arm64/Google Chrome for Testing.app/Contents/MacOS/Google Chrome for Testing"
    },
    "chromedriver_mac_arm": {
        "110.0.5481.77": "../drivers/chromedriver/mac_arm-110.0.5481.77/chromedriver",
        "114.0.5735.90": "../drivers/chromedriver/mac_arm-114.0.5735.90/chromedriver",
        "115.0.5790.170": "../drivers/chromedriver/mac_arm-115.0.5790.170/chromedriver-mac-arm64/chromedriver",
        "92.0.4515.107": "../drivers/chromedriver/mac_arm-92.0.4515.107/chromedriver"
    },
    "chromium_mac": {
        "92": "../browsers/chromium/mac-885361/chrome-mac/Chromium.app/Contents/MacOS/Chromium"
    },
    "chromium_mac_arm": {
        "110": "../browsers/chromium/mac_arm-1084160/chrome-mac/Chromium.app/Contents/MacOS/Chromium"
    },
    "firefox_mac_arm": {
        "stable_115.0": "../browsers/firefox/mac_arm-stable_115.0/Firefox.app/Contents/MacOS/firefox"
    },
    "geckodriver_mac_arm": {
        "0.35.0": "../drivers/geckodriver/mac_arm-0.35.0/geckodriver"
    }
}
```
Here we install "mac" version for chromium@92 and "mac_arm" for chromium@110, because there is no "mac_arm" version for chromium@92, even when there is chromedriver@92 for "mac_arm"

src/browser-pool/webdriver-pool.ts
KuznetsovRoman: `port` in this class acts like process id as it is unique

test/src/browser-installer/chrome/index.ts
```diff
@@ -0,0 +1,86 @@
...
+
+    beforeEach(() => {
+        pipeLogsWithPrefixStub = sandbox.stub();
+        installChromeStub = sandbox.stub().resolves("/browser/path");
+        installChromeDriverStub = sandbox.stub().resolves("/driver/path");
+        spawnStub = sandbox.stub().returns({ kill: sandbox.stub() });
+        getPortStub = sandbox.stub().resolves(12345);
+        waitPortStub = sandbox.stub().resolves();
+
+        runChromeDriver = proxyquire("../../../../src/browser-installer/chrome", {
```
sipayRT, line 27: it might be better to test this functionality with integration tests. There are too many stubs right now

test/src/browser-installer/run.ts
```diff
@@ -0,0 +1,56 @@
...
+            "./chrome": { runChromeDriver: runChromeDriverStub },
+            "./edge": { runEdgeDriver: runEdgeDriverStub },
+            "./firefox": { runGeckoDriver: runGeckoDriverStub },
+            "./safari": { runSafariDriver: runSafariDriverStub },
+        }).runBrowserDriver;
+    });
+
+    afterEach(() => sandbox.restore());
+
+    [true, false, undefined].forEach(debug => {
```
sipayRT, line 31: o_O

test/src/cli/commands/install-browsers/index.ts
```diff
@@ -0,0 +1,161 @@
...
+        sandbox.stub(Testplane, "create").returns(testplaneStub as Testplane);
+
+        sandbox.stub(process, "exit");
+
+        sandbox.spy(Command.prototype, "action");
+
+        installBrowsersWithDriversStub = sandbox.stub();
+
+        cli = proxyquire("../../../../../src/cli", {
+            [path.resolve(process.cwd(), "src/cli/commands/install-browsers")]: proxyquire(
```
sipayRT, line 48: no, you can't do that. Double `proxyquire` nesting is too much :)

src/browser-installer/registry/index.ts
KuznetsovRoman: > can we solve this problem just with /(\w+)_(\d+)/ regexp?

I don't think so. "129.0b9" is greater than "129.0b7"

src/browser-installer/install.ts
```diff
@@ -0,0 +1,112 @@
...
+              ]).then(binaries => binaries[0])
+            : installFirefox(browserVersion, { force });
+    } else if (/edge/i.test(browserName)) {
+        const { installEdgeDriver } = await import("./edge");
+
+        if (installWebDriver) {
+            await installEdgeDriver(browserVersion, { force });
+        }
+
+        return null;
```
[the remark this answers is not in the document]
KuznetsovRoman: Because we can't download `edge` or `safari`. We return `null` so `NewBrowser` wouldn't set desiredCapabilities.
With `safari` and `edge` currently, as we are using globally installed browser version, we don't need to specify binaryPath. This will change after edge autodownload support though.

src/browser-installer/chrome/browser.ts
```diff
@@ -0,0 +1,61 @@
+import { resolveBuildId, canDownload, install as puppeteerInstall } from "@puppeteer/browsers";
+import { MIN_CHROME_FOR_TESTING_VERSION } from "../constants";
```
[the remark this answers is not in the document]
KuznetsovRoman: No.
We have these chrome versions:
- MIN_CHROMIUM_VERSION (73)
- MIN_CHROMIUM_MAC_ARM_VERSION (93)
- MIN_CHROME_FOR_TESTING_VERSION (113)

- [73, 92]. Minimal chrome to use. Chromium is used. (mac x64 version for mac arm, because there is no mac arm for these versions)
- [93, 112]. Chromium is still used, but now Mac_arm version for mac.
- 113+. Chrome for testing is used.

src/browser-installer/chrome/browser.ts
```diff
@@ -0,0 +1,61 @@
...
+    type DownloadProgressCallback,
+} from "../utils";
+import { getBinaryPath, getMatchingBrowserVersion, installBinary } from "../registry";
+import { normalizeChromeVersion } from "../utils";
+
+export const installChrome = async (version: string, { force = false } = {}): Promise<string> => {
+    const milestone = getMilestone(version);
+
+    if (Number(milestone) < MIN_CHROME_FOR_TESTING_VERSION) {
+        browserInstallerDebug(`couldn't install chrome@${version}, installing chromium instead`);
```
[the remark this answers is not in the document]
KuznetsovRoman: It is not user warning log. It won't be logged out to the user. It is debug log.

src/browser-installer/install.ts
```diff
@@ -0,0 +1,112 @@
...
+            `Couldn't install browser '${browserName}', as it is not supported`,
+            `Currently supported for installation browsers: 'chrome', 'firefox`,
+        ].join("\n"),
+    );
+
+    if (!browserName) {
+        throw unsupportedBrowserError;
+    }
+
+    if (!browserVersion) {
```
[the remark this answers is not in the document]
KuznetsovRoman: > `browserVersion` is not required param

Currently it is with "--local".
browser version `latest` is hard to manage, because it will be different between multiple developers in one project, so it is an anti-pattern.

We could think about installing "latest" by default, but with `npx testplane` it will be treated as "latest installed version" (i want `npx testplane --local` to fully work in offline), so it could use `N` version, when `N + 1` is available

test/src/browser-installer/chrome/index.ts
KuznetsovRoman: Yes, everything is stubbed, because it is not important at `chrome/index.ts` level
We only need to know `installChromeDriver` is called/not called here, as well as `spawn` is called/not called.

`installChromeDriver` is tested in another test file, and none of those stubs contains any important logic

I still think those tests do test `chrome/index.ts` functionality (and nothing more)

test/src/cli/commands/install-browsers/index.ts
KuznetsovRoman: Just copied it from here: https://github.com/gemini-testing/testplane/blob/master/test/src/cli/commands/list-tests/index.ts#L39-L50

Is there any better way?

src/cli/commands/install-browsers/index.ts
```diff
@@ -0,0 +1,96 @@
...
+                    browsers = Object.keys(testplane.config.browsers);
+                }
+
+                const browsersToInstall = browsers.map(browser => {
+                    if (browser in testplane.config.browsers) {
+                        const browserConfig = testplane.config.browsers[browser];
+                        const { browserName, browserVersion } = browserConfig.desiredCapabilities || {};
+
+                        return { browserName, browserVersion };
+                    } else if (browser.includes("@")) {
```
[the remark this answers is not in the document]
sipayRT: sorry, my fault. I mean that at first you need to check `chrome@100` and then check the version from the config

test/src/cli/commands/install-browsers/index.ts
sipayRT: yes - integration test ;)

src/browser-installer/utils.ts
```diff
@@ -0,0 +1,187 @@
...
+
+    return null as never;
+};
+
+export const downloadFile = async (url: string, filePath: string): Promise<void> => {
+    const writeStream = createWriteStream(filePath);
+    const response = await fetch(url);
+
+    if (!response.ok || !response.body) {
+        throw new Error(`Unable to download file from ${url}`);
```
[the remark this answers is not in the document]
KuznetsovRoman: Used statusText, as if there was an error like "offline" or "404", we would get it on "await fetch"

src/cli/commands/install-browsers/index.ts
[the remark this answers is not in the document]
KuznetsovRoman: Check `browser.includes("@")` is weak and user might use `npx testplane install-deps chrome@foobar` or `npx install foo@bar`, and he might have `browserId` `foo@bar`.

I don't think someone would actually use `@` in `browserId`, but if he does, and he typed `foo@bar`, he mean `install this browser` and not "browser `foo` of version `bar`"

test/src/browser-installer/chrome/driver.ts
```diff
@@ -0,0 +1,114 @@
...
+    it("should download driver if it is not downloaded", async () => {
+        getMatchingDriverVersionStub.withArgs(Driver.CHROMEDRIVER, sinon.match.string, "115").returns(null);
+        resolveBuildIdStub.withArgs(Driver.CHROMEDRIVER, sinon.match.string, "115").resolves("115.0.5678.170");
+        installBinaryStub
+            .withArgs(Driver.CHROMEDRIVER, sinon.match.string, "115.0.5678.170", sinon.match.func)
+            .resolves("/new/driver/path");
+
+        const driverPath = await installChromeDriver("115");
+
+        assert.equal(driverPath, "/new/driver/path");
```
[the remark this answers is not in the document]
KuznetsovRoman: Now its only 2 and in both cases we call "download" function

src/browser-installer/install.ts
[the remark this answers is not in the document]
KuznetsovRoman: I suggest to implement this in next PR

sipayRT (approved):

PR #1030, opened 2024-11-15
speed up the installation of dependencies

DudaGod:
### What is done:

- reinstall webdriverio, @wdio/utils and devtools in order to update locate-app instade package-lock, which was weight ~30mb. Now it weights 1.7 mb
- start to bundle `glob-extra` and `@wdio/utils-cjs` in order to reduce weight of packages and do not install them from npm
- remove @wdio/globals which not used anymore
- move @types packages from deps to dev deps

As a result:
- production node_modules weights on 31% less (217mb vs 314 mb)
- production deps installed on 16% faster (18.6 sec vs 22.32 sec). Check on 5 runs on master and current branch, use command - `time npm i --production` and `rm -rf node_modules && npm cache clean -f` before each install


src/base-testplane.ts
```diff
@@ -1,4 +1,4 @@
-import _ from "lodash";
+import _ from "./bundle/lodash";
```
KuznetsovRoman, line 1: Will we have to always use import from "./bundle/lodash"?
Can it be automatically handled at compile-time so all "import _ from 'lodash'" would be replaced with "import _ from './bundle/lodash'"?

src/bundle/@wdio-utils.ts
```diff
@@ -0,0 +1,5 @@
+/* eslint-disable @typescript-eslint/no-var-requires */
+const bundle = require("./cjs");
```
KuznetsovRoman, line 2: Can we not use require here?
If import does not work here, i think `const bundle = import("./cjs")` would.

src/test-reader/sets-builder/index.ts
```diff
@@ -1,7 +1,6 @@
 import path from "path";
-import * as globExtra from "glob-extra";
 import _ from "lodash";
-
+import * as globExtra from "../../bundle/glob-extra";
```
KuznetsovRoman, line 3: What part of the PR is responsible for creating "../../bundle/glob-extra"?
I only see "bundle/@wdio-utils" and "bundle/cjs"

DudaGod:
Oh, I forgot to send comments, which I wrote when publish this PR =(

PR #1033, opened 2024-11-27
fix: correctly specify types for "mochaOpts" and "patternsOnReject"

DudaGod:


src/config/types.ts
```diff
@@ -128,23 +128,16 @@ export interface ExpectOptsConfig {
...
+    ui?: string | ((suite: Mocha.Suite) => void);
 }
 
 export interface SystemConfig {
     debug: boolean;
     mochaOpts: MochaOpts;
     expectOpts: ExpectOptsConfig;
     ctx: { [name: string]: unknown };
-    patternsOnReject: Array<string>;
+    patternsOnReject: Array<string | RegExp>;
```
DudaGod, line 140: It can be specified as RegExp. Like this:
```
patternsOnReject: [/foo/i, /bar/, /baz/g]
```

Under the hood we wrap all items from `patternsOnReject` to RegExp. Like this - `new RegExp(pattern)`. So it works correctly if user specify regexp value.

src/config/types.ts
```diff
@@ -128,23 +128,16 @@ export interface ExpectOptsConfig {
...
 
-export interface MochaOpts {
-    /** milliseconds to wait before considering a test slow. */
-    slow?: number;
-
-    /** timeout in milliseconds or time string like '1s'. */
-    timeout?: number;
-
-    /** string or regexp to filter tests with. */
-    grep?: string | RegExp;
```
DudaGod, line 139: All these options specified inside `Mocha.MochaOptions`, so I just use it. Moreover it has a lot of other options that user can use.

src/config/types.ts
```diff
@@ -128,23 +128,16 @@ export interface ExpectOptsConfig {
...
+export interface MochaOpts extends Omit<Mocha.MochaOptions, "ui"> {
+    ui?: string | ((suite: Mocha.Suite) => void);
```
DudaGod, line 132: Currently ui field inside mocha specified incorrectly. It says that I can use only: `bdd`, `tdd` and other prepared interfaces. But actually I can specify just a string to the file which will be required by mocha or use `function`

src/config/types.ts
Dannymex420: ok

Dannymex420 (approved):

PR #1036, opened 2024-11-29
feat: auto ubuntu packages download for local browsers

KuznetsovRoman:
CI side (collect-ubuntu-browser-dependencies, is not being built):
- Download browsers, if necessary
- Resolve direct shared objects dependencies to apt package names per each ubuntu release
- Save direct dependencies into "src/browser-installer/ubuntu-packages/autogenerated" (which is being packed)
- Also saves its cache into "collect-ubuntu-browser-dependencies" in order to speed up next launches

Cache is stored in VCS and is splitted to 2 files:
- Mutual cache. Ubuntu version independent (processed-browsers-linux.json). Removes the need to download browsers each time
- Release specific cache. Could be different on different ubuntu versions (shared-objects-map-ubuntu-%.json). Removes the need to run "apt-file search" each time

image

Client side (src/browser-installer/ubuntu-packages, being packed):
- Resolve direct dependencies to full list of apt packages
- Download and extract them
- Set `LD_LIBRARY_PATH` when launching webdriver so it could populate to browser, so browser would use downloaded packages


.github/workflows/collect-deps.yml
```diff
@@ -0,0 +1,43 @@
+name: Collect ubuntu browser dependencies
```
KuznetsovRoman, line 1: Scheduled action:
- Collects ubuntu dependencies for new browser versions
- Creates a PR to add dependency if there is any new library
- If there are no extra dependencies, do nothing.
- If there is extra browser, but no new dependencies, creates PR to update cache, so it would not need to download the browser later

src/collect-ubuntu-browser-dependencies/shared-object.ts
```diff
@@ -0,0 +1,34 @@
...
+        return "libc6";
+    }
+
+    const relevantPackageName = _.minBy(packages, packageName => calcLevenshtein(sharedObject, packageName)) as string;
+
+    return relevantPackageName;
+};
+
+export const getBinarySharedObjectDependencies = async (binaryPath: string): Promise<string[]> => {
+    const sharedObjectRegExp = /^\s*\dx\d+\s\(NEEDED\)\s*Shared library: \[(.*)\]/gm;
```
sipayRT, line 20: please, add an example for `readElf` to understand what is going on here

src/collect-ubuntu-browser-dependencies/shared-object.ts
```diff
@@ -0,0 +1,34 @@
...
+
+    const readElfResult = await readElf(binaryPath, { dynamic: true });
+
+    let regExpResult = sharedObjectRegExp.exec(readElfResult);
+    const sharedObjectDependencies: string[] = [];
+
+    while (regExpResult && regExpResult[1]) {
+        sharedObjectDependencies.push(regExpResult[1]);
+
+        regExpResult = sharedObjectRegExp.exec(readElfResult);
```
sipayRT, line 30: how does it work? why does the second call return a different value?

test/src/collect-ubuntu-browser-dependencies/shared-object.ts
```diff
@@ -0,0 +1,81 @@
...
+            const packageName = await searchSharedObjectPackage("libnss3.so");
+
+            assert.equal(packageName, "libnss3");
+        });
+    });
+
+    describe("getBinarySharedObjectDependencies", () => {
+        it("should return binary direct shared object deps", async () => {
+            readElfStub.resolves(`
+Dynamic section at offset 0xb00 contains 26 entries:
```
sipayRT, line 46: o_O

package.json
```diff
@@ -9,10 +9,11 @@
   ],
   "scripts": {
     "build": "tsc --build && npm run copy-static && npm run build-bundles",
-    "copy-static": "copyfiles 'src/browser/client-scripts/*' build",
+    "copy-static": "copyfiles 'src/browser/client-scripts/*' 'src/**/*.json' build",
     "build-node-bundle": "esbuild ./src/bundle/cjs/index.ts --outdir=./build/src/bundle/cjs --bundle --format=cjs --platform=node --target=ES2021",
     "build-browser-bundle": "node ./src/browser/client-scripts/build.js",
     "build-bundles": "concurrently -c 'auto' 'npm:build-browser-bundle' 'npm:build-node-bundle --minify'",
+    "resolve-ubuntu-dependencies": "ts-node ./src/collect-ubuntu-browser-dependencies",
```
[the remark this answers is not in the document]
KuznetsovRoman: Yes, it can

src/browser-installer/install.ts
```diff
@@ -6,7 +6,7 @@ import _ from "lodash";
 export const installBrowser = async (
     browserName?: string,
     browserVersion?: string,
-    { force = false, installWebDriver = false } = {},
+    { force = false, shouldInstallWebDriver = false, shouldInstallUbuntuPackages = true } = {},
```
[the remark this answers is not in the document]
KuznetsovRoman: Lets pretend you want to install browser.
Does it mean you also need webdriver? No, because you might want to run "devtools" protocol
Does it mean you also need ubuntu packages? Yes (check "if it is ubuntu" is inside)

I can change both to "true" though

src/browser-installer/chromium/revisions/win32.ts
```diff
@@ -8,7 +8,6 @@ export default {
     79: 707225,
     80: 722370,
     81: 737194,
-    82: 750023,
```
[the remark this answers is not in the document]
KuznetsovRoman: For `chrome for testing` versions (upstream) we dont need to manually download drivers, as this can be handled by `@puppeteer/browsers`

But for older versions we need to download chromedriver ourselves. We only can download chromedriver manually if we know its full version: `${number}.${number}.${number}.${number}`
To get latest chrome driver version for chrome/(ium) milestone we use this API:
`https://chromedriver.storage.googleapis.com/LATEST_RELEASE_<N>`, where `N` - chrome/(ium) milestone
Example: https://chromedriver.storage.googleapis.com/LATEST_RELEASE_112 (resolves 112.0.5615.49)

For unknown reason, this API just does not work when `N = 82`

src/browser-installer/constants.ts
```diff
@@ -7,7 +7,10 @@ export const MIN_CHROMEDRIVER_FOR_TESTING_VERSION = 115;
 export const MIN_CHROMEDRIVER_MAC_ARM_NEW_ARCHIVE_NAME = 106;
 export const MIN_CHROMIUM_MAC_ARM_VERSION = 93;
 export const MIN_CHROMIUM_VERSION = 73;
+export const MIN_FIREFOX_VERSION = 60;
```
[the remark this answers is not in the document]
KuznetsovRoman: Why?
Is there a reason why we dont need to support lower versions?

It only matters when we read browser executable elf files in order to check "which libraries does it depend on"
So increasing this number can a little bit reduce downloading apt packages size, but insignificantly

src/browser-installer/ubuntu-packages/apt.ts
```diff
@@ -0,0 +1,156 @@
...
+            }
+        });
+    });
+};
+
+export const installUbuntuPackages = async (packages: string[], destination: string): Promise<void> => {
+    if (!packages) {
+        browserInstallerDebug(`There are no ubuntu packages to install`);
+
+        return fs.ensureDir(destination);
```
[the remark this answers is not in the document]
KuznetsovRoman: We need some way to distinguish between these states:
- We already tried to install ubuntu packages, but they all already installed
- We haven't tried to install ubuntu packages

With this `fs.ensureDir` we save 2 `apt` execs

I will try to resolve the issue other way, via registry

src/browser-installer/ubuntu-packages/utils.ts
```diff
@@ -0,0 +1,86 @@
...
+
+    const fileContents = await fs.promises.readFile(OS_RELEASE_PATH, "utf8");
+    const result = {} as OsRelease;
+
+    for (const line of fileContents.split("\n")) {
+        if (!line.includes("=")) {
+            continue;
+        }
+
+        const splitPosition = line.indexOf("=");
```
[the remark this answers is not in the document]
KuznetsovRoman: > split

Ok

> replaceAll

What if there is some `"` in the middle?

src/browser-installer/ubuntu-packages/utils.ts
```diff
@@ -0,0 +1,86 @@
...
+
+let isUbuntuCached: boolean | null = null;
+
+export const isUbuntu = async (): Promise<boolean> => {
+    if (isUbuntuCached !== null) {
+        return isUbuntuCached;
+    }
+
+    isUbuntuCached = await osRelease()
+        .then(release => release.ID === "ubuntu")
```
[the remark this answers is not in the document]
KuznetsovRoman: Its `id`, not `name`

By specification (https://manpages.org/os-release/5), it is always lower case:

```
ID=
**A lower-case string** (no spaces or other characters outside of 0-9, a-z, ".", "_" and "-") identifying the operating system, excluding any version information and suitable for processing by scripts or usage in generated filenames. If not set, defaults to "ID=linux". Example: "ID=fedora" or "ID=debian".
```

src/collect-ubuntu-browser-dependencies/index.ts
```diff
@@ -0,0 +1,69 @@
...
+
+        cache.savePackageName(sharedObject, packageName);
+
+        return packageName;
+    };
+
+async function main(): Promise<void> {
+    const ubuntuMilestone = await getUbuntuMilestone();
+
+    logger.log(`Detected ubuntu release: "${ubuntuMilestone}"`);
```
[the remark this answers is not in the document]
KuznetsovRoman: It relies on `VERSION_ID` in `/etc/os-release`
If it is ubuntu, `VERSION_ID` exists and defined
`getUbuntuMilestone` implies it is ubuntu

src/collect-ubuntu-browser-dependencies/index.ts
```diff
@@ -0,0 +1,69 @@
+import _ from "lodash";
+import { EXTRA_FIREFOX_SHARED_OBJECTS } from "./constants";
+import { getBinarySharedObjectDependencies, searchSharedObjectPackage } from "./shared-object";
+import { Cache } from "./cache";
+import { fetchBrowsersMilestones } from "./browser-versions/index";
+import { downloadBrowserVersions } from "./browser-downloader";
+import { getUbuntuMilestone, writeUbuntuPackageDependencies } from "../browser-installer/ubuntu-packages";
+import logger from "../utils/logger";
+
+const createResolveSharedObjectToPackageName =
```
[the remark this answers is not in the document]
KuznetsovRoman: > createSharedObjectToPackageNameResolver

Resolver is a noun. We could call what is returned as `Resolver` only if its object with some method, but its function

This function creates function "resolveSharedObjectToPackageName", so it is called "createResolve..."

src/collect-ubuntu-browser-dependencies/shared-object.ts
KuznetsovRoman: In code section?
There is in test section

package.json
```diff
@@ -9,10 +9,11 @@
   ],
   "scripts": {
     "build": "tsc --build && npm run copy-static && npm run build-bundles",
-    "copy-static": "copyfiles 'src/browser/client-scripts/*' build",
+    "copy-static": "copyfiles 'src/browser/client-scripts/*' 'src/**/*.json' build",
```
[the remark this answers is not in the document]
KuznetsovRoman: Restricted to "src/**/[!cache]*/autogenerated/**/*.json"
So, jsons need to be in "autogenerated" directory

src/browser-installer/chromium/revisions/win32.ts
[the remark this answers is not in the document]
KuznetsovRoman: Renamed to ".json"

src/browser-installer/ubuntu-packages/apt.ts
```diff
@@ -0,0 +1,156 @@
...
+
+    const existingDependencies = await new Promise<string[]>((resolve, reject) => {
+        exec(`apt list ${dependencies.join(" ")} --installed`, (err, result) => {
+            if (err) {
+                reject(err);
+            } else {
+                const lines = result.split("\n");
+                const existingDependencies = lines
+                    .map(line => {
+                        const slashIndex = line.indexOf("/");
```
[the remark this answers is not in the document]
KuznetsovRoman: Why do we need "version"? It will be marked as unusable by linter

src/browser-installer/chrome/index.ts
```diff
@@ -6,41 +6,30 @@ import { DRIVER_WAIT_TIMEOUT } from "../constants";
...
-    const [chromeDriverPath] = await Promise.all([
+    const [chromeDriverPath, randomPort, chromeDriverEnv] = await Promise.all([
         installChromeDriver(chromeVersion),
-        installChrome(chromeVersion),
-        shouldInstallUbuntuPackageDependencies ? installUbuntuPackageDependencies() : null,
+        getPort(),
+        isUbuntu()
+            .then(isUbuntu => (isUbuntu ? getUbuntuLinkerEnv() : null))
+            .then(extraEnv => (extraEnv ? { ...process.env, ...extraEnv } : process.env)),
     ]);
```
KuznetsovRoman, line 23: Now we dont install browser and ubuntu packages here, and instead we assume it is already installed by "browser-installer/run.ts", which runs "browser-installer/install.ts" before calling "run*Driver"

src/browser-installer/run.ts
```diff
@@ -1,21 +1,27 @@
...
-    driverName: SupportedDriver,
+    browserName: SupportedBrowser,
     browserVersion: string,
     { debug = false } = {},
 ): Promise<{ gridUrl: string; process: ChildProcess; port: number }> => {
-    switch (driverName) {
-        case Driver.CHROMEDRIVER:
+    const installBrowserOpts = { shouldInstallWebDriver: true, shouldInstallUbuntuPackages: true };
+
+    await installBrowser(browserName, browserVersion, installBrowserOpts);
```
KuznetsovRoman, line 12: As now we need to install browser before running its driver, we expect "browserName" instead of "driverName" in order to pass "browserName" to "installBrowser"

src/browser-installer/utils.ts
```diff
@@ -29,21 +29,27 @@ export const Driver = {
 export type SupportedBrowser = (typeof Browser)[keyof typeof Browser];
 export type SupportedDriver = (typeof Driver)[keyof typeof Driver];
 
-export const getDriverNameForBrowserName = (browserName: SupportedBrowser): SupportedDriver | null => {
-    if (browserName === Browser.CHROME || browserName === Browser.CHROMIUM) {
-        return Driver.CHROMEDRIVER;
+export const getNormalizedBrowserName = (
```
KuznetsovRoman, line 32: As now we dont map "browserName" to "driverName" for "runBrowserDriver", we dont need "getDriverNameForBrowserName", and instead i created this function, so in "installBrowser" we can be sure it only receives valid "browserName" of `SupportedBrowser` type

KuznetsovRoman: This way we can handle cases with invalid browser name outside of "install", so we know, what exactly user wanted to do, and we can provide relevant error message:
- invalid browser name was used in order to run local browser
- invalid browser name was used in order to download a browser

src/browser-pool/webdriver-pool.ts
```diff
@@ -17,13 +17,13 @@ export class WebdriverPool {
...
+        browserName?: string,
+        browserVersion?: string,
         { debug = false } = {},
     ): ReturnType<typeof this.createWebdriverProcess> {
-        const driverName = getDriverNameForBrowserName(browserName);
+        const browserNameNormalized = getNormalizedBrowserName(browserName);
 
-        if (!driverName) {
+        if (!browserNameNormalized) {
             throw new Error(
```
KuznetsovRoman, line 27: As we now dont want to map "browserName" to "driverName", we still need to check if "browserName" is valid (because this method is called inside NewBrowser with its config's browserName, which is user provided value, which can be invalid)

src/browser/new-browser.ts
```diff
@@ -226,13 +225,26 @@ export class NewBrowser extends Browser {
...
+                    `Running auto local "${this._config.desiredCapabilities?.browserName}" is unsupported`,
+                    `Supported browsers: "chrome", "firefox", "safari", "edge"`,
+                ].join("\n"),
+            );
+        }
+
         const executablePath = await installBrowser(
-            this._config.desiredCapabilities?.browserName as SupportedBrowser,
-            this._config.desiredCapabilities?.browserVersion as string,
+            normalizedBrowserName,
```
KuznetsovRoman, line 241: "installBrowser" now only wants normalized browser name (so we could pull out "invalid browser name" handling in order to throw relevant errors)

src/browser/new-browser.ts
```diff
@@ -226,13 +225,26 @@ export class NewBrowser extends Browser {
...
+            this._config.desiredCapabilities?.browserVersion,
+            { shouldInstallWebDriver: false, shouldInstallUbuntuPackages: true },
```
KuznetsovRoman, line 243: Explicitly setting "should install" options, as they are all set to "false" now by default.

In "_addExecutablePath" we dont need to install webdriver as user might want to run browser in "devtools" protocol

.github/workflows/collect-deps.yml
```diff
@@ -0,0 +1,57 @@
...
+              with:
+                  node-version: 18
+                  registry-url: https://registry.npmjs.org
+            - name: Prepare Ubuntu
+              run: sudo apt-get update && sudo apt-get install -y apt-file && sudo apt-file update
+            - run: npm ci
+            - name: Config git
+              run: git config --global user.name "y-infra" && git config --global user.email "y-infra@yandex.ru"
+            - name: Fetch branches
+              run: git fetch --all
```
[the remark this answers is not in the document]
KuznetsovRoman: Removed

src/browser-installer/install.ts
```diff
@@ -25,28 +25,30 @@ export const installBrowser = async (
...
 
-        return installWebDriver
-            ? await Promise.all([
-                  installChrome(browserVersion, { force }),
-                  installChromeDriver(browserVersion, { force }),
-              ]).then(binaries => binaries[0])
-            : installChrome(browserVersion, { force });
+        return await Promise.all([
+            installChrome(browserVersion, { force }),
+            shouldInstallWebDriver && installChromeDriver(browserVersion, { force }),
```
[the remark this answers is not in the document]
KuznetsovRoman: Fixed

src/browser-installer/registry/cli-progress-bar.ts
```diff
@@ -1,35 +1,39 @@
...
 
     const register: RegisterProgressBarFn = (browserName, browserVersion) => {
         let bar: SingleBar;
 
-        const downloadProgressCallback: DownloadProgressCallback = (downloadedBytes, totalBytes) => {
+        const downloadProgressCallback: DownloadProgressCallback = (done, total = 100) => {
             if (!bar) {
-                const totalMB = Math.round((totalBytes / BYTES_PER_MEGABYTE) * 100) / 100;
-                bar = progressBar.create(totalMB, 0, { filename: `${browserName}@${browserVersion}` });
+                bar = progressBar.create(100, 0, { filename: `${browserName}@${browserVersion}` });
```
KuznetsovRoman, line 21: Now use percents as more versatile progress bar unit.
This way we can also use the same progress bar for ubuntu packages download

src/browser-installer/utils.ts
```diff
@@ -129,13 +142,42 @@ const getCacheDir = (envValueOverride = process.env.TESTPLANE_BROWSERS_PATH): st
 export const getRegistryPath = (envValueOverride?: string): string =>
     path.join(getCacheDir(envValueOverride), "registry.json");
 
+export const readRegistry = (registryPath: string): Registry => {
```
KuznetsovRoman, line 145: I decided to extend registry so it would store not only binary paths, but could store anything else
This way we can manage os packages by this registry file too.

src/browser-installer/ubuntu-packages/apt.ts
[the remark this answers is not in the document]
KuznetsovRoman: Resolved

src/browser-installer/utils.ts
```diff
@@ -36,7 +35,7 @@ export type BinaryKey = `${BinaryName}_${BrowserPlatform}`;
 export type OsName = string;
 export type OsVersion = string;
 export type OsPackagesKey = `${OsName}_${OsVersion}`;
-export type Registry = {
+export type RegistryFileContents = {
```
sipayRT, line 38: this should be part of the Registry module

src/browser-installer/registry/index.ts
```diff
@@ -12,204 +13,317 @@ import {
...
+        let fsData: Record<string, unknown>;
 
-    const comparator = browserName === Browser.FIREFOX ? firefoxVersionComparator : semverVersionsComparator;
-    const suitableBuildIdsSorted = suitableBuildIds.sort(comparator);
+        if (fs.existsSync(this.registryPath)) {
+            fsData = fs.readJSONSync(this.registryPath);
 
-    return suitableBuildIdsSorted[suitableBuildIdsSorted.length - 1];
-};
+            const isRegistryV0 = fsData && !fsData.meta;
```
sipayRT, line 259: lets get rid of V0 support. we can try to install browsers and drivers again if the `meta` field is missing (with generating new registry)

sipayRT (approved):

src/browser-installer/utils.ts
```diff
@@ -29,21 +29,27 @@ export const Driver = {
 export type SupportedBrowser = (typeof Browser)[keyof typeof Browser];
 export type SupportedDriver = (typeof Driver)[keyof typeof Driver];
 
-export const getDriverNameForBrowserName = (browserName: SupportedBrowser): SupportedDriver | null => {
-    if (browserName === Browser.CHROME || browserName === Browser.CHROMIUM) {
-        return Driver.CHROMEDRIVER;
+export const getNormalizedBrowserName = (
+    browserName?: string,
+): Exclude<SupportedBrowser, typeof Browser.CHROMIUM> | null => {
```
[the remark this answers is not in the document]
KuznetsovRoman: We dont need to, but `chromium` maps to `chrome`, as installing "chrome" and "chromium" has a single entry point, being "chrome", as "chromium" is not a valid W3C browserName

src/browser-installer/registry/index.ts
KuznetsovRoman: You mean, "delete V0 file"?

PR #1038, opened 2024-12-02
Fix types

DudaGod:


src/config/types.ts
```diff
@@ -128,8 +128,94 @@ export interface ExpectOptsConfig {
     interval: number;
 }
 
-export interface MochaOpts extends Omit<Mocha.MochaOptions, "ui"> {
```
DudaGod, line 131: Remove `Mocha.MochaOptions` in order to not get conflicts in `it`, `describe` etc. Reproduced when service has `@types/mocha` module (can be installed as dependencies from some other module).

src/config/types.ts
```diff
@@ -260,9 +346,13 @@ export interface SetsConfigParsed {
     browsers: Array<string>;
 }
 
+type PartialCommonConfig = Partial<Omit<CommonConfig, "system">> & {
+    system?: Partial<SystemConfig>;
```
DudaGod, line 350: Field `system` should not require specifying all fields

shadowusr (approved):

PR #1039, opened 2024-12-03
fix: ability to specify "mochaOpts.ui" as function

DudaGod:


src/config/types.ts
```diff
@@ -215,7 +215,7 @@ export interface MochaOpts {
     isWorker?: boolean;
 
     /** Interface name or path to file with custom interface implementation. */
-    ui?: string | Suite;
+    ui?: string | ((suite: unknown) => void);
```
DudaGod, line 218: Here I made a mistake with the type in the past PR. Suite specified as `unknown` because I can't use `Mocha.Suite` and it differ from our testplane suite. Moreover I don't want to describe `Mocha.Suite` because it has a lot of fields.

PR #1044, opened 2024-12-24
fix: remove unknown capabilities when running local browser

KuznetsovRoman:


src/browser/new-browser.ts
```diff
@@ -1,51 +1,53 @@
...
-    },
-    edge: {
-        capabilityName: "ms:edgeOptions",
-        getArgs: (): string[] => ["--headless"],
-    },
-    microsoftedge: {
+    [BrowserName.EDGE]: {
         capabilityName: "ms:edgeOptions",
         getArgs: (): string[] => ["--headless"],
     },
```
KuznetsovRoman, line 53: Refactored "headlessBrowserOptions" so its key is valid w3c browserName, which is received by "getNormalizedBrowserName" function, which is moved from "src/browser-installer" to "src/utils/browser"

src/browser/new-browser.ts
```diff
@@ -221,36 +228,50 @@ export class NewBrowser extends Browser {
...
+                const isW3CCapability = W3C_CAPABILITIES.includes(capabilityName);
+                const isVendorSpecificCapability = VENDOR_CAPABILITIES[browserNameW3C].includes(capabilityName);
+
+                return isW3CCapability || isVendorSpecificCapability;
+            })
+            .reduce((acc, capabilityName) => {
+                return set(acc, [capabilityName], capabilities[capabilityName as keyof WebdriverIO.Capabilities]);
+            }, {});
+
+        return filteredCapabilities;
```
KuznetsovRoman, line 274: Remove all capabilities, which are not "w3c capabilities" or "vendor specific capabilities", as local webdriver can't recognize those and fail to launch browser

src/browser/types.ts
```diff
@@ -7,6 +8,16 @@ import { OpenAndWaitCommand } from "./commands/openAndWait";
...
 
+export const BrowserName = {
+    CHROME: PuppeteerBrowser.CHROME,
+    CHROMIUM: PuppeteerBrowser.CHROMIUM,
+    FIREFOX: PuppeteerBrowser.FIREFOX,
+    SAFARI: "safari",
+    EDGE: "MicrosoftEdge",
+} as const;
+
+export type W3CBrowserName = Exclude<(typeof BrowserName)[keyof typeof BrowserName], PuppeteerBrowser.CHROMIUM>;
```
KuznetsovRoman, line 19: Moved from "src/browser-installer"

src/constants/config.ts
```diff
@@ -0,0 +1,38 @@
...
+export const WEBDRIVER_PROTOCOL = "webdriver";
+export const DEVTOOLS_PROTOCOL = "devtools";
+export const SAVE_HISTORY_MODE = {
+    ALL: "all",
+    NONE: "none",
+    ONLY_FAILED: "onlyFailed",
+};
+export const NODEJS_TEST_RUN_ENV = "nodejs";
+export const BROWSER_TEST_RUN_ENV = "browser";
+export const LOCAL_GRID_URL = "local";
```
KuznetsovRoman, line 12: ".js" -> ".ts"

src/browser/new-browser.ts
```diff
@@ -254,7 +260,18 @@ export class NewBrowser extends Browser {
...
```
shadowusr, line 274: I suggest to write a message if some capabilities were omitted, because I think we might stumble on this in the future in circumstances we can't foresee now. Something like "Dropping non-standard capabilities due to running in local mode". Perhaps this should be in debug logs.

shadowusr (approved):

PR #1045, opened 2024-12-25
chore: refactor(move) browser-installer utils

KuznetsovRoman:


src/constants/config.js
KuznetsovRoman, line 1: Renamed to ".ts"

sipayRT (approved):

PR #1051, opened 2025-01-09
feat: add 'list-browsers' command

KuznetsovRoman:


src/cli/commands/list-browsers/index.ts
```diff
@@ -0,0 +1,102 @@
...
+    return _(browsersArray).uniq().sortBy(["browserName", "browserVersion"]).value();
+};
+
+export const registerCmd = (cliTool: typeof commander, testplane: Testplane): void => {
+    cliTool
+        .command(commandName)
+        .description("Lists all browsers from the config")
+        .option(
+            "--type [type]",
+            "return browsers in specified type ('tags' / 'names')",
```
sipayRT, line 67: I think that this option will be unclear to users. Without an example they won't understand what `tag` and `name` are.

PR #1053, opened 2025-01-13
feat: add 'config' command

KuznetsovRoman:


src/cli/commands/config/index.ts
```diff
@@ -0,0 +1,25 @@
+import { Testplane } from "../../../testplane";
+import { CliCommands } from "../../constants";
+import logger from "../../../utils/logger";
+
+const { CONFIG: commandName } = CliCommands;
+
+export const registerCmd = (cliTool: typeof commander, testplane: Testplane): void => {
+    cliTool
+        .command(commandName)
+        .description("Lists all browsers from the config")
```
sipayRT, line 10: did you forget to change the description?

src/cli/commands/config/index.ts
```diff
@@ -0,0 +1,25 @@
...
+        .option("-c, --config <path>", "path to configuration file")
```
sipayRT, line 11: for `config` command you can pass the configuration file as an argument rather than as an option

KuznetsovRoman: No, we can't

src/cli/commands/config/index.ts
KuznetsovRoman: "Outputs Testplane config (including default and overriden by environment variables values)"

sipayRT (approved):

PR #1060, opened 2025-02-07
fix: correctly define screenshotsDir option for browsers

DudaGod:
### What is done

The screenshotsDir field is not parsed correctly for browsers if the user uses the deprecated value (hermione/screens) and the field itself is specified at the top level (i.e. not inside any browser).

### DIFF
--- a/src/config/browser-options.js
+++ b/src/config/browser-options.js
@@ -189,14 +189,19 @@ function buildBrowserOptions(defaultFactory, extra) {
                     throw new Error('"screenshotsDir" must be a string or function');
                 }
             },
-            map: (value, _, __, { isSetByUser }) => {
-                const deprecatedScreensPath = "hermione/screens";
-
-                if (!isSetByUser && fs.existsSync(deprecatedScreensPath) && !fs.existsSync(value)) {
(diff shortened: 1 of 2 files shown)


src/config/browser-options.js
```diff
@@ -189,14 +189,19 @@ function buildBrowserOptions(defaultFactory, extra) {
...
-            map: (value, _, __, { isSetByUser }) => {
-                const deprecatedScreensPath = "hermione/screens";
+            map: (value, config, __, { isSetByUser }) => {
+                if (isSetByUser) {
+                    return value;
+                }
 
-                if (!isSetByUser && fs.existsSync(deprecatedScreensPath) && !fs.existsSync(value)) {
-                    return deprecatedScreensPath;
+                const topLevelScreenshotsDir = _.get(config, "screenshotsDir");
```
DudaGod, line 197: for top level option this will be `undefined` because option is not set. But when parse browser this option will have value from top level. And will be used if user not set it specially for browser

sipayRT (approved):

KuznetsovRoman (approved):

PR #1061, opened 2025-02-10
chore: re-write existing-browser to typescript

shadowusr:
## What's done
- existing-browser and all related files, including history, client-bridge and many others were re-written to typescript
- fixed all unit tests that broke along the way

A reminder on why unit tests need fixing after migrating to TS: since we are using SWC, it transforms modules into non-editable getters, which cannot be stubbed in a `sinon.stub(module, "method")` manner. The only way to stub such modules is via proxyquire. In the future we'll migrate to vitest which has more advanced&convenient ways of doing this.


sipayRT (approved):
комменты только почисть, плз

PR #1062, opened 2025-02-24
feat: implement time travel mvp

shadowusr:
### New `DOM_SNAPSHOTS` event
This event is emitted as soon as new DOM snapshots are available, which makes browser window streaming possible. At most, this event is emitted before every action in test.

Example usage:
```typescript
import type { eventWithTime as RrwebEvent } from "@rrweb/types";

interface TestContext {
    testPath: string[];
    browserId: string;
}

interface SnapshotsData {
    rrwebSnapshots: RrwebEvent[];
}

testplane.on(testplane.events.DOM_SNAPSHOTS, (context: TestContext, data: SnapshotsData): void => {/* ... */});
```

### New `record` config option
Available record modes:
```typescript
enum RecordMode {
    // Record and save all test runs
    On = "on",
    // Do not record any test runs
    Off = "off",
    // Record and save all retries
    OnForRetries = "on-for-retries",
    // Record all test runs, but save only last failed run
    SaveLastFailedRun = "save-last-failed-run",
}
```

This option is `"off"` by default. Example usage:
```typescript
import {RecordMode} from "testplane";

export = {
    // ...
    record: RecordMode.SaveLastFailedRun,
};
```
Can also be an object (more options will be added in the future for more granular control):
```typescript
import {RecordMode} from "testplane";

export = {
    // ...
    record: { mode: RecordMode.SaveLastFailedRun },
};
```

### `executionContext.ctx` now has an `attempt` property
This new property holds a number of current attempt, starting from 0. Example usage:
```typescript
it("some sample test", ({browser}) => {
    const currentAttempt = browser.executionContext.ctx.attempt;
    console.log("Current attempt: " + currentAttempt);
});
```

### Installed rrweb to capture DOM snapshots (if enabled in project config)
Snapshots are captured on `before` history hook. This means that we try to install rrweb and get events before each action in user tests.

### New `runWithoutHistory` hook (for internal usage only)
Use this hook if you want to pass by history hooks. Anything that runs in its callback is not saved in history. Example usage:
```typescript
runWithoutHistory({ callstack: this._callstackHistory! }, async () => {
    await this._cleanupPageAnimations();
});
```


src/browser/history/callstack.ts
```diff
@@ -4,12 +4,22 @@ import { TestStepKey, TestStep } from "../../types";
...
         this._history = [];
         this._stack = [];
     }
 
+    /** Bypass mode indicates that no hooks should be invoked (e.g. `before()` / `after()`). Useful for system commands that should be run bypassing hooks. */
+    get isInBypassMode(): boolean {
+        return this._isInBypassMode;
+    }
+
+    setIsInBypassMode(flag: boolean): void {
```
sipayRT, line 19: why not just `setBypassMode`?

src/browser/history/index.ts
```diff
@@ -37,42 +41,79 @@ const mkHistoryNode = ({ name, args, elementScope, key, overwrite, isGroup }: No
...
+    if (callstack.isInBypassMode) {
+        return fn();
+    }
+
     return runWithHooks({
-        before: () => callstack.enter(mkHistoryNode(nodeData)),
+        before: async () => {
+            const recordMode = config.record.mode;
+            const isRetry = (session.executionContext?.ctx?.attempt ?? 0) > 0;
+            const shouldRecord = recordMode !== RecordMode.Off && !(recordMode === RecordMode.OnForRetries && !isRetry);
```
sipayRT, line 78: it should be separate helper I guess
and what about `SaveLastFailedRun` mode?

also, why do we need double negation here instead of `... && recordMode === RecordMode.OnForRetries && isRetry`

src/browser/history/rrweb.ts
```diff
@@ -0,0 +1,78 @@
...
+        }, rrwebCode),
+    );
+    /* eslint-enable @typescript-eslint/ban-ts-comment */
+}
+
+export function filterEvents(rrwebEvents: eventWithTime[]): eventWithTime[] {
+    return rrwebEvents.filter(e => {
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        const dataAny = e?.data as any;
+        return !dataAny?.attributes?.[0]?.attributes?.["cd_frame_id_"];
```
sipayRT, line 61: what's going on here?

src/config/types.ts
```diff
@@ -249,6 +249,21 @@ type ReadinessProbeObj = {
 
 type ReadinessProbe = ReadinessProbeFn | ReadinessProbeObj;
 
+export enum RecordMode {
+    // Record and save all test runs
+    On = "on",
+    // Do not record any test runs
+    Off = "off",
+    // Record and save all retries
+    OnForRetries = "on-for-retries",
```
sipayRT, line 258: `onRetries`?

src/config/types.ts
```diff
@@ -249,6 +249,21 @@ type ReadinessProbeObj = {
...
+    // Record all test runs, but save only last failed run
+    SaveLastFailedRun = "save-last-failed-run",
```
sipayRT, line 260: `onLastFailed`?

sipayRT (approved):

src/browser/history/index.ts
shadowusr: Okay, I will move it to a separate helper

> and what about SaveLastFailedRun mode?

Here we determine whether we should record. And recording-wise `SaveLastFailedRun` is equal to `On` — we need to record always, because we don't know if the run will fail. Then comes the saving part — on html-reporter side. There we decide whether to keep the recording or to throw it away.

src/browser/history/callstack.ts
shadowusr: To make it clear that this property is boolean... Otherwise bypass mode sounds like enum IMO

src/browser/history/rrweb.ts
shadowusr: I've left a comment. It is specific to chromedriver and rarely can appear on snapshots, but we are not interested in it.

src/config/types.ts
shadowusr: After some discussion we've agreed to have these options: on, off, retries-only, last-failed-run.

PR #1065, opened 2025-03-13
fix: correctly wrap "$$" to support all commands from expect

DudaGod:


src/browser/existing-browser.ts
```diff
@@ -259,13 +259,15 @@ export class ExistingBrowser extends Browser {
...
-                        [];
+                    const arr: WebdriverIO.Element[] & Partial<Pick<ElementArray, "parent" | "foundWith" | "selector" | "props">> = [];
+
                     const res = await origCommand(selector);
                     for await (const el of res) arr.push(el);
+
                     arr.parent = res.parent;
                     arr.foundWith = res.foundWith;
                     arr.selector = res.selector;
+                    arr.props = res.props;
```
DudaGod, line 270: This field helps to not fail when using `toBeElementsArrayOfSize` command from expect-webdriverio. Fails here - https://github.com/webdriverio/expect-webdriverio/blob/v3.6.0/src/util/formatMessage.ts#L11. So I add `props` to fix it.

PR #1073, opened 2025-04-05
fix: throw error if chrome not support bidi protocol

DudaGod:
### What is done

Chrome@127 and lower doesn't have feature that used inside webdriverio. So when you open url in that browser you got an error: `Navigation to 'https://webdriver.io' timed out as no request payload was received`. According to this error, it is not clear what the problem is. Therefore, when parsing the config, I throw an error if used chrome@127 and lower with BiDi protocol (specified `webSocketUrl: true` capability)

### DIFF
--- a/src/config/browser-options.js
+++ b/src/config/browser-options.js
@@ -7,6 +7,7 @@ const defaults = require("./defaults");
 const optionsBuilder = require("./options-builder");
 const utils = require("./utils");
 const { WEBDRIVER_PROTOCOL, DEVTOOLS_PROTOCOL, SAVE_HISTORY_MODE } = require("../constants/config");
+const { BROWSERS_SUPPORT_BIDI } = require("../constants/browser");
 const { isSupportIsolation } = require("../utils/browser");
 const { RecordMode } = require("./types");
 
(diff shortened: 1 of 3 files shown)


src/config/browser-options.js
```diff
@@ -31,9 +32,21 @@ exports.getPerBrowser = () => {
             validate: (value, config) => {
                 if (_.isNull(value) && _.isNull(config.desiredCapabilities)) {
                     throw new Error('Each browser must have "desiredCapabilities" option');
-                } else {
-                    utils.assertOptionalObject(value, "desiredCapabilities");
                 }
+
+                const { browserName, browserVersion, webSocketUrl } = value;
+
+                if (webSocketUrl === true && browserName === "chrome" && browserVersion) {
```
DudaGod, line 39: user may not specify option `browserVersion`, in this case will be used the latest version of chrome (for example with `gridUrl: local`)

src/config/browser-options.js
```diff
@@ -7,6 +7,7 @@ const defaults = require("./defaults");
 const optionsBuilder = require("./options-builder");
 const utils = require("./utils");
 const { WEBDRIVER_PROTOCOL, DEVTOOLS_PROTOCOL, SAVE_HISTORY_MODE } = require("../constants/config");
+const { MIN_CHROME_VERSION_SUPPORT_BIDI } = require("../constants/browser");
```
sipayRT, line 10: `MIN_CHROME_VERSION_BIDI_SUPPORT`?

src/config/browser-options.js
sipayRT: what about another browsers? firefox, for example

sipayRT (approved):

src/config/browser-options.js
DudaGod: ok, I support it in firefox too

src/config/browser-options.js
DudaGod: Hm, I don't know. I'll leave my own version, I like it better.

Moreover we already have `MIN_CHROME_VERSION_SUPPORT_ISOLATION`

PR #1074, opened 2025-04-05
chore(ci): setup pkg.pr.new

DudaGod:
### What is done:

- enable pkg.pr.new - https://github.com/stackblitz-labs/pkg.pr.new in order to trigger release in PR without publishing anything to npm

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -27,3 +27,7 @@ jobs:
                   cache: "npm"
             - run: npm ci
             - run: npm test
+
+            - name: Publish
+              if: ${{ startsWith(matrix.node-version, '20') }}
+              run: npx pkg-pr-new publish


.github/workflows/node.js.yml
```diff
@@ -27,3 +27,7 @@ jobs:
                   cache: "npm"
             - run: npm ci
             - run: npm test
+
+            - name: Publish
+              if: ${{ matrix.node-version == '20.18.1' }}
```
DudaGod, line 32: latest version of pkg-pr-new doesn't work with node@18. Moreover there is no need to publish package from few node versions

sipayRT (approved):

.github/workflows/node.js.yml
KuznetsovRoman: I dont like the fact that this specific version is mentioned 2 times
I would use "startsWith 20": https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions#startswith

Example: https://github.com/shannah/jdeploy/blob/1e12d6f303e4672e963cef965f31b9a3593c952b/action.yml#L67

Not important though

KuznetsovRoman (approved):

PR #1078, opened 2025-04-17
chore(ci): build project before publish to pkg-pr-new

DudaGod:


.github/workflows/node.js.yml
```diff
@@ -28,6 +28,10 @@ jobs:
             - run: npm ci
             - run: npm test
 
+            - name: Build
```
DudaGod, line 31: I don't know how it worked, but before publish package it must be builded

PR #1082, opened 2025-05-14
feat: request different types of browsers in parallel

DudaGod:
### Problem

With certain values of the options responsible for parallelism, tests in each browser can be performed one after the other. For example, with this global configuration for two browsers:
```
sessionsPerBrowser: 2,
testsPerSession: 1,
system: {
    parallelLimit: 5
},
```

If there are 10 tests at startup, the queue will be blocked by the first browser and the total concurrency will be 2, not 5.

### What is done

Implemented logic with request different types of browsers in parallel.


src/test-collection/index.ts
```diff
@@ -157,6 +157,23 @@ export class TestCollection {
         }
     }
 
+    eachTestAcrossBrowsers(cb: (test: Test, browserId: string, browserVersion?: string) => void): void {
```
DudaGod, line 160: Here I get all tests from all browsers (taking into account different versions in order to support logic from plugin - https://github.com/gemini-testing/testplane-browser-version-changer) and then call callback one by one from different types of browsers. It gives ability to request browsers in parallel.

Previously, all requests from each browser were executed one after the other. And everything works correctly unless the `parallelLimit` option is explicitly specified. Because all browsers have its own parallel limit which specified with option `sessionsPerBrowser`. But if user also specify `parallelLimit` then it can works incorrectly. Because used `LimitedPool` don't know nothing about which types of browsers are used. It just collect requests and limit them by specified value from `parallelLimit`

src/runner/browser-runner.ts
```diff
@@ -59,6 +59,10 @@ export class BrowserRunner extends Runner {
         return true;
     }
 
+    async waitTestsCompletion(): Promise<void> {
```
DudaGod, line 62: Implement new method in order to wait when all tests inside browsers will be completed.

src/runner/index.ts
```diff
@@ -110,31 +110,41 @@ export class MainRunner extends Runner {
...
+
+        this.activeBrowserRunners.forEach(runner => this.running.add(this._waitBrowserRunnerTestsCompletion(runner)));
 
         return this.running.done();
     }
 
-    protected async _runTestsInBrowser(testCollection: TestCollection, browserId: string): Promise<void> {
+    protected _addTestToBrowserRunner(test: Test, browserId: string): BrowserRunner {
+        const browserRunner = this.activeBrowserRunners.get(browserId) || this._createBrowserRunner(browserId);
+        browserRunner.addTestToRun(test);
```
DudaGod, line 129: Now all tests are added to `BrowserRunner` using method `addTestToRun`, because I need to add tests one by one, as a browser request is being executed under the hood. `MainRunner` (this file) -> `BrowserRunner` -> `RegularTestRunner` (here request for browser)

src/runner/index.ts
```diff
@@ -110,31 +110,41 @@ export class MainRunner extends Runner {
...
 
         this.activeBrowserRunners.set(browserId, runner);
 
-        await runner.run(testCollection);
+        return runner;
+    }
 
-        this.activeBrowserRunners.delete(browserId);
+    protected async _waitBrowserRunnerTestsCompletion(runner: BrowserRunner): Promise<void> {
+        await runner.waitTestsCompletion();
```
DudaGod, line 146: All tests added to browser runner without waiting for them to be completed. So I need to wait it after all browser requests are queued

sipayRT (approved):

PR #1085, opened 2025-05-16
feat: implement standalone browser mode

shadowusr:
For usage example, see `examples/standalone`


src/browser/standalone.ts
```diff
@@ -0,0 +1,125 @@
...
+        sessionOpts: {
+            capabilities: session.capabilities,
+            ...session.options
+        }
+    }, calibrator);
+
+    existingBrowser.publicAPI.overwriteCommand('deleteSession', async function() {
+        await existingBrowser.quit();
+        await newBrowser.kill();
+    });
```
KuznetsovRoman: Shouldn't we call original deleteSession?

src/browser/standalone.ts
```diff
@@ -0,0 +1,128 @@
...
+    const config = new Config({
+        browsers: {
+            [browserName]: browserConfig,
+        },
+        system: options.system || {
+            debug: true,
+        },
+    });
+
+    const webdriverPool = new WebdriverPool();
```
KuznetsovRoman, line 87: Didn't you consider to make it singleton?

KuznetsovRoman (approved):

src/browser/standalone.ts
```diff
@@ -0,0 +1,128 @@
...
+    | "windowSize"
+    | "orientation"
+    | "headers"
+    | "strictSSL"
+    | "user"
+    | "key"
+    | "system"
+>;
+
+export type StandaloneBrowserOptionsInput = Partial<StandaloneBrowserOptions>;
```
DudaGod, line 34: I can't just specify this option:
```
{
  system: {debug: false}
}
```

Looks like we should make partial nested options

src/browser/standalone.ts
```diff
@@ -0,0 +1,128 @@
...
+        throw new Error(
+            [
+                `Running browser "${browserName}" is unsupported`,
+                `Supported browsers: "chrome", "firefox", "safari", "edge"`,
+            ].join("\n"),
+        );
+    }
+
+    RuntimeConfig.getInstance().extend({
+        debug: options.system?.debug ?? true,
```
DudaGod, line 52: are you sure that it works correctly? I can't disable debug option

examples/standalone/README.md
```diff
@@ -0,0 +1,18 @@
+### Testplane Standalone Browser Usage Example
```
DudaGod, line 1: Looks like we should add .gitignore in shit folder in order to not commit captured screens

DudaGod (approved):

src/browser/standalone.ts
shadowusr: No, it's called inside kill, just like we do it in quit()

src/browser/standalone.ts
shadowusr: Fair, made it so that it was created only once.

PR #1087, opened 2025-05-20
fix: rename record option to timeTravel

shadowusr:
As Time Travel is still experimental feature, we do not consider this to be a breaking change.

This PR renames `record` config option and `RecordMode` enum which is exported to `timeTravel` and `TimeTravelMode` correspondingly.


sipayRT (approved):

src/config/types.ts
```diff
@@ -249,7 +249,7 @@ type ReadinessProbeObj = {
 
 type ReadinessProbe = ReadinessProbeFn | ReadinessProbeObj;
 
-export enum RecordMode {
```
KuznetsovRoman, line 252: Wouldn't it break html-reporter?

DudaGod (approved):

src/config/types.ts
shadowusr: No, we use safe imports there, so it wouldn't crash. It would just consider that time travel is not available (user is using old testplane version).

But I will test that, just to be sure.

PR #1089, opened 2025-05-20
fix: export runCli from testplane

shadowusr:


src/index.ts
```diff
@@ -7,6 +7,7 @@ import "./browser/types";
 import "expect-webdriverio";
 
 import { GlobalHelper } from "./types";
+export { run as runCli } from "./cli";
```
KuznetsovRoman, line 10: This is "feat"

KuznetsovRoman (approved):

PR #1091, opened 2025-05-29
chore: remove bluebird

KuznetsovRoman:


src/utils/promise.ts
```diff
@@ -0,0 +1,66 @@
...
+        try {
+            const result = fn.apply(this, args);
+            return Promise.resolve(result);
+        } catch (err) {
+            return Promise.reject(err);
+        }
+    };
+}
+
+export function promiseTimeout<T>(promise: Promise<T>, ms: number, message: string): Promise<T> {
```
DudaGod, line 14: why not just use tiny modules like `p-timeout`?

DudaGod (approved):

PR #1092, opened 2025-05-29
feat: add support of @testing-library/dom queries

talkenson:


src/browser/queries/types.ts
```diff
@@ -0,0 +1,69 @@
...
+export type Config = Pick<
+    BaseConfig,
+    | "asyncUtilTimeout"
+    | "computedStyleSupportsPseudoElements"
+    | "defaultHidden"
+    | "testIdAttribute"
+    | "throwSuggestions"
+>;
+
+export type WebdriverIOQueryReturnType<Element, ElementArray, T> = T extends Promise<HTMLElement>
```
shadowusr, line 23: I don't think it's a good idea to use webdriverio naming all around. We are planning to begin distancing testplane from wdio

src/browser/queries/index.ts
```diff
@@ -0,0 +1,265 @@
...
+        };
+    }
+}
+
+/*
+eslint-disable
+@typescript-eslint/explicit-function-return-type
+*/
+
+const DOM_TESTING_LIBRARY_UMD_PATH = path.join(
```
shadowusr, line 35: since this is a complete copy of https://github.com/testing-library/webdriverio-testing-library, i think we should give them credit, as required by MIT license

src/browser/commands/_addDTLCommands.ts
```diff
@@ -0,0 +1,10 @@
+import { Browser } from "../browser";
+import { setupBrowser } from "../queries";
+
+const _addDTLCommands = (browser: Browser): void => {
```
shadowusr, line 4: This looks really weird. commands directory has actual commands implementations, but this one is more of a utility function.

I think it would better fit around here: https://github.com/gemini-testing/testplane/blob/3981b08c2eff48448e459cb2c65b72150c4cdaaf/src/browser/browser.ts#L105

src/browser/queries/wdio-types.ts
```diff
@@ -0,0 +1,30 @@
...
+/*
+eslint-disable @typescript-eslint/no-explicit-any
+*/
+
+export type SelectorsBase = {
+    $: WebdriverIO.Browser["$"];
+    $$: WebdriverIO.Browser["$$"];
+};
+
+export type BaseWithExecute = {
```
shadowusr, line 11: Why are these needed at all? We already have typings for execute, addCommand, etc everywhere.

src/browser/queries/index.ts
```diff
@@ -0,0 +1,265 @@
...
+@typescript-eslint/no-unsafe-argument
+*/
+
+function configure(config: Partial<Config>) {
+    _config = config;
+
+    console.log(_config);
+}
+
+export * from "./types.js";
```
shadowusr, line 264: Why use extension? We don't use ESM yet.

shadowusr (changes requested):

src/browser/queries/types.ts
talkenson: renamed to testplane

src/browser/queries/index.ts
talkenson: copyright advert added to the top of index.ts

shadowusr (changes requested):
Thoughts I have after taking a look:

- Let's come up with a better configuration API. `browser.configureDTL` isn't a good choice for multiple reasons (you need browser instance to configure, naming feels off, you can't set this globally)

- We need to check an important edge-case: what if users followed through with our docs and uses wdio-testing-library? Then if he updates, tests will be failing, because of duplicate commands. We need to check this

- We need to check that timeouts for `find*` queries are being configured correctly. For example, how timeouts specified in our testplane config, interact with those. If they don't it's an issue, because they should affect testing-library timeouts.

src/browser/types.ts
```diff
@@ -151,6 +154,11 @@ declare global {
             switchToRepl: (this: WebdriverIO.Browser, ctx?: Record<string, unknown>) => Promise<void>;
 
             clearSession: (this: WebdriverIO.Browser) => Promise<void>;
+
+            /**
+             * Allows to configure DOM Testing Library
+             */
+            configureDTL: typeof configure;
```
shadowusr: Here, we still have configureDTL command

src/index.ts
```diff
@@ -7,12 +7,14 @@ import "./browser/types";
...
 
 import { GlobalHelper } from "./types";
+
 export { run as runCli } from "./cli";
 export { Testplane as default } from "./testplane";
 export { Key } from "@testplane/webdriverio";
 export * from "./mock";
 
 export * as unstable from "./unstable";
+export * as queries from "./browser/queries";
```
shadowusr: I think it's better to remove this export, because it exports too much and I don't think items that are being exported will be popular among users. It's better to export as little as possible, because anything that becomes part of public API will need to be supported indefinitely.

talkenson:
I think now it looks okay, but feel free to ask for more changes

PR #1097, opened 2025-06-14
feat: implement captureDomSnapshot command

shadowusr:
### What's done?

Implemented a new command, `unstable_captureDomSnapshot`. Desided to give it `unstable` prefix to be able to change it without notice for a while. This command is, in fact, pretty stable and thoroughly tested on various complex pages like search results, habr articles, playground pages, etc.

High level overview how it works: we perform `browser.execute` and traverse the whole DOM tree while building an efficient DOM snapshot in our custom format. So this is pretty cross-browser.

The format of this snapshot is pretty compact. Each node is serialized into one line. All unnecessary tags and attributes are excluded. Form states, focus and hover states are captured as well.

What does the format look like? It's YAML, and below is a snapshot of ya.ru (you can also look at unit tests):
<details>

<summary>snapshot</summary>

```yaml
- body:
 - main.body__wrapper:
  - div.headline:
   - div.headline__personal:
    - a.home-link2.alice.headline__personal-item.home-link2_color_quaternary[aria-label=Алиса href="https://alice.yandex.ru/?utm_source=yandex&utm_campaign=morda_header_oknyx&utm_medium=interface"]:
     - svg.alice__icon
    - div:
     - div:
      - a.home-link2.headline__personal-item.headline__personal-item_services[aria-label=Сервисы href=https://yandex.ru/all title=Сервисы role=button]:
       - div.headline__personal-icon:
        - svg
    - a.home-link2.headline__personal-enter.home-link2_color_black[href="https://passport.yandex.ru/auth?retpath=https%3A%2F%2Fsso.passport.yandex.ru%2Fpush%3Fretpath%3Dhttp..."] "Войти"
    - div:
     - div.usermenu-portal:
      - div:
       - a.home-link2.headline__personal-menu.headline__personal-item[aria-label="Профиль, вход не выполнен" role=button tabindex=0 href="https://passport.yandex.ru/auth?origin=morda&retpath=https%3A%2F%2Fya.ru%2F%3Fnr%3D1%26redirect_ts%3..."]:
        - div.headline__personal-icon:
         - svg
   - div[@hidden]:
    - div.alerts[role=complementary aria-label=Уведомления]:
     - h1.a11y-hidden[@hidden] "Уведомления"
     - div.alert:
      - a.home-link2.alert__content.home-link2_color_inherit[href="https://yandex.ru/pogoda/146?lat=44.948237&lon=34.100318&utm_content=GroupedTomorrow&utm_medium=aler..."]:
[...]

PR #1098, opened 2025-06-14
feat: add ability to capture DOM snapshots of specific elements

shadowusr:
⚠️  This PR is built on top of #1097 — so only last commit needs review.

### What's done?

Added ability to capture snapshots of individual elements.

Example usage:
```typescript
const snapshotData = browser.unstable_captureDomSnapshot('.some-selector');

const anotherSnapshotData = browser.$('.another-selector').unstable_captureDomSnapshot();
```


src/browser/commands/captureDomSnapshot.ts
```diff
@@ -16,14 +16,39 @@ export interface CaptureSnapshotResult {
...
+    } else if (
+        selectorOrElementOrOptions &&
+        typeof selectorOrElementOrOptions === "object" &&
+        "tagName" in selectorOrElementOrOptions
+    ) {
+        element = selectorOrElementOrOptions as globalThis.Element;
+        options = maybeOptions || {};
+    } else {
+        selector = undefined;
+        element = undefined;
```
KuznetsovRoman: Lets use "null" instead of explicit "undefined"

src/browser/commands/captureDomSnapshot.ts
```diff
@@ -362,14 +387,39 @@ export const captureDomSnapshotInBrowser = ({
...
 
-    const startElement = document.body || document.documentElement;
-    if (!startElement) {
-        return {
-            snapshot: "# No elements found",
-            omittedTags: [],
-            omittedAttributes: [],
-            textWasTruncated: false,
-        };
+    let startElement: globalThis.Element | null;
```
KuznetsovRoman: What is "globalThis.Element"?

shadowusr: Replaced with just `Element`. At first thought it would be easier to distinguish it from `WebdriverIO.Element`.

PR #1099, opened 2025-06-16
fix: use webdriverio connection retries

KuznetsovRoman:


src/browser/new-browser.ts
```diff
@@ -176,7 +176,6 @@ export class NewBrowser extends Browser {
             capabilities,
             automationProtocol: devtools ? DEVTOOLS_PROTOCOL : config.automationProtocol,
             connectionRetryTimeout: config.sessionRequestTimeout || config.httpTimeout,
-            connectionRetryCount: 0, // testplane has its own advanced retries
```
KuznetsovRoman, line 179: No, it hasn't

PR #1103, opened 2025-06-20
fix: use RAF timing in rrweb snapshots

shadowusr:


src/browser/history/rrweb.ts
```diff
@@ -14,57 +14,91 @@ export async function installRrwebAndCollectEvents(
...
+                                    }
+
+                                    // @ts-expect-error
+                                    window.rrwebEvents.push(event);
+                                });
+                            },
                         });
+
+                        // @ts-expect-error
+                        window.rrweb.record.addCustomEvent("color-scheme-change", {
```
shadowusr, line 68: Тут только форматирование изменилось! Надо смотреть только на строки 22-26 и 48-63

sipayRT (approved):

src/browser/history/rrweb.ts
```diff
@@ -14,57 +14,91 @@ export async function installRrwebAndCollectEvents(
...
+            (rrwebRecordFnCode, serverTime) => {
                 // @ts-expect-error
-                if (!window.rrweb) {
-                    window.eval(rrwebRecordFnCode);
+                if (!window.__testplaneTiming) {
                     // @ts-expect-error
-                    window.lastProcessedRrwebEvent = -1;
-                    // @ts-expect-error
-                    window.rrwebEvents = [];
+                    window.__testplaneTiming = {
```
KuznetsovRoman, line 22: Do we always have "__geminiCore" in our global scope from client-bridge?
If yes, then maybe we should stick to using existing "__geminiCore" instead of creating new variables?
Some of our big projects track extra global variables and would throw error here

shadowusr: We completely re-write __geminiCore in client bridge. Also, we store time travel snapshots in separate field. I suggest refactoring all of the injected data in a separate PR.

PR #1111, opened 2025-07-16
fix(ct): correctly work in headless mode

DudaGod:
### What is done?

- we need to wait response from master when emit init event from worker

### DIFF
--- a/src/runner/browser-env/vite/socket.ts
+++ b/src/runner/browser-env/vite/socket.ts
@@ -66,11 +66,17 @@ function handleWorkerEvents(
     socket: Socket<ClientViteEvents, ViteClientEvents>,
     io: Server<ClientViteEvents, ViteClientEvents>,
 ): void {
-    socket.on(WorkerEventNames.initialize, payload => {
-        const { runUuid } = socket.handshake.auth;
-        socket.join(runUuid);
+    socket.on(WorkerEventNames.initialize, (payload, cb) => {
+        try {
+            const { runUuid } = socket.handshake.auth;
+            socket.join(runUuid);
 
-        WORKER_ENV_BY_RUN_UUID.set(runUuid, payload);
+            WORKER_ENV_BY_RUN_UUID.set(runUuid, payload);
+
+            cb(null);
+        } catch (err) {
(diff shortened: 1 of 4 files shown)


src/runner/browser-env/vite/socket.ts
```diff
@@ -66,11 +66,17 @@ function handleWorkerEvents(
...
+        try {
+            const { runUuid } = socket.handshake.auth;
+            socket.join(runUuid);
 
-        WORKER_ENV_BY_RUN_UUID.set(runUuid, payload);
+            WORKER_ENV_BY_RUN_UUID.set(runUuid, payload);
+
+            cb(null);
+        } catch (err) {
+            cb(prepareError(err as Error));
```
DudaGod, line 78: Now event in master return `null` or `error` to worker in order to inform that everything was completed and worker can open vite url. Previously, the worker opened the url without waiting and this led to errors.

src/worker/browser-env/runner/test-runner/index.ts
```diff
@@ -191,7 +191,7 @@ export class TestRunner extends NodejsEnvTestRunner {
                     this._handleRunExpectMatcher(browser, expectMatchers),
                 );
 
-                this._socket.emit(WorkerEventNames.initialize, {
+                const err = (await this._socket.timeout(SOCKET_MAX_TIMEOUT).emitWithAck(WorkerEventNames.initialize, {
```
DudaGod, line 194: `emitWithAck` - emitted event and wait response from master

test/src/worker/browser-env/runner/test-runner/index.ts
```diff
@@ -425,8 +439,6 @@ describe("worker/browser-env/runner/test-runner", () => {
                     const promise = run_();
                     await clock.tickAsync(1);
                     socket.emit(BrowserEventNames.initialize, []);
-
-                    await clock.tickAsync(1);
```
DudaGod, line 429: This test and test below fails in trunk after remove bluebird. I fix them by remove this timeout

sipayRT (approved):

PR #1112, opened 2025-07-19
chore: re-write screenshots logic to typescript

shadowusr:


src/browser/client-scripts/index.js
```diff
@@ -223,33 +396,270 @@ exports.resetZoom = function () {
...
+        return util.buildZChain(el);
+    });
+
+    // 3. Detect interfering elements
+    var root = document.documentElement;
+    var allElements = root.querySelectorAll ? root.querySelectorAll("*") : [];
+
+    var interferingRects = [];
+
+    allElements.forEach(function (el) {
```
KuznetsovRoman, line 475: This operation looks very heavy
Did you check "how much time it takes" on page with large amount of nodes?

src/browser/client-scripts/index.js
```diff
@@ -223,33 +396,270 @@ exports.resetZoom = function () {
...
+
+        // Skip invisible elements
+        if (br.width < 1 || br.height < 1) {
+            return;
+        }
+
+        // Skip elements that don't horizontally intersect with capture area
+        if (br.right <= captureAreaInViewportCoords.left || br.left >= captureAreaInViewportCoords.right) {
+            return;
+        }
```
KuznetsovRoman, line 498: What about vertical intersection?

src/browser/screen-shooter/utils.ts
```diff
@@ -0,0 +1,256 @@
...
+            params.y === 0 || resultScrollOffset.top + clientSize.height >= scrollSize.height;
+        const reachedHorizontalScrollLimit =
+            params.x === 0 || resultScrollOffset.left + clientSize.width >= scrollSize.width;
+
+        while (
+            resultScrollOffset.left === originalScrollOffset.left &&
+            resultScrollOffset.top === originalScrollOffset.top &&
+            !(reachedVerticalScrollLimit && reachedHorizontalScrollLimit)
+        ) {
+            if (iterations++ > 100000) {
```
KuznetsovRoman: Why 100000?

src/browser/screen-shooter/index.ts
```diff
@@ -0,0 +1,206 @@
...
+            this._selectorsToCapture,
+        ).catch(() => null);
+
+        debug("boundingRectBeforeScroll: %O", boundingRectsBeforeScroll);
+
+        const physicalScrollHeight = Math.max(
+            Math.min(nextNotCapturedArea.height, page.safeArea.height),
+            2 * page.pixelRatio,
+        );
+        const logicalScrollHeight = Math.ceil(physicalScrollHeight / page.pixelRatio) - 1;
```
KuznetsovRoman, line 126: why "-1"?

src/browser/screen-shooter/errors/vertical-overflow-error.ts
```diff
@@ -0,0 +1,18 @@
+import { Size } from "@testplane/webdriverio/build/commands/element";
+import { Rect } from "../../../image";
+
+export class VerticalOverflowError extends Error {
+    constructor(readableCaptureAreaDescr: string, captureArea: Rect, viewport: Size) {
+        const message = `Could not capture ${readableCaptureAreaDescr}, because it is larger than viewport height.
```
KuznetsovRoman, line 6: 👍

src/browser/screen-shooter/errors/horizontal-overflow-error.ts
```diff
@@ -0,0 +1,22 @@
+import type { Rect, Size } from "../../../image";
+
+export class HorizontalOverflowError extends Error {
+    constructor(readableCaptureAreaDescr: string, captureArea: Rect, viewport: Size) {
+        const message = `Could not capture ${readableCaptureAreaDescr}, because it is outside of horizontal viewport bounds.
```
KuznetsovRoman, line 5: 👍

src/browser/screen-shooter/index.ts
shadowusr: Added a comment. That's to account for fractional top edge of scroll container. If it's fractional, without -1 we'd get 1px lines on resulting screenshot. But this way, we'll never accidentally capture that top fractional edge.

src/browser/screen-shooter/utils.ts
shadowusr: Just to wait for the scroll to happen. In practice though we hardly ever need this. But I've lowered this value to 10,000.

Did not want to add async logic here, because support in old browsers like IE would be hard to achieve and IMO it's not worth the overcomplication

src/browser/client-scripts/index.js
shadowusr: We are dealing with vertical intersection below, everything below 500th line. Basically if element can't intersect horizontally we can be sure it won't intersect and can be safely skipped. But if it can potentially intersect, then we are interested only in certain fixed, sticky and absolute elements.

src/browser/client-scripts/index.js
shadowusr: Doesn't look too heavy to me... Average page has 5-10k elements, that's just 10k iterations with pretty cheap logic inside.

I've tried measuring how much time we spend on this on different pages and turns out it's between 20ms to 80ms.

But I think we can add an option to disable this logic in the future

PR #1113, opened 2025-07-22
chore: remove unused code

KuznetsovRoman:
We had:
- "Runner" pure abstract class at "runner/runner.ts" 
- "Runner" pure abstract class at "worker/runner/test-runner/runner.ts"
- "Runner" class at "worker/runner/index.js"

Now we only have one "Runner"

### DIFF
--- a/src/runner/browser-runner.ts
+++ b/src/runner/browser-runner.ts
@@ -1,6 +1,6 @@
 import { EventEmitter } from "events";
 import _ from "lodash";
-import { Runner } from "./runner";
+import { CancelableEmitter } from "./types";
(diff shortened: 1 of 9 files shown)


src/runner/browser-runner.ts
```diff
@@ -41,14 +40,6 @@ export class BrowserRunner extends Runner {
         return this._browserId;
     }
 
-    async run(testCollection: TestCollection): Promise<void> {
```
KuznetsovRoman, line 44: - Is never used
- Is not a part of public API

DudaGod (approved):

PR #1115, opened 2025-07-25
perf: speed up testplane launch

KuznetsovRoman:


package.json
```diff
@@ -88,6 +88,7 @@
     "gemini-configparser": "1.4.1",
     "get-port": "5.1.1",
     "import-meta-resolve": "4.0.0",
+    "load-esm": "1.0.2",
```
KuznetsovRoman, line 91: Contains like 1 LOC:
```js
module.exports = { loadEsm: (module) => import(module) };
```
In our project, typescript compiles all "import" to "require" and we can't prevent it
So, this package is used

src/browser-installer/constants.ts
```diff
@@ -20,6 +20,7 @@ export const MIN_CHROMIUM_MAC_ARM_VERSION = 93;
 export const MIN_CHROMIUM_VERSION = 73;
 export const MIN_FIREFOX_VERSION = 60;
 export const MIN_EDGEDRIVER_VERSION = 94;
+export const DRIVER_WAIT_INTERVAL = 25; // 20ms
```
KuznetsovRoman, line 23: We are almost blocked while waiting for webdriver launch
Checking "if port is busy" more often saves some time by price of some CPU cycles

src/browser-pool/basic-pool.ts
```diff
@@ -32,6 +32,7 @@ export class BasicPool implements Pool {
     }
 
     async getBrowser(id: string, opts: BrowserOpts = {}): Promise<NewBrowser> {
+        const { NewBrowser } = await import("../browser/new-browser");
```
KuznetsovRoman, line 35: "NewBrowser" takes some time to import
using lazy-load here (and in many other places) to reduce "time to require" and therefore - time to start workers, which is crucial

src/browser/commands/assert-view/errors/base-state-error.ts
```diff
@@ -1,4 +1,4 @@
-import { ImageInfo, RefImageInfo } from "../../../../types";
+import type { ImageInfo, RefImageInfo } from "../../../../types";
```
KuznetsovRoman, line 1: "import" -> "import type" also saves some time

src/browser/commands/assert-view/errors/image-diff-error.ts
```diff
@@ -110,6 +108,6 @@ export class ImageDiffError extends BaseStateError {
     }
 
     saveDiffTo(diffPath: string): Promise<null> {
-        return Image.buildDiff({ diff: diffPath, ...this.diffOpts });
+        return import("../../../../image").then(m => m.Image.buildDiff({ diff: diffPath, ...this.diffOpts }));
```
KuznetsovRoman, line 111: "Image" is fat, and we dont need to requrie "Image" module every time

src/browser/commands/switchToRepl.ts
```diff
@@ -60,7 +60,8 @@ export default (browser: Browser): void => {
         const testCwd = path.dirname(session.executionContext.ctx.currentTest.file!);
         process.chdir(testCwd);
 
-        const replServer = repl.start({ prompt: "> " });
+        const replServer = await import("node:repl").then(m => m.start({ prompt: "> " }));
```
KuznetsovRoman, line 63: given that "node:repl" only imported here, we could save ~10ms by lazy-loading it

src/browser/existing-browser.ts
```diff
@@ -82,6 +82,8 @@ export class ExistingBrowser extends Browser {
             this._startCollectingCustomCommands();
         }
 
+        const isolationPromise = this._performIsolation({ sessionCaps, sessionOpts });
+
         this._extendStacktrace();
         this._addSteps();
         this._addHistory();
```
KuznetsovRoman, line 87: These methods take ~40ms each
We could do it while performing isolation (CPU is idle, while waiting for browser to create new page...)

src/browser/existing-browser.ts
```diff
@@ -95,7 +97,10 @@ export class ExistingBrowser extends Browser {
             "testplane: init browser",
             async () => {
                 this._addCommands();
-                await this._performIsolation({ sessionCaps, sessionOpts });
+
+                await isolationPromise;
+
+                this._callstackHistory?.clear();
```
KuznetsovRoman, line 101: because order in "add history" and "perform isolation" becomes undefined, removing this step, if its completed

src/browser/new-browser.ts
```diff
@@ -71,23 +70,9 @@ export class NewBrowser extends Browser {
     async init(): Promise<NewBrowser> {
         this._session = await this._createSession();
 
-        this._extendStacktrace();
-        this._addSteps();
-        this._addHistory();
```
KuznetsovRoman, line 76: It can be omitted in new-browser

src/test-reader/sets-builder/index.ts
```diff
@@ -78,9 +78,12 @@ export class SetsBuilder {
                 .map(ignorePattern => path.resolve(projectRoot, ignorePattern));
         }
 
-        return this.#transformDirsToMasks()
-            .then(() => this.#resolvePaths(projectRoot))
-            .then(() => globExtra.expandPaths(this.#filesToUse, expandOpts, globOpts as { ignore: string[] }))
+        const resolvePathsPromise = this.#transformDirsToMasks().then(() => this.#resolvePaths(projectRoot));
+        const globExtraPromise = import("../../bundle/glob-extra");
```
KuznetsovRoman, line 82: lazy-loading in master in order to init workers earlier
preloading in worker because we are waiting for "runTest" command from master and we would need it anyway

src/testplane.ts
```diff
@@ -133,7 +135,11 @@ export class Testplane extends BaseTestplane {
         eventsUtils.passthroughEventAsync(signalHandler, this, MasterEvents.EXIT);
 
         await this._init();
+
         runner.init();
+
+        preloadWebdriverIO();
+
         await runner.run(
             await this._readTests(testPaths, { browsers, sets, grep, replMode }),
```
KuznetsovRoman: only start preloading wdio to be sure it does not affect time to start workers
then we can start preloading wdio, which would be done while cpu has some idle time

src/utils/preload-utils.ts
```diff
@@ -0,0 +1,13 @@
+import { loadEsm } from "load-esm";
+
+export const preloadWebdriverIO = async (): Promise<void> => {
+    await loadEsm("@testplane/webdriverio").catch(() => {});
```
KuznetsovRoman, line 8: `require("@testplane/webdriverio")` is cheap because it does not compile almost any of wdio code:
```js
exports.remote = async function(params, remoteModifier) {
  const { remote } = await import("./node.js");
  return remote(params, remoteModifier);
};
```

KuznetsovRoman, line 8: `require("@testplane/webdriverio")` is cheap because it does not compile almost any of wdio code:
```js
exports.remote = async function(params, remoteModifier) {
  const { remote } = await import("./node.js");
  return remote(params, remoteModifier);
};
```

src/utils/processor.js
```diff
@@ -22,9 +23,21 @@ process.on("unhandledRejection", (reason, p) => {
     ipc.emit(WORKER_UNHANDLED_REJECTION, { error });
 });
 
-module.exports = async (module, methodName, args, cb) => {
+preloadMochaReader();
+preloadTestTransformer();
+preloadWebdriverIO();
```
KuznetsovRoman, line 28: worker has some spare time, before master didn't send "runTest" command

src/utils/workers-registry.js
```diff
@@ -47,6 +47,8 @@ module.exports = class WorkersRegistry extends EventEmitter {
     }
 
     register(workerFilepath, exportedMethods) {
+        this._workerFarm.loadModule(workerFilepath, _.noop);
```
KuznetsovRoman: before:
- "workerFilePath" is compiled only on call

after:
- "workerFilePath" is compiled on register, so method would be called earlier

src/worker/runner/index.js
```diff
@@ -30,17 +28,25 @@ module.exports = class Runner extends AsyncEmitter {
...
             file,
             config: this._config.forBrowser(browserId),
             browserAgent,
             attempt,
         });
 
-        return runner.run({ sessionId, sessionCaps, sessionOpts, state });
+        runner.prepareBrowser({ sessionId, sessionCaps, sessionOpts, state });
+
+        const tests = await this._testParser.parse({ file, browserId });
```
KuznetsovRoman, line 45: in worker, preparing browser does not depend on test
so we can start preparing browser before parsing test file

src/worker/runner/test-runner/index.js
```diff
@@ -18,31 +18,55 @@ module.exports = class TestRunner {
...
     }
 
-    async run({ sessionId, sessionCaps, sessionOpts, state }) {
-        await this.prepareToRun({ sessionId, sessionCaps, sessionOpts, state });
+    assignTest(test) {
+        this._test = test.clone();
+        this._test.testplaneCtx = _.cloneDeep(test.testplaneCtx) || {};
+    }
+
+    async prepareBrowser({ sessionId, sessionCaps, sessionOpts, state }) {
```
KuznetsovRoman, line 37: floating promise, but does not throw
throws on "prepareToRun" with test and tool context to save compatibility

src/browser/new-browser.ts
```diff
@@ -119,9 +106,60 @@ export class NewBrowser extends Browser {
...
-        return remote(sessionOpts);
+        const fullSessionOpts = Object.assign(sessionOpts!, thirdPartyGridCustomOptions);
+        const session = await WebDriver.newSession(fullSessionOpts);
+
+        const detectedSessionEnvFlags = sessionEnvironmentDetector({
+            capabilities: session.capabilities,
+            requestedCapabilities: fullSessionOpts.capabilities,
+        });
+
+        return attach({
```
KuznetsovRoman, line 153: While we waiting webdriver session, webdriverio is being preloaded (preloading started after workers are created)
unfortunately, we have to use wdio even in master, because SESSION_START is emitted from master:
https://testplane.io/docs/v8/reference/testplane-events/#session_start

PR #1116, opened 2025-07-29
feat: implement devServer reuseExisting option

KuznetsovRoman:


src/dev-server/utils.ts
```diff
@@ -79,6 +79,38 @@ const defaultIsReadyFn = (response: Awaited<ReturnType<typeof globalThis.fetch>>
     return response.status >= 200 && response.status < 300;
 };
 
+export const probeServer = async (
+    // eslint-disable-next-line @typescript-eslint/ban-types
+    readinessProbe: Exclude<Config["devServer"]["readinessProbe"], Function>,
+): Promise<boolean> => {
+    if (typeof readinessProbe.url !== "string") {
+        throw new Error("devServer.readinessProbe.url should be set to url");
```
sipayRT, line 87: we can add the current passed value here

src/dev-server/index.ts
```diff
@@ -15,6 +15,22 @@ export const initDevServer: InitDevServer = async ({ testplane, devServerConfig,
         return;
     }
 
+    if (devServerConfig.reuseExisting) {
+        if (typeof devServerConfig.readinessProbe === "function" || !devServerConfig.readinessProbe.url) {
+            throw new Error(
+                "When 'reuseExisting' is set to 'true' in 'devServer' config, it is required to set 'devServer.readinessProbe.url'",
```
sipayRT: > When 'reuseExisting' is set...

When 'reuseExisting' **option** is set
...to set 'devServer.readinessProbe.url' **option**

test/src/dev-server/index.ts
```diff
@@ -169,4 +172,188 @@ describe("dev-server", () => {
...
+                assert.fail("Expected error to be thrown");
+            } catch (error) {
+                assert.match(
+                    (error as Error).message,
+                    /When 'reuseExisting' is set to 'true' in 'devServer' config, it is required to set 'devServer.readinessProbe.url'/,
+                );
+            }
+
+            assert.notCalled(spawnStub);
+            assert.notCalled(probeServerStub);
```
sipayRT, line 205: why do we need to check it here?

test/src/dev-server/index.ts
```diff
@@ -169,4 +172,188 @@ describe("dev-server", () => {
...
+                    timeouts: {
+                        waitServerTimeout: 30000,
+                        probeRequestTimeout: 1000,
+                        probeRequestInterval: 500,
+                    },
+                },
+            });
+
+            assert.calledOnceWith(probeServerStub, {
+                url: "http://localhost:3000",
```
sipayRT: you can move this object into a variable to avoid code duplication

test/src/dev-server/index.ts
```diff
@@ -169,4 +172,188 @@ describe("dev-server", () => {
...
+                isReady: customIsReady,
+                timeouts: {
+                    waitServerTimeout: 30000,
+                    probeRequestTimeout: 1000,
+                    probeRequestInterval: 500,
+                },
+            });
+            assert.calledWith(loggerStub.log, "Reusing existing dev server");
```
sipayRT, line 276: it would be better to check that customIsReady stub was called

test/src/dev-server/index.ts
KuznetsovRoman: "probeServerStub" throws if "readinessProbe" is function or if url is not set
So its important that "probeServerStub" is not called

test/src/dev-server/index.ts
KuznetsovRoman: It is a part of another module and should not be tested in "test/src/dev-server/index.ts"
I will write corresponding test in "dev-server/utils.ts"

src/dev-server/utils.ts
KuznetsovRoman: In this function we dont need to
This exception can't be shown to user: this function is only called when "readinessProbe.url" is string and it is checked in function, which is calling this

PR #1117, opened 2025-07-30
perf: decrease ts transformation latency

KuznetsovRoman:


package.json
```diff
@@ -80,6 +80,7 @@
     "debug": "2.6.9",
     "edgedriver": "5.6.1",
     "error-stack-parser": "2.1.4",
+    "esbuild": "0.25.8",
```
KuznetsovRoman, line 83: Direct dependency of "vite", which is already in dependencies

package.json
```diff
@@ -94,6 +95,7 @@
     "looks-same": "9.0.1",
     "micromatch": "4.0.5",
     "mocha": "10.2.0",
+    "pirates": "4.0.7",
```
KuznetsovRoman, line 98: Light:
pirates@4.0.7: 0 dependencies, 4.53 KiB packed, 12.27 KiB unpacked, 3.71 KiB bundled, 1.76 KiB minified, 978 B gzipped, 966 B deflated, 851 B brotlied

src/error-snippets/source-maps.ts
```diff
@@ -3,12 +3,17 @@ import url from "url";
...
+import { transformCode } from "../utils/typescript";
 import type { SufficientStackFrame, ResolvedFrame } from "./types";
 
 export const extractSourceMaps = async (
     fileContents: string,
     fileName: string,
 ): Promise<BasicSourceMapConsumer | null> => {
+    if (fileContents.indexOf(SOURCE_MAP_URL_COMMENT) === -1) {
+        fileContents = transformCode(fileContents, { sourceFile: fileName, sourceMaps: true });
+    }
```
KuznetsovRoman, line 15: here we have original source, but stacktrace refers to transformed code
So we are transforming original file the same way it was transformed and pull out source maps

src/error-snippets/source-maps.ts
```diff
@@ -22,8 +27,11 @@ export const extractSourceMaps = async (
             : fileContents.slice(sourceMapsStartIndex + SOURCE_MAP_URL_COMMENT.length, sourceMapsEndIndex);
 
     const sourceMaps = await getSourceCodeFile(url.resolve(fileName, sourceMapUrl));
+    const consumer = (await new SourceMapConsumer(sourceMaps)) as BasicSourceMapConsumer;
+
+    consumer.file = consumer.file || fileName;
```
KuznetsovRoman, line 32: file is missed for some reasom

package.json
DudaGod: is the version the same as in vite?

src/browser-installer/constants.ts
```diff
@@ -20,6 +20,7 @@ export const MIN_CHROMIUM_MAC_ARM_VERSION = 93;
 export const MIN_CHROMIUM_VERSION = 73;
 export const MIN_FIREFOX_VERSION = 60;
 export const MIN_EDGEDRIVER_VERSION = 94;
+export const DRIVER_WAIT_INTERVAL = 25; // 20ms
```
DudaGod: is the comment correct? I don't understand how 25 == 20ms?

src/browser/new-browser.ts
```diff
@@ -71,23 +72,9 @@ export class NewBrowser extends Browser {
...
-                callstack: this._callstackHistory!,
-                config: this._config,
-            },
-            "testplane: init browser",
-            async () => {
-                this._addCommands();
-                this.restoreHttpTimeout();
-                await this._setPageLoadTimeout();
-            },
-        );
```
DudaGod, line 90: Why you remove it?

src/browser/third-party-grid.ts
```diff
@@ -0,0 +1,152 @@
+// https://github.com/webdriverio/webdriverio/blob/ed92a6f2abfc55b3c4ae887a222371554a7ada61/packages/webdriverio/src/utils/driver.ts
+import type { Capabilities, Options } from "@testplane/wdio-types";
```
DudaGod: Why we copy this code from wdio?

DudaGod (approved):

src/browser-installer/constants.ts
KuznetsovRoman: > is the comment correct?

No

src/browser/new-browser.ts
KuznetsovRoman: Master process only creates webdriver session
All of those is not needed here, because tests are executed in workers

src/browser/third-party-grid.ts
KuznetsovRoman: in order to not import it from wdio.
This sequence:
- "import webdriver"
- "create session using webdriver, while also importing webdriverio"
- "attach to session with webdriverio"

Is faster, than:
- "import webdriverio"
- "create session with webdriverio"

Because in first case we are compiling webdriverio while waiting for session

package.json
KuznetsovRoman: vite depends on esbuild with "^" specifier, so it vite accepts any version

PR #1118, opened 2025-08-05
Keep browser alive after end of tests

sonic16x:


src/test-reader/index.ts
```diff
@@ -75,7 +80,7 @@ function validateTests(testsByBro: Record<string, Test[]>, options: TestReaderOp
         return;
     }
 
-    const stringifiedOpts = convertOptions(_.omit(options, "replMode"));
+    const stringifiedOpts = convertOptions(_.omit(options, "replMode", "replMode"));
```
shadowusr: This looks wrong, why do we have two replMode's here?

docs/debugging.md
shadowusr, line 1: We need a complete example on how to use new `attachToBrowser` method. At the very least in the release description, similar to this: https://github.com/gemini-testing/testplane/releases/tag/v8.28.4

shadowusr, line 1: Also we need to make sure to document this new CLI option on our website testplane.io

shadowusr, line 1: Found a critical bug in the `--keep-browser` behavior:

- When using `--keep-browser` option, then attaching to the browser and then calling `deleteSession()`, only browser is closed. Chromedriver process is left hanging
- When using `--keep-browser-on-fail` option, even if test passes and browser closes, chromedriver is not killed in the end

Overall, we need to think how we should think how these zombie-chromedriver processes should be handled. Maybe along with that chromedriver we should spawn a watchdog process that will kill it and itself when the browser itself is not running.

Or maybe at the very least state this caveat in the docs explicitly. I'm worried about a case when you run browser with `--keep-browser` in headful mode then close the browser by hand and get forever running chromedriver process as a result.

src/test-reader/index.ts
sonic16x: Fixed

docs/debugging.md
sonic16x: Fixed

PR #1119, opened 2025-08-14
chore: sharp -> jsquash

KuznetsovRoman:


src/utils/eight-bit-rgba-to-png.ts
```diff
@@ -0,0 +1,77 @@
...
+    pointer += resultBuffer.write("IEND", pointer, "ascii");
+    const iendCrc = crc32(Buffer.from(resultBuffer.buffer, iendPointer, pointer - iendPointer));
+    pointer = resultBuffer.writeUInt32BE(iendCrc, pointer);
+
+    if (pointer !== resultBuffer.byteLength) {
+        throw new Error("Got malformed input while trying to convert rgba to png");
+    }
+
+    return resultBuffer;
+};
```
DudaGod, line 77: Looks like we have exactly the same code in looks-same. It would be cool to put it in separate unpublished packages when we move to monorepo

PR #1129, opened 2025-09-05
fix: use own cdp implementation

KuznetsovRoman:


package.json
```diff
@@ -115,6 +115,7 @@
     "vite": "5.1.6",
     "wait-port": "1.1.0",
     "worker-farm": "1.7.0",
+    "ws": "8.18.3",
```
KuznetsovRoman, line 117: using this version because we already have it as "@testplane/webdriver@9.5.12" dependency

src/browser/cdp/emitter.ts
```diff
@@ -0,0 +1,23 @@
+import * as logger from "../../utils/logger";
+import { EventEmitter } from "events";
+
+export class CDPEventEmitter<Events extends { [key in keyof Events]: unknown }> extends EventEmitter {
```
KuznetsovRoman, line 4: This class provides typings for eventemitter, like that:
image
image

Also it wraps callbacks so they don't produce unhandled rejection

src/browser/cdp/index.ts
```diff
@@ -0,0 +1,45 @@
...
+import { CDPProfiler } from "./profiler";
+import type { CDPEvent } from "./types";
+
+export class CDP {
+    private readonly connection: CDPConnection;
+    public readonly target: CDPTarget;
+    public readonly profiler: CDPProfiler;
+
+    static async create(browser: Browser): Promise<CDP | null> {
+        if (!browser.publicAPI.isChromium) {
```
KuznetsovRoman, line 13: cdp is only available in chromium-based browsers

src/browser/cdp/profiler.ts
```diff
@@ -0,0 +1,80 @@
+import { CDPConnection } from "./connection";
+import { CDPEventEmitter } from "./emitter";
+import type { CDPDebuggerLocation, CDPSessionId, CDPScriptCoverage, CDPProfile } from "./types";
+
+interface StartPreciseCoverageRequest {
```
KuznetsovRoman, line 5: copied typings from https://chromedevtools.github.io/devtools-protocol/1-3/Profiler/

src/browser/cdp/utils.ts
```diff
@@ -0,0 +1,29 @@
...
+    attempt = 0,
+    factor = 2,
+    jitter = 100,
+}: { baseDelay?: number; attempt?: number; factor?: number; jitter?: number } = {}): Promise<void> => {
+    const delay = Math.round(baseDelay * factor ** attempt + Math.random() * jitter);
+
+    return new Promise(resolve => setTimeout(resolve, delay).unref());
+};
+
+export const extractRequestIdFromBrokenResponse = (message: string): number | null => {
```
KuznetsovRoman, line 12: as we have seen, we might get broken message, which can't be parsed with json.parse
This function extracts request "id", so we can stop waiting for response for it until timeout and just retry it

src/browser/cdp/ws-endpoint.ts
```diff
@@ -0,0 +1,52 @@
+import { debugCdp } from "./debug";
+import type { Browser } from "../types";
+
+const endpointsCache = new WeakMap<Browser, string>();
+
+export const getWsEndpoint = async (browser: Browser): Promise<string | null> => {
```
KuznetsovRoman, line 6: "1 priority" is our custom implementation: https://github.com/webdriverio/webdriverio/issues/9323

src/browser/existing-browser.ts
```diff
@@ -346,35 +353,27 @@ export class ExistingBrowser extends Browser {
...
         }
 
-        const puppeteer = await this._session.getPuppeteer();
-        const browserCtxs = puppeteer.browserContexts();
+        if (!this._cdp) {
+            logger.warn("Unable to establish CDP connection, skip performing isolation");
+            return;
+        }
 
-        const incognitoCtx = await puppeteer.createIncognitoBrowserContext();
```
KuznetsovRoman, line 352: "createIncognitoBrowserContext" -> "createBrowserContext"
https://github.com/puppeteer/puppeteer/issues/11834

src/browser/existing-browser.ts
```diff
@@ -346,35 +353,27 @@ export class ExistingBrowser extends Browser {
...
-            if (ctx.isIncognito()) {
-                await ctx.close();
-                continue;
-            }
-
-            for (const page of await ctx.pages()) {
-                await page.close();
-            }
-        }
+        await Promise.all(browserContextIds.map(contextId => cdpTarget.disposeBrowserContext(contextId)));
```
KuznetsovRoman: we dont have to close pages manually:

> Deletes a BrowserContext. All the belonging pages will be closed

Source: https://chromedevtools.github.io/devtools-protocol/1-3/Target/#method-disposeBrowserContext

Also, CDP (and our implementation of it) support parallel requests

test/src/browser/cdp/connection.ts
```diff
@@ -0,0 +1,475 @@
...
+    waitForConnection: Promise<WebSocket>;
+    connectionsCounter: number;
+    requestsCounter: number;
+    closeConnections: () => void;
+};
+
+let wsServerConnection: WebSocket | null = null;
+let wsServer: StubWebSocketServer | null = null;
+
+const createWsServer = (): StubWebSocketServer => {
```
KuznetsovRoman, line 22: using real websocket echo server, as its near to impossible to create working "Websocket" stub

test/src/browser/cdp/connection.ts
```diff
@@ -0,0 +1,475 @@
...
+    let requestsCounter = 0;
+
+    const hangingPromise = new Promise<WebSocket>(resolve => {
+        resolveHangingPromise = resolve;
+    });
+
+    Object.defineProperty(wss, "waitForConnection", { value: hangingPromise });
+    Object.defineProperty(wss, "connectionsCounter", { get: () => connectionsCounter });
+    Object.defineProperty(wss, "requestsCounter", { get: () => requestsCounter });
+    Object.defineProperty(wss, "closeConnections", { value: () => wss.clients.forEach(ws => ws.close()) });
```
KuznetsovRoman, line 36: mocked websocket server helpers

src/browser/cdp/connection.ts
KuznetsovRoman, line 1: Core implementation: CDP websocket connection

Features:
- Connection retries with exponential backoffs
- Connection timeouts
- Concurrent-safety: multiple parallel requests use the same connection
- Requests retries (if needed) with exponential backoffs
- Requests timeouts
- Automatically handles reconnects
- Detects network instability and keeps connection via ping/pong healthcheck

src/browser/cdp/connection.ts
```diff
@@ -0,0 +1,561 @@
...
+    DISCONNECTED, // Not connected, able to connect
+    CONNECTING, // Connection is being established
+    CONNECTED, // Connection established
+    CLOSED, // Connection is disposed and does not require reconnecting
+}
+
+type OnEventMessageFn = (cdpEventMessage: CDPEvent) => unknown;
+
+export class CDPConnection {
+    public onEventMessage: OnEventMessageFn | null = null;
```
KuznetsovRoman, line 32: Didn't want to make this class EventEmitter because of events overhead
Straight "onEventMessage" should be enough here

src/browser/cdp/connection.ts
```diff
@@ -0,0 +1,561 @@
...
+    private readonly cdpWsEndpoint: string;
+    private readonly headers?: Record<string, string>;
```
KuznetsovRoman: These 2 are stored for reconnects

src/browser/cdp/connection.ts
```diff
@@ -0,0 +1,561 @@
...
+    private onPong: (() => void) | null = null;
+    private pingInterval: ReturnType<typeof setInterval> | null = null;
+    private pingSubsequentFails = 0;
```
KuznetsovRoman: ping/pong connection keepalive/healthcheck

"onPong" stores listener for "pong" so we could remove listener later

src/browser/cdp/connection.ts
```diff
@@ -0,0 +1,561 @@
...
+    private onConnectionCloseFn: (() => void) | null = null; // Defined, if there is connection attempt at the moment
```
KuznetsovRoman: This callback function is called to stop waiting for ws connection (if state is "connecting") when ".close()" method is called

src/browser/cdp/connection.ts
```diff
@@ -0,0 +1,561 @@
...
+    private wsConnectionStatus: WsConnectionStatus = WsConnectionStatus.DISCONNECTED;
```
KuznetsovRoman: as this class handles reconnections, it manages own state:
```ts
enum WsConnectionStatus {
    DISCONNECTED, // Not connected, able to connect
    CONNECTING, // Connection is being established
    CONNECTED, // Connection established
    CLOSED, // Connection is disposed and does not require reconnecting
}
```
"Closed" state is terminal

src/browser/cdp/connection.ts
```diff
@@ -0,0 +1,561 @@
...
+    private wsConnection: WebSocket | null = null;
+    private wsConnectionPromise: Promise<WebSocket> | null = null;
```
KuznetsovRoman: this is used to ensure concurrent-safety

Multiple "request" are waiting for the same connection

src/browser/cdp/connection.ts
```diff
@@ -0,0 +1,561 @@
...
+    private requestId = 0;
+    private pendingRequests: Record<number, (response: Record<string, unknown> | CDPError) => void> = {};
```
KuznetsovRoman: Requests multiplexer
When request is sent, `pendingRequests[requestId]` stores callback to resolve/reject request promise

src/browser/cdp/connection.ts
```diff
@@ -0,0 +1,561 @@
...
+        }
+
+        this.wsConnectionStatus = WsConnectionStatus.CONNECTING;
+
+        this.wsConnectionPromise = (async (): Promise<WebSocket> => {
+            try {
+                for (let retriesLeft = CDP_CONNECTION_RETRIES; retriesLeft >= 0; retriesLeft--) {
+                    const result = await this.tryToEstablishWsConnection(this.cdpWsEndpoint);
+
+                    if (this.wsConnectionStatus === WsConnectionStatus.CLOSED) {
```
KuznetsovRoman: when connection is resolved, we might have a frame where connection is immediately closed. This fixes race condition, which could happen (extremely rare)

src/browser/cdp/connection.ts
```diff
@@ -0,0 +1,561 @@
...
+                                );
+                                this.tryToReconnect();
+                            }
+                        });
+
+                        return result;
+                    }
+
+                    if (!(result instanceof CDPError) || result instanceof CDPConnectionTerminatedError) {
+                        throw result;
```
KuznetsovRoman, line 212: "Error" and "CDPConnectionTerminatedError" should not be retried

src/browser/cdp/connection.ts
```diff
@@ -0,0 +1,561 @@
...
+    /** @description Handles websocket incoming messages, resolving pending requests */
+    private onMessage(data: RawData): void {
+        const message = data.toString("utf8");
+
+        debugCdp(`< ${message}`);
+
+        try {
+            const jsonParsedMessage: CDPMessage = JSON.parse(message);
+
+            if (!("id" in jsonParsedMessage)) {
```
KuznetsovRoman, line 255: message is considered as event if it has no request id in it

src/browser/cdp/connection.ts
```diff
@@ -0,0 +1,561 @@
...
+
+        if (!cdpWsEndpoint) {
+            throw new CDPError({ message: `Couldn't determine CDP endpoint for session ${sessionId}` });
+        }
+
+        return new this(cdpWsEndpoint, headers);
+    }
+
+    /** @description Tries to establish ws connection with timeout */
+    private async tryToEstablishWsConnection(endpoint: string): Promise<WebSocket | Error> {
```
KuznetsovRoman: Makes single try to establish connection

src/browser/cdp/connection.ts
```diff
@@ -0,0 +1,561 @@
...
+                resolve(err as Error);
+            }
+        });
+    }
+
+    /**
+     * @description creates ws connection with retries or returns existing one
+     * @note Concurrent requests with same params produce same ws connection
+     */
+    private async getWsConnection(): Promise<WebSocket> {
```
KuznetsovRoman: Retries "tryToEstablishWsConnection"

src/browser/cdp/connection.ts
```diff
@@ -0,0 +1,561 @@
...
+
+        if (this.requestId >= CDP_MAX_REQUEST_ID) {
+            this.requestId = 0;
+        }
+
+        return id;
+    }
+
+    /** @description establishes ws connection, sends request with timeout and waits for response */
+    private async tryToSendRequest(
```
KuznetsovRoman: Makes single try to send a request and resolves its response

src/browser/cdp/connection.ts
```diff
@@ -0,0 +1,561 @@
...
+                        WsConnectionStatus.DISCONNECTED,
+                    );
+                    this.tryToReconnect();
+                }
+            });
+        });
+    }
+
+    /** @description Performs high-level CDP request with retries and timeouts */
+    async request<T>(
```
KuznetsovRoman, line 391: retries "tryToSendRequest"

src/browser/cdp/connection.ts
```diff
@@ -0,0 +1,561 @@
...
+                        message,
+                        requestId,
+                    }),
+                );
+            }
+        }
+    }
+
+    /** @description Closes websocket connection, terminating all pending requests */
+    close(): void {
```
KuznetsovRoman, line 482: Should only be called once, when we would not need this connection in the future (for example, when deleting browser session)

src/browser/cdp/connection.ts
```diff
@@ -0,0 +1,561 @@
...
+            }
+        });
+
+        ws.on("pong", onPong);
+
+        const pingInterval = (this.pingInterval = setInterval(() => {
+            if (!this.isWebSocketActive(ws)) {
+                clearInterval(pingInterval);
+                return;
+            }
```
KuznetsovRoman: on each tick we have to check it first, because we might catch a frame where this callback is called when this pingInterval is no longer active because ws connection is reestablished

src/browser/cdp/index.ts
```diff
@@ -0,0 +1,45 @@
...
+import { CDPTarget } from "./target";
+import { CDPProfiler } from "./profiler";
+import type { CDPEvent } from "./types";
+
+export class CDP {
+    private readonly _connection: CDPConnection;
+    public readonly target: CDPTarget;
+    public readonly profiler: CDPProfiler;
+
+    static async create(browser: Browser): Promise<CDP | null> {
```
shadowusr, line 12: This may be controversial, but IMO `create` methods don't bring any value and only pollute codebase. But not critical to me

src/browser/cdp/connection.ts
```diff
@@ -0,0 +1,561 @@
...
+    private _requestId = 0;
+    private _pendingRequests: Record<number, (response: Record<string, unknown> | CDPError) => void> = {};
+
+    private constructor(cdpWsEndpoint: string, headers?: Record<string, string>) {
+        this._cdpWsEndpoint = cdpWsEndpoint;
+        this._headers = headers;
+    }
+
+    /** @description Creates CDPConnection without establishing it */
+    static async create(browser: Browser): Promise<CDPConnection> {
```
shadowusr, line 51: Here as well I think there's little value in create() method and in receiving our Browser god-object if we in fact only need headers and wsEndpoint. This increases code coupling. For example when we chose to expose this API, users will be forced to create our Browser object, when they could just pass the endpoint and it would work fine

Not critical

src/browser/cdp/connection.ts
```diff
@@ -0,0 +1,561 @@
...
+                const cdpConnectionInstance = this;
+                const ws = new WebSocket(endpoint, { headers: this._headers });
+
+                let isSettled = false;
+
+                const onTimeout = setTimeout(() => {
+                    ws.close();
+                    done(
+                        new CDPTimeoutError({
+                            message: `Couldn't establish CDP connection to "${endpoint}" in ${CDP_CONNECTION_TIMEOUT}ms`,
```
shadowusr, line 86: It would be really nice if we had more actionable errors here and in all other places as well. I think a good error message states clearly "what happenned and why" and "what can be done to fix". Imagining myself as a user receiving the "Couldn't establish CDP connection to ... in 1000ms", I'd be pretty puzzled.

Not critical, but very nice to have. I don't think that all these new errors are actionable right now.

src/browser/cdp/connection.ts
```diff
@@ -0,0 +1,561 @@
...
+                    this._onConnectionCloseFn = onConnectionCloseFn;
+                }
+
+                // eslint-disable-next-line
+                const cdpConnectionInstance = this;
+                const ws = new WebSocket(endpoint, { headers: this._headers });
+
+                let isSettled = false;
+
+                const onTimeout = setTimeout(() => {
```
shadowusr: This sounds like a handler, but is in fact a number, right? I'd call it onTimeoutId or something

src/browser/cdp/connection.ts
```diff
@@ -0,0 +1,561 @@
...
+            return ws;
+        }
+
+        if (this._wsConnectionStatus === WsConnectionStatus.CONNECTED && ws && ws.readyState !== ws.OPEN) {
+            this.closeWsConnection("CDP connection was in invalid state", WsConnectionStatus.DISCONNECTED);
+        }
+
+        this._wsConnectionStatus = WsConnectionStatus.CONNECTING;
+
+        this._wsConnectionPromise = (async (): Promise<WebSocket> => {
```
shadowusr, line 161: Can we move this iife to a helper/protected method with human readable name? The logic here is pretty hard, especially the part where you reassign wsConnectionPromise inside.

src/browser/cdp/connection.ts
```diff
@@ -0,0 +1,561 @@
...
+
+        if (this._requestId >= CDP_MAX_REQUEST_ID) {
+            this._requestId = 0;
+        }
+
+        return id;
+    }
+
+    /** @description establishes ws connection, sends request with timeout and waits for response */
+    private async tryToSendRequest(
```
shadowusr: IMO we had a great habit of prefixing private/protected fields and methods with `_`. I see no reason to give it up

src/browser/cdp/error.ts
```diff
@@ -0,0 +1,50 @@
...
+
+        this.name = this.constructor.name;
+        this.code = code;
+        this.requestId = requestId;
+    }
+
+    isNonRetryable(): boolean {
+        // JSON-RPC Protocol Errors
+        // CDP State/Execution Errors
+        return Boolean(this.code && ((this.code >= -32700 && this.code <= -32600) || this.code === -32000));
```
shadowusr, line 30: is there a doc somewhere with description of those magic codes? I think a link wouldn't hurt

src/browser/cdp/profiler.ts
shadowusr, line 1: I would move domain implementations to a separate directory "domains". Now, they are mixed with types, utils, etc.

src/browser/cdp/target.ts
```diff
@@ -0,0 +1,127 @@
...
+        const params = { url, browserContextId };
+
+        const res = await this._connection.request<CreateTargetResponse>("Target.createTarget", { params });
+
+        return res.targetId;
+    }
+
+    /** @link https://chromedevtools.github.io/devtools-protocol/1-3/Target/#method-activateTarget */
+    async activateTarget(targetId: CDPTargetId): Promise<void> {
+        return this._connection.request("Target.createTarget", { params: { targetId } });
```
shadowusr: I think this one's wrong, isn't it?

src/browser/cdp/profiler.ts
```diff
@@ -0,0 +1,80 @@
...
+        const res = await this._connection.request<StartPreciseCoverageResponse>("Profiler.startPreciseCoverage", {
+            sessionId,
+            params,
+        });
+
+        return res.timestamp;
+    }
+
+    async stopPreciseCoverage(sessionId: CDPSessionId): Promise<number> {
+        return this._connection.request("Profiler.stopPreciseCoverage", { sessionId });
```
shadowusr, line 83: Are you sure it returns a number? All other CDP methods seem to return objects

src/browser/existing-browser.ts
```diff
@@ -346,35 +353,29 @@ export class ExistingBrowser extends Browser {
...
-            // eslint-disable-next-line @typescript-eslint/no-explicit-any
-            const incognitoWindowId = windowIds.find(id => id.includes((page.target() as any)._targetId));
-
-            await this._session.switchToWindow(incognitoWindowId!);
+            await this._session.switchToWindow(windowIds.find(id => id.includes(incognitoWindowId))!);
+        } else {
+            await cdpTarget.activateTarget(incognitoWindowId);
         }
 
         if (this._session.isBidi) {
```
shadowusr, line 412: We got rid of this in master, because puppeteer could do cleanup of unnecessary contexts with bidi, too. Would it remain possible with our own implementation?

test/src/browser/cdp/connection.ts
```diff
@@ -0,0 +1,475 @@
...
+            // Simulate connection drop
+            getWsServerConnection().close();
+
+            // Make second request (should trigger reconnection)
+            const result2 = await connection.request("Successfull.Method");
+            assert.deepEqual(result2, { id: 3 });
+        });
+
+        it("should handle connection termination and reconnect", async () => {
+            const result1 = await connection.request("Successfull.Method");
```
shadowusr: minor typo — successful, not critical

src/browser/cdp/index.ts
KuznetsovRoman: Its async
Since we can't make async constructor, i think its better to encapsulate "get browser ws endpoint" logic to "cdp" scope

shadowusr: That's a common complaint, but you don't need async constructor. You can get all necessary values before calling the constructor and just pass all those values

src/browser/cdp/connection.ts
KuznetsovRoman: I think i will leave it be for now, and its not critical while it is internals
We can change it when (and if) we will decide to make it public

src/browser/cdp/connection.ts
KuznetsovRoman: I would love to leave actionable errors here, explaining "why it happened" and "what to do", if i knew, why could it happen and how to resolve these issues.

My plan to these kind of errors where you dont know, why it might happen, is the following:
- Release this version with non-actionable errors
- Saturate error messages with actionability, based on user stories (user comes with this error, we check, why it happened and what to do in this case, then update error message)

src/browser/cdp/connection.ts
KuznetsovRoman: > especially the part where you reassign wsConnectionPromise inside

Thats why i keep it inside, so we could see all complex `_wsConnectionPromise` transitions scoped to just one method, instead it being reassigned in multiple ones

src/browser/cdp/error.ts
KuznetsovRoman: Left a link to https://www.jsonrpc.org/specification#error_object

src/browser/cdp/profiler.ts
KuznetsovRoman: Its done in next PR

src/browser/existing-browser.ts
KuznetsovRoman: I will take a look at the code

Issue #1128, opened 2025-09-05
Ability to disable headless mode using cli opiton


DudaGod:
### Description

In a project created with `npm init testplane@latest` headless mode is enabled by default. If the user wants to turn it off for one run, then he must use the option `--healdess=false` or `--headless=0` or `--headless=no`. But it doesn't work. It is necessary to support this opportunity.

### Verify latest release

- [x] I verified that the issue exists in the latest Testplane release

### Which area(s) of Testplane are affected? (leave empty if unsure)

CLI

### Reproduction steps

- npm init testplane@latest (in some test folder)
- npx testplane --headless=false

### Actual Behavior

Get an error like:
```
<rejected> Error: "headless" option should be "new" or "old", but got "false"
// ...
```

### Expected Behavior

Headless mode is disabled.

### Which Node.js version are you using?

v18.12.1

PR #1134, opened 2025-09-12
fix: support headless=false from cli

pumano:


DudaGod:
### Problem

Described here - https://github.com/gemini-testing/testplane/issues/1128

### What is done

Added parse functions for option "headless" if it specified through cli or env. Support value: "1", "yes", "true" as true (boolean) and "0", "no", "false" as false (boolean).

I couldn't write a proper unit test because I don't have access to the functions `parseCli` and `parseEnv` of the configparser package.

### Special thanks

Thanks a lot to @pumano for his contribution.

PR #1138, opened 2025-09-22
fix: tests location when calling list-tests cli command

KuznetsovRoman:


src/utils/typescript.ts
```diff
@@ -106,14 +110,43 @@ export const registerTransformHook = (isSilent: boolean = false): void => {
...
+
+            revertTransformHook = addHook(
+                (code, filename) => transformCode(code, { sourceFile: filename, sourceMaps: true, isSilent }),
+                {
+                    exts: TRANSFORM_EXTENSIONS,
+                    matcher: filename => !filename.includes("node_modules"),
+                    ignoreNodeModules: false,
+                },
+            );
+        };
```
KuznetsovRoman, line 126: Usually we register transform hook without sourcemaps so we dont waste any time generating it

This function will re-register transform hook with sourcemaps

src/utils/typescript.ts
```diff
@@ -106,14 +110,43 @@ export const registerTransformHook = (isSilent: boolean = false): void => {
...
+            revertTransformHook();
+
+            revertTransformHook = addHook(
+                (code, filename) => transformCode(code, { sourceFile: filename, sourceMaps: true, isSilent }),
+                {
+                    exts: TRANSFORM_EXTENSIONS,
+                    matcher: filename => !filename.includes("node_modules"),
+                    ignoreNodeModules: false,
+                },
+            );
```
DudaGod: can we get rid of duplication here (only value of `sourceMaps` flag changes)?

test/src/utils/typescript.ts
```diff
@@ -48,4 +50,22 @@ describe("utils/typescript", () => {
...
     });
+
+    describe("enableSourceMaps", () => {
+        it("should not do anything if transform hook is not registered", () => {
+            ts.enableSourceMaps();
+
+            assert.notCalled(addHookStub);
+        });
+
+        it("should re-register transform hook with source maps", () => {
```
DudaGod, line 61: But you didn't check that the flag `sourceMap: true` is specified on second call

KuznetsovRoman: We cant.
We just pass callback `(code, filename) => transformCode(...)` to "addHook"
So "addHook" was just called with some function and thats all we can get.
"flag sourceMap: true" is internal detail of "registerTransformHook" function implementation

src/utils/typescript.ts
KuznetsovRoman: Done

PR #1139, opened 2025-09-22
Save and restore state

sonic16x:
Use this example for test
```typescript
describe("test saveState and restoreState", () => {
    it("saveStorage", async ({browser}) => {
        await browser.url("https://authenticationtest.com/simpleFormAuth/");

        const emailInput = await browser.$("#email");
        await emailInput.setValue("simpleForm@authenticationtest.com");

        const passwordInput = await browser.$("#password");
        await passwordInput.setValue("pa$$w0rd");

        const logInButton = await browser.$('[type="submit"]');
        await logInButton.click();

        const title = await browser.$("h1");

        await expect(await title.getText()).toBe("Login Success");

        await browser.saveState({
            path: path.join(__dirname, 'dump_login.json'),
        });
    })

    it("restoreState", async ({browser}) => {
        await browser.url("https://authenticationtest.com/coverage/?id=1");

        const navbarText = await browser.$(".navbar-text");
        expect(await navbarText.getText()).toBe("Please Sign In");

        await browser.restoreState({
            path: path.join(__dirname, 'dump_login.json'),
        });

        await browser.url("https://authenticationtest.com/coverage/?id=1");

        const navbarTextLink = await browser.$(".navbar-text a");
        expect(await navbarTextLink.getText()).toBe("Sign Out");
    })
})
```


test/integration/standalone/standalone-save-state.test.ts
```diff
@@ -0,0 +1,89 @@
...
+        await browser.url("http://localhost:3000/");
+
+        status = await browser.$("#status");
+
+        // check that we are not logged in
+        assert.strictEqual(await status.getText(), "You are not logged in");
+    });
+
+    it("saveState", async function () {
+        // input login
```
shadowusr, line 47: Too many comments by AI =)

test/integration/standalone/standalone-save-state.test.ts
```diff
@@ -0,0 +1,89 @@
+import { strict as assert } from "assert";
+import { launchBrowser } from "../../../src/browser/standalone";
+import { BROWSER_CONFIG } from "./constants";
+import { SaveStateData } from "../../../src";
+
+import { AuthServer } from "./mock-auth-page/server";
+import process from "node:process";
+
+describe("Standalone Browser E2E Tests", function () {
```
shadowusr: Let's name this suite accordingly, like "saveState and restoreState tests"

src/browser/commands/restoreState/clearAllIndexedDB.ts
```diff
@@ -0,0 +1,27 @@
+export async function clearAllIndexedDB(): Promise<void> {
```
shadowusr: This is not used anywhere

src/browser/commands/restoreState/index.ts
```diff
@@ -0,0 +1,144 @@
+import fs from "fs-extra";
+import _ from "lodash";
+
+// import { clearAllIndexedDB } from "./clearAllIndexedDB";
```
shadowusr: Comments

src/browser/commands/restoreState/index.ts
```diff
@@ -0,0 +1,144 @@
...
+                            if (frameData.sessionStorage && options.sessionStorage) {
+                                await session.execute<void, [Record<string, string>, "sessionStorage"]>(
+                                    restoreStorage,
+                                    frameData.sessionStorage,
+                                    "sessionStorage",
+                                );
+                            }
+
+                            // @TODO: will make it later
+                            // if (frameData.indexDB && options.indexDB) {
```
shadowusr: comment

src/browser/commands/restoreState/index.ts
```diff
@@ -0,0 +1,144 @@
...
+
+// import { clearAllIndexedDB } from "./clearAllIndexedDB";
+// import { restoreIndexedDB } from "./restoreIndexedDB";
+import { restoreStorage } from "./restoreStorage";
+
+import * as logger from "../../../utils/logger";
+import type { Browser } from "../../types";
+import { DEVTOOLS_PROTOCOL, WEBDRIVER_PROTOCOL } from "../../../constants/config";
+import { defaultOptions, getWebdriverFrames, SaveStateData, SaveStateOptions } from "../saveState";
+import { Protocol } from "devtools-protocol";
```
shadowusr: we don't have it in deps

src/browser/commands/restoreState/restoreIndexedDB.ts
```diff
@@ -0,0 +1,67 @@
+import { DumpIndexDB, DumpStoreIndexDB } from "../saveState/dumpIndexedDB";
```
shadowusr: Do we support indexed db or not? As far as I remember we agreed to give up on it for now and come back to it in the future. If we decided not to support it, let's get rid of unused code related to indexed db

src/browser/commands/saveState/index.ts
```diff
@@ -0,0 +1,174 @@
...
+                            frameData.localStorage = localStorage;
+                        }
+
+                        if (sessionStorage && options.sessionStorage) {
+                            frameData.sessionStorage = sessionStorage;
+                        }
+                    }
+
+                    // @TODO: will make it later
+                    // if (options.indexDB) {
```
shadowusr: unnecessary comment

src/browser/commands/saveState/index.ts
```diff
@@ -0,0 +1,174 @@
...
+                        value: cookie.value.value,
+                        domain: cookie.domain,
+                        path: cookie.path,
+                        expires: cookie.expiry,
+                        httpOnly: cookie.httpOnly,
+                        secure: cookie.secure,
+                    }));
+                }
+
+                await session.switchToParentFrame();
```
shadowusr, line 89: I don't understand why switch to parent frame here, then do it again on line 114

Moreover, I don't understand the logic here.
1. We get N iframes at the current level context.
2. We switch to parent frame N times (???)
3. We iterate from -1 to N (?)

And another issue is that we don't switch back to the frame user had active before. So calling saveState actually mutates browser state, while it should not.

Shouldn't we have about the following algorithm?
1. Save current frame id and start from top-level browser context
2. Recursively:
    1. Save storage values on current level
    2. Iterate over all frames at current level and do the same thing
3. Restore current frame id

src/browser/commands/saveState/dumpIndexedDB.ts
```diff
@@ -0,0 +1,106 @@
+export interface DumpIndexDB {
```
shadowusr: Not clear if we need this too

src/browser/commands/restoreState.ts
```diff
@@ -0,0 +1,4 @@
+import restoreState from "./restoreState/index";
```
shadowusr: What's the purpose of these files?

package.json
```diff
@@ -32,7 +32,7 @@
     "prettier-watch": "onchange '**' --exclude-path .prettierignore -- prettier --write {{changed}}",
     "test-unit": "_mocha \"test/!(integration)/**/*.js\"",
     "test": "npm run test-unit && npm run check-types && npm run lint",
-    "test-integration": "mocha -r ts-node/register -r test/integration/standalone/preload-browser.fixture.ts test/integration/standalone/standalone.test.ts",
+    "test-integration": "mocha -r ts-node/register -r test/integration/standalone/preload-browser.fixture.ts test/integration/standalone/standalone.test.ts test/integration/standalone/standalone-save-state.test.ts",
```
shadowusr: Let's maybe use a mask here, if we want to run more than one file now

src/browser/existing-browser.ts
```diff
@@ -60,6 +62,7 @@ export class ExistingBrowser extends Browser {
     protected _meta: Record<string, unknown>;
     protected _calibration?: CalibrationResult;
     protected _clientBridge?: ClientBridge;
+    private allCookies: Map<string, Protocol.Network.CookieParam> = new Map();
```
shadowusr: Let's keep using the `_` prefix for private/protected fields for consistency

src/browser/existing-browser.ts
```diff
@@ -111,6 +118,67 @@ export class ExistingBrowser extends Browser {
...
     }
 
+    async startCollectCookies(): Promise<void> {
+        if (!this._session) {
+            return;
+        }
+
+        this.allCookies = new Map();
+
+        this.publicAPI.addCommand("getAllRequestsCookies", async () => {
```
shadowusr: Here, we add a whole new command to user-facing API. Is this intentional?

How is it different from `getAllCookies`? The names are very similar and it's confusing to have them both.

I see we are using it in saveState. Perhaps we should make it a method of existing browser class that's unavailable to users instead of a new command?

src/browser/existing-browser.ts
```diff
@@ -111,6 +118,67 @@ export class ExistingBrowser extends Browser {
...
+            }));
+        });
+
+        const puppeteer = await this._session.getPuppeteer();
+
+        if (puppeteer) {
+            const pages = await puppeteer.pages();
+
+            if (pages.length) {
+                pages[0].on("response", async res => {
```
shadowusr: Why only the first page? Do we need to unsubscribe from these events at some point?

Let's reduce code nesting level by reversing conditions: `if(!puppeteer) {return;}`, etc.

src/browser/commands/restoreState/index.ts
```diff
@@ -0,0 +1,144 @@
...
+        }
+
+        switch (browser.config.automationProtocol) {
+            case WEBDRIVER_PROTOCOL: {
+                if (restoreState.cookies && options.cookies) {
+                    await session.setCookies(restoreState.cookies);
+                }
+
+                if (restoreState.framesData) {
+                    await session.switchToParentFrame();
```
shadowusr, line 68: Same thing as in saveState, what's with switching to parent frames? It doesn't look correct

test/integration/standalone/standalone-save-state.test.ts
sonic16x: It is not AI :D

test/integration/standalone/standalone-save-state.test.ts
sonic16x: Fixed

src/browser/commands/restoreState/clearAllIndexedDB.ts
sonic16x: Removed

src/browser/commands/restoreState/index.ts
sonic16x: Removed

src/browser/commands/restoreState/index.ts
sonic16x: Removed

src/browser/commands/restoreState/index.ts
sonic16x: Fixed

src/browser/commands/restoreState/restoreIndexedDB.ts
sonic16x: Removed

src/browser/commands/saveState/index.ts
sonic16x: Removed

src/browser/commands/saveState/index.ts
sonic16x: Need switch to parent frame every time because new call switchFrame will be in current child frame

src/browser/commands/saveState/dumpIndexedDB.ts
sonic16x: Removed

src/browser/commands/restoreState/index.ts
sonic16x: Before iterations by frames also need switch to parent once

src/browser/commands/restoreState.ts
sonic16x: Removed

package.json
sonic16x: Fixed

src/browser/existing-browser.ts
sonic16x: Fixed

src/browser/existing-browser.ts
sonic16x: Fixed

src/browser/existing-browser.ts
sonic16x: Fixed and make special function for detect active page in pupeter (popular solution from internet)

src/browser/types.ts
```diff
@@ -76,6 +79,10 @@ declare global {
 
             getConfig(this: WebdriverIO.Browser): Promise<BrowserConfig>;
 
+            getAllRequestsCookies(): Promise<Array<Cookie>>;
```
shadowusr: We don't need this one

src/browser/commands/restoreState/index.ts
```diff
@@ -0,0 +1,156 @@
...
+
+export type RestoreStateOptions = SaveStateOptions & {
+    data?: SaveStateData;
+    refresh?: boolean;
+};
+
+export default (browser: Browser): void => {
+    const { publicAPI: session } = browser;
+
+    session.addCommand("restoreState", async (_options: RestoreStateOptions) => {
```
shadowusr, line 26: Would be neat to handle those `Request failed with status 400 due to invalid cookie domain: invalid cookie domain:` errors with user-friendly error messages.

These are thrown when trying to call restoreState on `about:blank` or url that's not matching one used in `saveState`.

src/browser/types.ts
sonic16x: Fixed

src/browser/commands/restoreState/index.ts
sonic16x: Fixed

PR #1144, opened 2025-09-26
feat: ability to connect to repl server by port

DudaGod:
### What is done

- added the `--repl-port` cli option in order to be able to specify port on which create net server to be able to connect to repl server. If port is not specified then free port is engaged
- add broadcast messages from all sockets and main terminal in order to be synchronized


src/browser/commands/switchToRepl.ts
```diff
@@ -54,23 +62,62 @@ export default (browser: Browser): void => {
...
+            chalk.yellow(
+                `You have entered to REPL mode via terminal (test execution timeout is disabled). Port to connect to REPL from other terminals: ${runtimeCfg.replMode.port}`,
+            ),
+        );
 
         const currCwd = process.cwd();
         const testCwd = path.dirname(session.executionContext.ctx.currentTest.file!);
         process.chdir(testCwd);
 
-        const replServer = await import("node:repl").then(m => m.start({ prompt: "> " }));
```
KuznetsovRoman, line 63: We should return dynamic import here.

When its static, like you did it now, it is imported on each testplane run, even when repl is not used.

So it should be dynamic, so we could save some time when we dont need repl feature.

KuznetsovRoman (changes requested):

PR #1145, opened 2025-10-07
fix(repl): show error message without trace

sipayRT:


src/test-reader/index.ts
```diff
@@ -9,6 +9,20 @@ import type { Config } from "../config";
...
+    constructor(modeName: string, testsCount: number, browsersToRun: string[]) {
+        const message =
+            `In ${modeName} only 1 test in 1 browser should be run, but found ${testsCount} tests` +
+            `${testsCount === 0 ? ". " : ` that run in ${browsersToRun.join(", ")} browsers. `}` +
+            `Try to specify cli-options: "--grep" and "--browser" or use "testplane.only.in" in the test file.`;
+
+        super(message);
+        this.name = "SingleTestModeError";
+        // Удаляем стектрейс про EPIPE, который не нужен пользователю
+        delete this.stack;
```
KuznetsovRoman, line 22: By deleting "stack" property you are producing invalid Error object:

Specs: https://nodejs.org/api/errors.html#errorstack

If you want to override stack, make it comply node.js stacktraces:

> The first line is formatted as <error class name>: <error message>, and is followed by a series of stack frames (each line beginning with "at ").

So, we could overwrite it with `SingleTestModeError: ${message}`

KuznetsovRoman (changes requested):

src/test-reader/index.ts
sipayRT: yes, but in that case the message will be duplicated

KuznetsovRoman (approved):

src/test-reader/index.ts
KuznetsovRoman: If it doesn't lead to an error, its fine, i guess

PR #1230, opened 2026-04-06
feat(selectivity): add ignoreSourceMapUrls config option

readyyyk:
Allow skipping source map fetching for specific script/stylesheet URLs matching glob patterns (via micromatch). This prevents errors when CDN dependencies reference source maps that don't exist.


KuznetsovRoman:
Implemented in https://github.com/gemini-testing/testplane/pull/1246
Testplane v8.44.0

PR #1238, opened 2026-04-14
feat: rewrite composite-image and implement robust unit tests

shadowusr:
### What's done?
- Rewritten the composite image renderer almost from ground up
- Composite image now can handle all sorts of unexpected things, including: element size changes mid capture, overlapping segments, segments coming in random order, when one element moves and the other doesn't and many others.
- Overall, composite image now produces best possible screenshots based on observed data
- Implemented a set of "honest" unit tests that test all these edge cases using fixture images that are generated before launching unit tests


src/browser/screen-shooter/composite-image/debug-utils.ts
shadowusr, line 1: This file is used only for debugging, you probably don't want to spend time reviewing it. It can produce images like this when `TESTPLANE_DEBUG_SCREENSHOTS` is set:

Image

Note the green and red rects. Green means safe area, red means area we want to capture.

Very useful to visually spot when something goes wrong during debug.

test/src/browser/screen-shooter/composite-image/fixtures/generate.ts
shadowusr, line 1: This file generates a set of scenarios before running tests. Each scenario is a set of following:
- Capture areas 
- Ignore areas
- Safe areas
- Actual viewport images

And in the end we verify that result matches the expectation.

Example result of generated fixtures you can view over here: https://github.com/gemini-testing/testplane/commit/ad8f3222027597c63f8dc1032a6723fd3ab9f404

src/browser/screen-shooter/composite-image/debug-utils.ts
```diff
@@ -0,0 +1,180 @@
...
+    visibleCoveringRect: Rect<"viewport", "device">,
+    destinationDirPath: string | null,
+): Promise<void> {
+    if (!process.env.TESTPLANE_DEBUG_SCREENSHOTS || !destinationDirPath) {
+        return;
+    }
+
+    try {
+        const viewportDebugDir = path.join(destinationDirPath, `viewports`);
+        fs.mkdirSync(viewportDebugDir, { recursive: true });
```
KuznetsovRoman, line 145: Throws, if directory exist

src/browser/screen-shooter/composite-image/index.ts
```diff
@@ -1,244 +1,611 @@
...
+            if (previousSpecHeight !== currentSpecHeight) {
+                return previousSpecHeight - currentSpecHeight;
             }
         }
+
+        return null;
+    }
+
+    private _shouldTreatHeightChangeAsStartShift(heightDelta: number | null): heightDelta is number {
+        const shouldShiftFromStart = true;
```
KuznetsovRoman, line 316: Looks like an error
Leftover from local debug?

src/browser/screen-shooter/composite-image/debug-utils.ts
```diff
@@ -0,0 +1,180 @@
...
+import { CaptureSpec } from "../../client-scripts/screen-shooter/types";
+
+type DebugRectColor = { r: number; g: number; b: number; a: number };
+
+export type ViewportDebugRect = {
+    rect: Rect<"viewport", "device">;
+    color: DebugRectColor;
+};
+
+/* This file is used for debugging purposes only, to produce images with capture areas, safe areas, etc. visible when TESTPLANE_DEBUG_SCREENSHOTS is set */
```
KuznetsovRoman, line 16: Can we also include short description from PR comment here?

```
Green means safe area, red means area we want to capture.
```

Also, what is the color "r: 255, g: 105, b: 180"?

src/browser/screen-shooter/composite-image/index.ts
```diff
@@ -1,244 +1,611 @@
...
+    getMinCoord,
+    getHeight,
+    intersectYBands,
+    getMaxLength,
+    intersectXBands,
+    Length,
+    fromCaptureAreaToViewport,
+    fromViewportToCaptureArea,
+    getCoveringRect,
+} from "../../isomorphic/geometry";
```
KuznetsovRoman, line 25: What is "isomorphic/geometry"?
"src/browser/screen-shooter/composite-image" + "../../isomprphic/geometry" = "src/browser/isomorphic", but "src/browser/isomorphic" does not exist on this branch

src/image.ts
```diff
@@ -126,19 +139,22 @@ export class Image {
         let bufferPointer = 0;
         let sourceOffset = (rect.top * this._width + rect.left) * RGBA_CHANNELS;
 
-        const bytesToCopy = Math.min(rect.width, this._width - rect.left) * RGBA_CHANNELS;
+        const actualWIdth = Math.min(rect.width, this._width - rect.left);
```
KuznetsovRoman, line 142: `actualWIdth` -> `actualWidth` typo

test/src/browser/screen-shooter/composite-image/fixtures/generate.ts
```diff
@@ -0,0 +1,928 @@
...
+            file: path.relative(rootDir, chunkPath),
+            safeArea,
+            captureSpecs: [{ full: captureBoundingRect, visible: captureBoundingRect }],
+            ignoreBoundingRects,
+        });
+    }
+
+    const expectedImage = await compositeImage.render();
+    const expectedPath = path.posix.join(scenarioDir, "expected.png");
+    await expectedImage.save(expectedPath);
```
KuznetsovRoman, line 448: So, you are using "compositeImage.render" in order to generate images right before unit tests
Which are then compared to "actual", generated by another "compositeImage.render" from the same code base?

Looks like any change in "compositeImage.render" logic will not fail test, because "expected" and "actual" are generated from basically the same code

KuznetsovRoman (changes requested):

src/browser/screen-shooter/composite-image/debug-utils.ts
shadowusr: Not with `recursive: true` =)

src/browser/screen-shooter/composite-image/index.ts
shadowusr: Removed the constant. I was going to implement a more complex logic here, but looks like it's enough for now

src/browser/screen-shooter/composite-image/debug-utils.ts
shadowusr: It's pink-ish. Left a comment!

Screenshot 2026-04-17 at 1 15 18 AM

src/browser/screen-shooter/composite-image/index.ts
shadowusr: I think we've figured it out in chat — this directory does exist on this branch

test/src/browser/screen-shooter/composite-image/fixtures/generate.ts
shadowusr: Huh! Nice catch — thanks for spotting that. It's been a lot of itertions and changing things back and forth, apparently at some point I've lost track where those fixtures are coming from =)

Commited those fixtures instead of generating them on the fly.

shadowusr:
Fixed all of your issues in these commits:
- https://github.com/gemini-testing/testplane/pull/1242/commits/03be04897432f782f8320d55508c0cf9151f8b57
- https://github.com/gemini-testing/testplane/pull/1242/commits/0201a34be2c48de6eae029a31158e4832c791c40

They are attached to the tip of my changes. Though not ideal, it's more reliable to apply changes on top of all commits as opposed to applying them in the middle. But I agree this is a weird flow overall =(

Could you please have another look at this? @KuznetsovRoman

KuznetsovRoman (approved):

PR #1240, opened 2026-04-14
feat: improve and refactor screen-shooter and split it into 3 versions

shadowusr:
### What's done?
- There are now 3 types of screen-shooters:
  - Elements screenshooter
  - Page screenshooter
  - Viewport screenshooter
- Elements screenshooter now includes various stability enhancements:
  - It can wait for unstable capture are to settle
  - It can retry if capture area moves or changes its size
  - It can capture elements with unstable sizes

Overall, we have 3 paths in elements screen shooter:
1. "Happy" path, if element position is stable, we start capturing it right away. If element position/size changes, we go to (2).
2. "Unstable" path. Here, we first verify area stability by doing this:
    1. Scroll through the whole area, saving capture area state at each offset, getting "left" checkpoints
    2. Do it again, getting "right" checkpoints
    3. Compare them, if they match, proceed, it means area changes its size, but deterministically
    4. If they don't, throw and try again
3. Finally, there's "best effort" path in the end. If capture area is unstable in size and we can't get 100% correct screenshot, we do what we can and capture what's possible, but this screenshot may contain seams/black stripes. In practice though, it's rare.

These stability checks come with a perf cost. But for now, it's the only guaranteed way to handle highly complex and unstable pages even on exotic browsers like ios safari.

Example:
Screenshot 2026-04-14 at 5 08 49 PM


src/browser/commands/assert-view/index.js
```diff
@@ -31,7 +32,21 @@ const getIgnoreDiffPixelCountRatio = value => {
 };
 
 module.exports.default = browser => {
-    const screenShooter = ScreenShooter.create(browser);
+    const browserProperties = {
+        isWebdriverProtocol: browser.isWebdriverProtocol,
+        shouldUsePixelRatio: browser.shouldUsePixelRatio,
+        needsCompatLib: browser.needsCompatLib,
+    };
```
DudaGod, line 39: Isn't it easier:

```
const {isWebdriverProtocol, shouldUsePixelRatio, needsCompatLib} = browser;
const browserProperties = {isWebdriverProtocol, shouldUsePixelRatio, needsCompatLib};
```

src/browser/commands/assert-view/index.js
```diff
@@ -31,7 +32,21 @@ const getIgnoreDiffPixelCountRatio = value => {
...
+    const screenShooterPromise = ElementsScreenShooter.create({
```
DudaGod, line 40: why not `elementsScreenshoterPromise` ?

src/browser/commands/assert-view/index.js
```diff
@@ -153,11 +159,44 @@ module.exports.default = browser => {
...
+    const assertView = async (state, selectors, opts) => {
+        opts = getDefaultOpts(opts);
+
+        let debugId = "debugId";
+        try {
+            const test = session.executionContext.ctx.currentTest;
+            debugId = `${test.fullTitle()}.${browser.id}.${state}`;
+            opts.debugId = debugId;
+        } catch {
+            /**/
```
DudaGod, line 171: why it's empty?

src/browser/commands/assert-view/index.js
```diff
@@ -153,11 +159,44 @@ module.exports.default = browser => {
         testplaneCtx.assertViewResults.add({ stateName: state, refImg: refImg });
     };
 
+    const assertView = async (state, selectors, opts) => {
+        opts = getDefaultOpts(opts);
+
+        let debugId = "debugId";
+        try {
+            const test = session.executionContext.ctx.currentTest;
+            debugId = `${test.fullTitle()}.${browser.id}.${state}`;
```
DudaGod, line 168: In which case we can fail here?

src/browser/screen-shooter/operations/pointer-events.ts
```diff
@@ -0,0 +1,42 @@
...
+            pointerDebug("Skipping pointer move because session is not W3C");
+        }
+
+        try {
+            pointerDebug("Trying to move pointer to reset hover states");
+            await browser
+                .action("pointer", { parameters: { pointerType: "mouse" } })
+                .move({ duration: 0, origin: "viewport", x: 10, y: 10 })
+                .pause(10)
+                .move({ duration: 0, origin: "viewport", x: 0, y: 0 })
```
DudaGod, line 35: Why we move first on 10x10 and after pause to 0x0? Why not immediately set to 0x0 without 10x10?

src/browser/screen-shooter/elements-screen-shooter.ts
```diff
@@ -0,0 +1,876 @@
...
+            throw new Error("No selectors to capture passed to ElementsScreenShooter.capture");
+        }
+
+        let retriesCount = 0;
+        // This error may never be thrown, but just in case.
+        let originalError: unknown = new Error(
+            `An unknown error happened while capturing screenshot for selectors: ${JSON.stringify(selectorsToCapture)}`,
+        );
+
+        while (retriesCount < retriesLimit) {
```
DudaGod, line 132: Did we have 3 retries here before?

src/browser/screen-shooter/full-page-screen-shooter.ts
```diff
@@ -0,0 +1,267 @@
...
+                const cleanupMessage = cleanupError instanceof Error ? cleanupError.message : String(cleanupError);
+                console.warn(
+                    `Warning: failed to cleanup after full page screenshot.\nCleanup error: ${cleanupMessage}`,
+                );
+            }
+        }
+    }
+
+    private async _captureImpl(opts: FullPageCaptureOpts): Promise<Image> {
+        const COMPOSITING_ITERATIONS_LIMIT = 50;
```
DudaGod, line 84: it looks like we declare the same variable in elements-screen-shooter. Maybe we need to put it in a common place?

And why 50?

DudaGod (approved):
Amazing contribution 🔥 
Especially great that there are clear logs with the execution time everywhere.

src/browser/screen-shooter/full-page-screen-shooter.ts
```diff
@@ -0,0 +1,267 @@
...
+            const hasElementPositionsChanged = !this._isElementPositionsProbeEqual(
+                lastElementPositionsProbe,
+                scrollResult.elementPositionsProbe,
+            );
+            if (!hasElementPositionsChanged) {
+                debug(
+                    "Element positions probe did not change after scroll; falling back to initial viewport screenshot",
+                );
+                return viewportImage;
+            }
```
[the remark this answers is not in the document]
shadowusr: yeah that's exactly the goal lol

src/browser/screen-shooter/elements-screen-shooter.ts
```diff
@@ -0,0 +1,876 @@
...
+            retriesCount++;
+            startedAt = performance.now();
+            try {
+                perfDebug(`Starting capture attempt.`);
+
+                await new Promise(resolve => setTimeout(resolve, 1000));
```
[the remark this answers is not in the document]
shadowusr: yep, my bad, removed in subsequent commit!

src/browser/commands/assert-view/index.js
shadowusr: This block is only needed for debug, so no point in logging anything. The comment is present to satisfy eslint

src/browser/commands/assert-view/index.js
shadowusr: well since this block is only needed for debug i wanted to be 100% sure it doesn't throw.

It can throw. If executionContext is null. If something's wrong with test's fullTitle, etc.

src/browser/screen-shooter/operations/pointer-events.ts
shadowusr: The cursor may already be at 0x0, in which case this would be a noop and the hover state wouldn't be cleared. Moving it twice guarantees the desired effect — clearing hover after suppressing pointer events on page.

src/browser/screen-shooter/elements-screen-shooter.ts
shadowusr: We didn't — that's a new logic. 3 was chosen as an arbitrary number. For now, looks like a decent balance between speed and stability.

src/browser/screen-shooter/full-page-screen-shooter.ts
shadowusr: Moved it to constants.

I think it's a good idea to set a hard limit on any potentially infinite process. 50 is just an arbitrary number, IMO a reasonable one — doubt anyone will want to have screenshots spanning more than 50 screens =)

shadowusr:
Fixed remaining issues in #1242, to apply them on top of all other changes and reliably test everything at once (though i agree this approach is not ideal)

PR #1241, opened 2026-04-14
test: implement screen-shooter integration tests

shadowusr:
### What's done?
- Implemented integrations tests for element screen shooter, based on Testplane Standalone API


test/integration/screen-shooter/screen-shooter.test.ts
```diff
@@ -0,0 +1,277 @@
...
+
+        _.set(browserConfig.desiredCapabilities, "goog:chromeOptions", {
+            args: [
+                "--force-device-scale-factor=3",
+                "--high-dpi-support=1",
+                "--screen-info={devicePixelRatio=3}",
+                "--enable-gpu-rasterization",
+                "--enable-oop-rasterization",
+                "--ignore-gpu-blocklist",
+                "--enable-features=FractionalScrollOffsets",
```
DudaGod, line 228: why use such arguments?

DudaGod (approved):
🔥

test/integration/screen-shooter/screen-shooter.test.ts
shadowusr: It was a long process of trial and error to force browser in headless mode to return fractional scroll offsets without rounding. In practice, fractional offsets caused major issues on reasl websites & required a dedicated fix, so I wanted to have a test for that case. Without those args, however, chrome just rounded `window.scrollTop` from 0.5 to 1. These args allow to keep it as-is.

PR #1243, opened 2026-04-16
feat: add ability to use wsdriver

KuznetsovRoman:
Tests are failing because of typing issues
Those will be resolved after "wdio-types" and "webdriver" packages are updated to include this commit:
https://github.com/gemini-testing/webdriverio/pull/40


src/browser/cdp/connection.ts
KuznetsovRoman, line 1: Old "websocket connection management + connection retries" logic moved to "src/ws-connection/index.ts"

Now this class only has code to make "cdp-request" and handle "cdp-response"

src/ws-connection/index.ts
```diff
@@ -0,0 +1,547 @@
...
+                resolve(response);
+            }
+
+            pendingRequests[requestId] = done;
+
+            ws.send(requestMessage, error => {
+                if (!error) {
+                    this._pingShouldSkip = true;
+                    return;
+                }
```
KuznetsovRoman, line 363: Successful "send" skips next ping
There is no point to send "ping" requests if requests are being sent

src/ws-connection/index.ts
```diff
@@ -0,0 +1,547 @@
...
+    }
+
+    protected async _establishConnection(): Promise<void> {
+        await this._getWsConnection();
+    }
+
+    protected _forceReconnect(sessionAbortMessage: string): void {
+        this._closeWsConnection(sessionAbortMessage, WsConnectionStatus.DISCONNECTED);
+        this._tryToReconnect();
+    }
```
KuznetsovRoman: "_establishConnection" is used in "wsdriver-request", because we need establised connection to receive response headers and see, which compression methods are available

"_forceReconnect" is used in "wsdriver-request" when invalid message was received, which is sign of broken connection.

Other than those 2 methods + "skip ping", rest of the file is just refactored "cdp-connection", cleaned of all cdp-related stuff

src/browser/cdp/utils.ts
```diff
@@ -1,14 +1,3 @@
-export const exponentiallyWait = ({
```
KuznetsovRoman, line 1: Moved to "ws-connection/utils"

src/browser/wsdriver/connection.ts
```diff
@@ -0,0 +1,335 @@
...
+    sessionId: string;
+    headers?: Record<string, string>;
+    requestTimeout: number;
+    clientSupportedCompressionTypes: Array<(typeof WSD_COMPRESSION_TYPE)[keyof typeof WSD_COMPRESSION_TYPE]>;
+    supportedVersions: Record<number, boolean>;
+}
+
+export class WSDriverConnection extends WsConnection<IncomingWsDriverMessage, RawData> {
+    private _clientSupportedCompressionTypes: Array<(typeof WSD_COMPRESSION_TYPE)[keyof typeof WSD_COMPRESSION_TYPE]>;
+    private _serverSupportedCompressionType?: WsDriverCompressionType;
```
KuznetsovRoman: We only store one compression type, even if server sends multiple available because the best one is picked (out of client-supported methods)

src/browser/cdp/error.ts
```diff
@@ -1,51 +1,33 @@
...
+}
 
-        if (code) {
-            errorMessage += `\n\tErrorCode: ${code}`;
-        }
+export class CDPConnectionEstablishmentError extends WsConnectionEstablishmentError {}
+export class CDPConnectionBreakError extends WsConnectionBreakError {}
+export class CDPConnectionTerminatedError extends WsConnectionTerminatedError {}
+export class CDPConnectionTimeoutError extends WsConnectionTimeoutError {}
+export class CDPRequestTimeoutError extends WsRequestTimeoutError {}
```
KuznetsovRoman, line 20: WsConnection has its own Errors, which are part of business logic

In "CDPConnection" and "WsDriverConnection" we extend those errors and pass the classes to "WsConnection", so it can throw custom errors and rely on its "isRetryable" logic, which is different on CDP and WsDriver

src/browser/wsdriver/constants.ts
```diff
@@ -0,0 +1,14 @@
+export const WSD_CONNECTION_TIMEOUT = 15000; // 15 sec
+export const WSD_CONNECTION_RETRIES = 3;
+export const WSD_CONNECTION_RETRY_BASE_DELAY = 500;
+
+export const WSD_REQUEST_RETRIES = 3;
+export const WSD_REQUEST_RETRY_BASE_DELAY = 500;
+
+export const WSD_COMPRESSION_THRESHOLD_BYTES = 1024;
```
KuznetsovRoman, line 8: 1024 is a bit less than MTU + headers
If message weights less than MTU, there is no point in compression

src/browser/wsdriver/request.ts
KuznetsovRoman, line 1: Creating binary "request message" from got-like arguments (url + options)

src/browser/wsdriver/response.ts
KuznetsovRoman, line 1: Creating got-like response from  binary "response message"

src/browser/browser.ts
```diff
@@ -138,13 +138,15 @@ export class Browser {
                         req = this._config[optName](req);
                     }
 
-                    if (!req.headers!["X-Request-ID"]) {
-                        req.headers!["X-Request-ID"] = `${
+                    const requestHeaders = req.headers as Record<string, string>;
```
KuznetsovRoman, line 141: New version of `"@types/node"` complained about it
New version was necessary because old one has weird types when it comes to `RawData` -> `Buffer` conversion

test/src/browser/cdp/connection.ts
KuznetsovRoman, line 1: General "connection management" logic is moved to "ws-connection" tests

test/src/ws-connection/index.ts
```diff
@@ -0,0 +1,399 @@
...
+                    requestId,
+                }),
+            );
+        }
+    }
+
+    async request<T = void>(method: string, params?: Record<string, unknown>): Promise<T> {
+        let result!: T | WsError;
+
+        for (let retriesLeft = 3; retriesLeft >= 0; retriesLeft--) {
```
KuznetsovRoman, line 149: retries logic here is used in order to test request retries on connection breaks

src/ws-connection/index.ts
```diff
@@ -0,0 +1,547 @@
...
+    private _pingInterval: ReturnType<typeof setInterval> | null = null;
+    private _pingSubsequentFails = 0;
+    private _onConnectionCloseFn: (() => void) | null = null; // Defined, if there is connection attempt at the moment
+    private _wsConnectionStatus: WsConnectionStatus = WsConnectionStatus.DISCONNECTED;
+    private _wsConnection: WebSocket | null = null;
+    private _wsConnectionPromise: Promise<WebSocket> | null = null;
+    private _requestId = 0;
+    private _pendingRequests: Record<number, (response: ResponseMessageType | WsError) => void> = {};
+
+    protected constructor(cdpWsEndpoint: string, { headers, debugFn, retries, timeouts, errors }: WsConnectionOptions) {
```
shadowusr: if this is a an abstract base class which is a parent of both cdp connection and wsd connections, why do we have cdp mentions here?

src/ws-connection/index.ts
```diff
@@ -0,0 +1,547 @@
...
+        this._debugFn(`⟳ Trying to reconnect; endpoint: "${this._cdpWsEndpoint}"`);
+
+        this._getWsConnection()
+            .then(() => this._debugFn(`\u2713 Successfully reconnected to session; endpoint: "${this._cdpWsEndpoint}"`))
+            .catch(() =>
+                this._debugFn(`\u2718 Couldn't reconnect to session automatically; endpoint: "${this._cdpWsEndpoint}"`),
+            );
+    }
+
+    protected async _establishConnection(): Promise<void> {
```
shadowusr: imho this method feels awkward. Looking at the usage site:

```ts
await this._establishConnection();

if (!this._responseHeaders || !this._responseHeaders[WSD_ACCEPT_ENCODING_HEADER]) {
    return (this._serverSupportedCompressionType = WsDriverCompression.None);
}
```

So the actual goal of calling this method is to get response headers and retrieve supported compression type. However, this is quite unexpected and not explicit, it's very easy to change internal logic of _getWsConnection and forget that it's expected that it sets those fields.

If we want to get connection properties like compression type, we could use explicit method that returns them e.g. `const connectionProperties = await this._getConnectionProperties()`

src/ws-connection/index.ts
```diff
@@ -0,0 +1,547 @@
...
+    if (ws.readyState !== ws.CONNECTING) {
+        ws.close();
+    } else {
+        ws.once("open", () => {
+            ws.close();
+        });
+    }
+};
+
+export abstract class WsConnection<
```
shadowusr: IMO the "composition over inheritance" rule is very relevant here.

We extend WsConnection in cdp connection class and wsdriver connection class just to reuse parts of its logic. In essence, this class implements low-level transport. cdp conneciton/wsdriver connection are high-level request-makers.

I can hardly see the "is-a" relationship here

Besides, we try to make cdp connection and wsdriver connection similar as if they were drop-in replacements for each other (implementing common interface), but they aren't! They are completely different classes with different `request` signatures

I see the following issues with inheritance approach:
- No clear separation between public/private API in WsConnection — you could argue that `protected` methods are "public api" for child classes and `private` are private, but isn't that awkward?
- High-level cdp/wsd classes being aware of transport details
- Wrong semantics and zero value from inheritance

**TLDR**
IMO the code would be much cleaner if this class was hiding all low-level transport stuff from its consumers and cdp/wsd classes just accepted instance of WsConnection in constructor.

src/browser/wsdriver/connection.ts
```diff
@@ -0,0 +1,335 @@
...
+
+interface WSDriverConnectionOptions {
+    sessionId: string;
+    headers?: Record<string, string>;
+    requestTimeout: number;
+    clientSupportedCompressionTypes: Array<(typeof WSD_COMPRESSION_TYPE)[keyof typeof WSD_COMPRESSION_TYPE]>;
+    supportedVersions: Record<number, boolean>;
+}
+
+export class WSDriverConnection extends WsConnection<IncomingWsDriverMessage, RawData> {
```
shadowusr: In https://github.com/gemini-testing/webdriverio/pull/40, you say

> "customWdRequestAgent" should have the same API as _libRequest

But there are no checks on type level. I'd expect something like
```typescript
class WSDriverConnection implements RequestAgent {}
```

Where `RequestAgent` is an interface imported from our wdio fork and being implemented by both our class and built-in request agent

shadowusr: I think it would be more correct to call this class `WSDriverRequestAgent`.

We have:
- WsConnection
- WSDriverConnection
- CDPConnection

What do they have in common? In my opinion, nothing. `WsConnection` is a good pick for low-level transport. This class is request agent and used as request agent. `CDPConnection` is closer to request agent as well, but i am not sure, it's somewhere in between

src/browser/wsdriver/connection.ts
```diff
@@ -0,0 +1,335 @@
...
+        headers = {},
+        browserConfig,
+    }: {
+        sessionId: string;
+        sessionCaps: WebdriverIO.Capabilities;
+        headers: Record<string, string>;
+        browserConfig: BrowserConfig;
+    }): WSDriverConnection {
+        if (!sessionCaps["se:wsdriver"]) {
+            throw new WSDriverError({ message: "Couldn't determine wsdriver endpoint" });
```
shadowusr: I'd make this error much more detailed. What did we try, what user can do to fix, etc.

Same for below.

src/browser/wsdriver/request.ts
```diff
@@ -0,0 +1,99 @@
...
+        shouldCompress ? connectionOptions.compressionType : WsDriverCompression.None,
+        true,
+    );
+    const requestMethod = getRequestMethod(requestOptions.method);
+
+    const resultMessage = Buffer.alloc(
+        OUTGOING_MESSAGE_AUXILIARY_BYTES + Buffer.byteLength(command) + Buffer.byteLength(compressedPayload),
+    );
+
+    let ptr = 0;
```
shadowusr, line 84: Pretty sure we have a design doc on the message structure, can we add a comment referencing it here?

src/browser/existing-browser.ts
```diff
@@ -274,6 +275,19 @@ export class ExistingBrowser extends Browser {
             requestedCapabilities: sessionOpts.capabilities,
         };
 
+        if (this._config.useWsDriver && sessionCaps && sessionCaps["se:wsdriver"]) {
+            const supportsWsDriverV1 = sessionCaps["se:wsdriverVersion"]?.split(", ").includes("1");
```
shadowusr, line 281: Hmmm what is `se:wsdriverVersion` expected to contain if we split it by `, `? I thought it was just a version number.

shadowusr (approved):
The additions are great! 🔥 

Comments below are not blockers, but I do strongly believe those changes would make the architecture much cleaner and simpler.

src/browser/existing-browser.ts
KuznetsovRoman: If server supports multiple versions, he can respond with `"1, 2, 3"`
It is expected that client can send messages of different versions on one connection

PR #1246, opened 2026-04-23
feat(selectivity): add ability to filter source map urls

KuznetsovRoman:


src/browser/cdp/selectivity/css-selectivity.ts
```diff
@@ -50,7 +58,20 @@ export class CSSSelectivity {
...
 
-        const sourceMapResolvedUrl = urlResolve(sourceURL, sourceMapURL);
+        let sourceMapResolvedUrl = urlResolve(sourceURL, sourceMapURL);
+
+        const mapResult = this._mapSourceMapUrl
+            ? this._mapSourceMapUrl({ type: "css", sourceUrl: sourceURL, sourceMapUrl: sourceMapResolvedUrl })
+            : true;
+
+        if (!mapResult) {
+            this._stylesSourceMap[styleSheetId] ||= null;
```
KuznetsovRoman, line 68: marks stylesheet as "handled" so we would not pull its source manually, treating it as "lost event"

src/browser/cdp/selectivity/js-selectivity.ts
```diff
@@ -109,74 +131,107 @@ export class JSSelectivity {
     private _ensureScriptsAreLoading(coverage: CDPScriptCoverage[]): void {
         coverage.forEach(({ scriptId, url }) => {
             const fixedUrl = url || this._scriptIdToSourceUrl[scriptId];
+            const wasProcessed =
+                Object.hasOwn(this._scriptsSource, scriptId) && Object.hasOwn(this._scriptsSourceMap, scriptId);
+            const isAnonymous = !fixedUrl;
+            const urlWasCorrected = url && url !== this._scriptIdToSourceUrl[scriptId];
```
KuznetsovRoman, line 137: Case "previously we got event with empty url, but now we got it"
For some reason, cdp might send multiple "scriptParsed" events for same script + "takeCoverage" also returns script's "url", which can be different

src/browser/cdp/selectivity/js-selectivity.ts
```diff
@@ -109,74 +131,107 @@ export class JSSelectivity {
...
 
             // Not dropping sources to fs the end of test (when "stop" is called) because we use it immediately
-            const scriptSourcePromise = this._cdp.debugger
-                .getScriptSource(this._sessionId, scriptId)
-                .then(({ scriptSource }) => {
-                    setCachedSelectivityFile(CacheType.Asset, fixedUrl, scriptSource).catch(() => {});
-                    return scriptSource;
-                })
-                .catch((err: Error) => err);
+            const scriptSourcePromise = (async (): Promise<string | Error> => {
```
KuznetsovRoman, line 145: Now we can reach this part if we got valid new "url" and we want to recalculate "sourceMapUrl"

If "sourceMapUrl" was lost, it can be regenerated via "sourceUrl" + "sourceMapUrl" comment in source file

In this code we check "if script was loaded previously" and retrieving it, if yes, with fallback to "getScriptSource", if it can't be reached locally

src/browser/cdp/selectivity/js-selectivity.ts
```diff
@@ -109,74 +131,107 @@ export class JSSelectivity {
...
+                                  sourceMapsEndIndex,
+                              );
 
-                const resolvedSourceMapUrl = urlResolve(fixedUrl, sourceMapURL);
+                    let resolvedSourceMapUrl = urlResolve(fixedUrl, sourceMapURL);
 
-                this._scriptIdToSourceMapUrl[scriptId] ||= resolvedSourceMapUrl;
+                    const mappedResult = this._mapSourceMapUrl
+                        ? this._mapSourceMapUrl({ type: "js", sourceUrl: fixedUrl, sourceMapUrl: resolvedSourceMapUrl })
+                        : true;
```
KuznetsovRoman, line 193: As we got sourceMapUrl through this fallback, it should be mapped too.

src/browser/cdp/selectivity/utils.ts
```diff
@@ -220,7 +220,7 @@ const warnUnsupportedProtocol = memoize((protocol: string, dependency: string):
 export const transformSourceDependencies = (
     cssDependencies: Set<string> | null,
     jsDependencies: Set<string> | null,
-    mapDependencyPathFn?: null | ((relativePath: string) => string | void),
+    mapDependencyPathFn?: null | ((relativePath: string) => string | boolean | void),
```
KuznetsovRoman, line 227: tolerating boolean values allows users to handle it as "filter" function, which is more convenient to some users (when it comes to "sourceMapUrl", but we can get similar API on them)

sonic16x (approved):

PR #1250, opened 2026-05-13
chore: check next version of wdio

DudaGod:


DudaGod:
merge it accidentally, I'm going to revert it

PR #1259, opened 2026-05-21
fix(wsdriver): provide connection error info

KuznetsovRoman:


src/browser/wsdriver/request.ts
```diff
@@ -63,7 +63,7 @@ export const constructWsDriverRequest = async (
         throw new Error("Not session-related request");
     }
 
-    const command = url.pathname.slice(commandStartIdx + connectionOptions.sessionPrefix.length);
+    const command = decodeURIComponent(url.pathname.slice(commandStartIdx + connectionOptions.sessionPrefix.length));
```
KuznetsovRoman, line 66: It can fail without decoding for some special appium commands

src/ws-connection/error.ts
```diff
@@ -26,13 +26,17 @@ export abstract class WsError extends Error {
...
+    constructor({ message, requestId, statusCode }: { message: string; requestId?: number; statusCode?: number }) {
+        super({ message, code: statusCode || WS_ERROR_CODE.CONNECTION_ESTABLISHMENT, requestId });
 
         this.name = this.constructor.name;
     }
 
     isRetryable(): boolean {
+        if (this.code && this.code >= 400 && this.code < 500) {
+            return false;
+        }
```
KuznetsovRoman, line 38: If we received 4xx on upgrade request, we dont need to retry it

sonic16x (approved):

PR #1260, opened 2026-05-21
fix: read ambigous TS files with type-stripping enabled with require first

shadowusr:
### What's done?

Before, we had the following issue when node's built in typescript types stripping (https://nodejs.org/download/release/v24.1.0/docs/api/typescript.html#type-stripping) is enabled:

1. Mocha tries to read tests with `readFilesAsync`, which `import()`s them first.
2. Node strips types and tries to read file as cjs (since root package json is not type: module)
3. Node fails and sees `import`s in typescript file and tries to re-read the file as ESM, issuing the warning `MODULE_TYPELESS_PACKAGE_JSON] Warning: Module type of some-file.hermione.ts?rand=0.4001007512274388 is not specified and it doesn't parse as CommonJS. Reparsing as ES module because module syntax was detected `
4. This fails too, finally mocha reads it with require, our transform hooks apply and file reads fine.

This means 3 file reads and 1 warning in console.

This PR adds the following logic:
- If node's type stripping is active and file is typescript with ambigous extension (not cts/mts), we first try to read it with `loadFile` (which uses `require`) and secondly if it fails, `loadFileAsync`.

Also added 4 test projects that test file reading under different circumstances e2e.


.github/workflows/module-loading.yml
```diff
@@ -0,0 +1,29 @@
...
+        runs-on: ubuntu-latest
+
+        steps:
+            - uses: actions/checkout@v4
+
+            - name: Use Node.js 24.x
+              uses: actions/setup-node@v4
+              with:
+                  node-version: 24.x
+                  cache: "npm"
```
KuznetsovRoman, line 20: I think we should run another node.js versions also: like "20" and "22"

If you checked it once and it works on 18/20/22 - its fine, we can keep "only node 24" check

src/test-reader/mocha-reader/index.js
```diff
@@ -77,6 +81,66 @@ async function readFiles(files, { esmDecorator, config, eventBus, runnableOpts,
...
 }
 
+async function loadMochaFiles(mocha, files, { esmDecorator }) {
+    if (["true", "1"].includes(process.env.TESTPLANE_LOAD_FILES_ASYNC) || !files.some(shouldLoadWithRequire)) {
+        await mocha.loadFilesAsync({ esmDecorator });
+        return;
+    }
+
+    const originalFiles = mocha.files;
+    mocha.lazyLoadFiles(true);
```
KuznetsovRoman, line 91: With this option, it would be ok that we load files in "for" loop instead of loading them asynchronously with some kind of "Promise.all"?

src/test-reader/mocha-reader/index.js
```diff
@@ -77,6 +81,66 @@ async function readFiles(files, { esmDecorator, config, eventBus, runnableOpts,
...
+
+    try {
+        for (const file of files) {
+            mocha.files = [file];
+
+            if (!shouldLoadWithRequire(file)) {
+                await mocha.loadFilesAsync({ esmDecorator });
+                continue;
+            }
```
KuznetsovRoman, line 100: Why isn't it written like that?
```js
const loadWithRequireFiles = [];
const loadDefaultFiles = [];

for (const file of files) {
    if (shouldLoadWithRequire(file)) {
        loadWithRequireFiles.push(file)
    } else {
        loadDefaultFiles.push(file)
    }
}

// Load "loadDefaultFiles" with some kind of Promise.all
// Load "loadWithRequireFiles" with "mocha.loadFiles"
```

KuznetsovRoman (approved):
🔥

.github/workflows/module-loading.yml
shadowusr: ESM wouldn't work on node < v20.17.0, and only 24 has typescript types stripping turned on by default. I'll verify once again locally on different node versions, but going to ultimately keep the check turned on for 24 only

src/test-reader/mocha-reader/index.js
shadowusr: yeah it's fine. the lazyLoadFiles only sets mocha's internal flag. in fact, loadFilesAsync sets this exact flag under the hood. here, we set it, because we use sync version — loadFiles, to prevent files from loading again when `run()` is called.

src/test-reader/mocha-reader/index.js
shadowusr: We wouldn't gain anything from doing that: mocha itself loads files sequentially, here's the source of loadFilesAsync:

```js
exports.loadFilesAsync = async (
  files,
  preLoadFunc,
  postLoadFunc,
  esmDecorator
) => {
  for (const file of files) {
    preLoadFunc(file);
    const result = await exports.requireOrImport(
      path.resolve(file),
      esmDecorator
    );
    postLoadFunc(file, result);
  }
};

```

PR #1261, opened 2026-05-25
fix!: remove global helpers (it,describe,...)

KuznetsovRoman:


src/index.ts
```diff
@@ -46,24 +46,3 @@ export type { SaveStateData } from "./browser/commands/saveState";
 
 import type { TestDefinition, SuiteDefinition, TestHookDefinition } from "./test-reader/test-object/types";
 export type { TestDefinition, SuiteDefinition, TestHookDefinition };
-
-declare global {
```
KuznetsovRoman, line 50: Now we dont pollute global by default

src/types/globals.ts
```diff
@@ -0,0 +1,18 @@
...
+import type { SuiteDefinition, TestDefinition, TestHookDefinition } from "../test-reader/test-object/types";
+
+export type GlobalItType = TestDefinition;
+export type GlobalDescribeType = SuiteDefinition;
+export type GlobalBeforeEachType = TestHookDefinition;
+export type GlobalAfterEachType = TestHookDefinition;
+export type GlobalTestplaneType = GlobalHelper;
+export type GlobalHermioneType = GlobalHelper;
+
+export type TestplaneGlobals = {
```
KuznetsovRoman, line 11: Instead, we have type of testplane-related globals to use

src/globals.ts
```diff
@@ -0,0 +1,16 @@
...
+    GlobalTestplaneType,
+    TestplaneGlobals,
+} from "./types/globals";
+
+export const it: GlobalItType = (globalThis as unknown as TestplaneGlobals).it;
+export const describe: GlobalDescribeType = (globalThis as unknown as TestplaneGlobals).describe;
+export const beforeEach: GlobalBeforeEachType = (globalThis as unknown as TestplaneGlobals).beforeEach;
+export const afterEach: GlobalAfterEachType = (globalThis as unknown as TestplaneGlobals).afterEach;
+export const testplane: GlobalTestplaneType = (globalThis as unknown as TestplaneGlobals).testplane;
+export const hermione: GlobalHermioneType = (globalThis as unknown as TestplaneGlobals).hermione;
```
KuznetsovRoman, line 16: Types of globals are exported instead of being declared

So user could type:
```ts
import {it, describe} from "testplane";
```

Also same exported types will be used in "@testplane/globals"

sipayRT (approved):

PR #1273, opened 2026-06-04
feat!: add async function config support

KuznetsovRoman:


src/cli/index.ts
```diff
@@ -58,7 +58,7 @@ export const run = async (opts: TestplaneRunOpts = {}): Promise<void> => {
     }
 
     const configPath = preparseOption(program, "config") as string;
-    testplane = Testplane.create(configPath);
+    testplane = await Testplane.create(configPath);
```
KuznetsovRoman, line 61: `Testplane.create` is now async, because config is parsed inside of it

test/src/cli/commands/install-deps/index.ts
```diff
@@ -9,14 +9,14 @@ import type { Config } from "../../../../../src/config";
...
-    let cli: { run: () => void };
+    let cli: { run: () => Promise<void> };
     let loggerStub: { log: SinonStub; warn: SinonStub; error: SinonStub };
     let testplaneStub: Writable<Testplane>;
     let installBrowsersWithDriversStub: SinonStub;
 
     const installBrowsers_ = async (argv: string = ""): Promise<void> => {
         process.argv = ["foo/bar/node", "foo/bar/script", "install-deps", ...argv.split(" ")].filter(Boolean);
-        cli.run();
+        await cli.run();
```
KuznetsovRoman, line 19: `cli.run` was always async

sonic16x (approved):

PR #1275, opened 2026-06-08
feat: add bidi client

KuznetsovRoman:


src/browser/bidi/connection.ts
KuznetsovRoman, line 1: First file with some code (which is 80% similar to CDP one)

src/browser/bidi/index.ts
KuznetsovRoman, line 1: Second file with some code, which is 99% similar to CDP one

PR #1276, opened 2026-06-10
fix(selectivity): anonymous styles

KuznetsovRoman:


src/browser/cdp/selectivity/css-selectivity.ts
```diff
@@ -73,6 +73,11 @@ export class CSSSelectivity {
 
         let sourceMapResolvedUrl = urlResolve(sourceURL, sourceMapURL);
 
+        if (!URL.canParse(sourceMapResolvedUrl)) {
```
KuznetsovRoman, line 76: Instead of checking "is there any base url" we see "if it resolved to valid url" (no source url + embedded sourcemap = valid)

src/browser/cdp/selectivity/js-selectivity.ts
```diff
@@ -46,19 +46,22 @@ export class JSSelectivity {
...
+        // No source map or error
+        if (!sourceMapURL || url.startsWith("chrome-error://")) {
             this._scriptsSource[scriptId] ||= null;
             this._scriptsSourceMap[scriptId] ||= null;
             return;
         }
 
-        if (this._scriptsSource[scriptId] && this._scriptsSourceMap[scriptId]) {
+        // Anonymous or processed
+        if (!url || (this._scriptsSource[scriptId] && this._scriptsSourceMap[scriptId])) {
```
KuznetsovRoman: If script is anonymous, dont set "null" to scripts map so it would not be considered as "handled" and would be processed if needed, on "stop" call if this script explicitly was in the coverage

package.json
```diff
@@ -50,7 +50,7 @@
   },
   "homepage": "https://testplane.io/",
   "engines": {
-    "node": ">= 18.0.0"
+    "node": ">= 18.17.0"
```
KuznetsovRoman, line 53: URL.canParse

sonic16x (approved):

PR #1277, opened 2026-06-15
fix: fix remaining issues in new assertView algorithm and implement cropMargins options

shadowusr:
### What's done?

- Implemented the `cropMargins` options that lets users specify crop areas at the camera level. Intended to be used instead of old hacks like generating cover elements to circumvent scrollbars. Now, you can just specify `cropMargins: { right: 8 }` to always cut 8 pixels off each screenshot

- Fixed lots of edge cases, each edge case is described at the corresponing commit message. Note: at first, i decided to auto-include fixed-positioned descendants in capture specs, but later reverted this, so don't spend time on this if you are going to review this PR commit-by-commit


src/browser/camera/index.ts
```diff
@@ -20,12 +21,15 @@ import { NEW_ISSUE_LINK } from "../../constants/help";
 const debug = makeDebug("testplane:screenshots:camera");
 
 export type ScreenshotMode = "fullpage" | "viewport" | "auto";
+export type { CropMargins } from "./utils";
```
KuznetsovRoman, line 24: Duplicate import:
L7: `import type { CropMargins } from "./utils";`
L24: `import type { CropMargins } from "./utils";`

src/browser/camera/utils.ts
```diff
@@ -24,6 +33,40 @@ export const isFullPage = (
...
+            );
+        }
+    }
+
+    return result;
+};
+
+export const cropMarginsToRect = (
+    imageArea: Rect<"image", "device">,
+    cropMargins: CropMargins | undefined,
```
KuznetsovRoman: Lets replace all occurences of `Type | undefined` into `?: Type`:
`cropMargins: CropMargins | undefined` -> `cropMargins?: CropMargins`

src/browser/camera/index.ts
```diff
@@ -82,15 +86,23 @@ export class Camera {
...
         const calibrationArea = shouldApplyCalibration ? this._calibratedArea : null;
 
-        const calibratedImageArea = this._cropAreaToCalibratedArea(imageArea, calibrationArea);
+        const calibratedImageArea = this._cropAreaToIntersection(imageArea, calibrationArea);
+        const cropMarginsArea = utils.cropMarginsToRect(imageArea, opts?.cropMargins);
+        const croppedImageArea = getIntersection(calibratedImageArea, cropMarginsArea);
+        if (croppedImageArea === null) {
+            throw new Error(
+                `Invalid cropMargins option: resulting screenshot crop area is empty. ` +
+                    `imageSize: ${prettySize(imageArea)}, cropMargins: ${JSON.stringify(opts?.cropMargins ?? {})}`,
```
KuznetsovRoman: Why do we need `?? {}` here?
IMO, it would be better to remove it, since "null" and "undefined" can be easily json stringified
And we would have more accurate error, because "null" and "undefined" would not be clamped to `{}`.

src/browser/camera/utils.ts
```diff
@@ -24,6 +33,40 @@ export const isFullPage = (
...
+): Rect<"image", "device"> => {
+    const margins = normalizeCropMargins(cropMargins);
+
+    return {
+        top: margins.top as Coord<"image", "device", "y">,
+        left: margins.left as Coord<"image", "device", "x">,
+        width: ((imageArea.width as number) - margins.left - margins.right) as typeof imageArea.width,
+        height: ((imageArea.height as number) - margins.top - margins.bottom) as typeof imageArea.height,
+    };
```
KuznetsovRoman: nit: i think we can get rid of 4 "as ..." at a cost of 1 `as Rect<"image", "device">` at L74

src/browser/client-scripts/screen-shooter/operations.ts
```diff
@@ -263,15 +301,18 @@ export function computeSafeArea(
         if (getIntersection(fromBcrToRect(bcr), viewportRect) === null) continue;
 
         let likelyInterferes = false;
+        let skipZIndexCheck = false;
```
KuznetsovRoman: nit: lets name it more boolean-like: **should**SkipZIndexCheck

src/browser/client-scripts/screen-shooter/operations.ts
```diff
@@ -360,7 +403,7 @@ export function computeSafeArea(
             logger?.("decided to shrink bottom");
         }
 
-        if (resultingHeight < origHeight / 2) {
+        if (resultingHeight < origHeight * 0.3) {
```
KuznetsovRoman, line 407: Magic number
Why "/2" was not enough and "*0.3" is?
Lets leave a comment about it

src/browser/screen-shooter/composite-image/index.ts
```diff
@@ -424,6 +437,59 @@ export class CompositeImage {
         }
     }
 
+    private _expandCandidatesToFullArea(candidates: SegmentCandidate[]): void {
```
KuznetsovRoman, line 462: some code comments "what this method does" would be nice

src/browser/screen-shooter/elements-screen-shooter.ts
```diff
@@ -33,6 +34,7 @@ class CaptureAreaSizeChangeError extends Error {
 }
 
 const debug = makeDebug("testplane:screenshots:elements-screen-shooter");
+const SCROLL_OVERLAP = 1;
```
KuznetsovRoman: nit: lets add "_PX" to it, so it would look like a number instead of boolean of some sort

src/browser/screen-shooter/full-page-screen-shooter.ts
```diff
@@ -67,6 +68,13 @@ export class FullPageScreenShooter {
     }
 
     async capture(opts: FullPageCaptureOpts = {}): Promise<Image> {
+        if (!opts.disableHover) {
+            opts.disableHover = DisableHoverMode.WhenScrollingNeeded;
+        }
+        if (opts.disableAnimation === undefined) {
```
KuznetsovRoman: I suggest `if (opts.disableAnimation !== false)`
Or `opts.disableAnimation ??= true`

src/browser/screen-shooter/full-page-screen-shooter.ts
```diff
@@ -199,8 +209,8 @@ export class FullPageScreenShooter {
     }
 
     private _isElementPositionsProbeEqual(
-        leftProbe: ElementPositionsProbe<"device">,
-        rightProbe: ElementPositionsProbe<"device">,
+        leftProbe: ElementPositionsProbe<"device">[],
+        rightProbe: ElementPositionsProbe<"device">[],
```
KuznetsovRoman: Since those are arrays, lets name them with plural form:
leftProbe**s**, rightProbe**s**

test/src/browser/screen-shooter/composite-image/fixtures/generate.ts
```diff
@@ -915,6 +929,291 @@ if (process.argv.includes("generate")) {
...
+                height: 20 as Length<"device", "y">,
+            };
+            const offscreenBox = {
+                left: toPageX(100),
+                top: toPageY(340),
+                width: 80 as Length<"device", "x">,
+                height: 20 as Length<"device", "y">,
+            };
+
+            const page = Buffer.alloc((pageWidth as number) * (pageHeight as number) * 4);
```
KuznetsovRoman: Magic number "4".
Perhaps we should use "RGBA_CHANNELS" const

KuznetsovRoman (approved):

src/browser/camera/index.ts
shadowusr: That's export, not import. If it was duplicate it wouldn't compile

src/browser/client-scripts/screen-shooter/operations.ts
shadowusr: There's a e2e test on this exactly. Stumbled upon a real world case where we needed to shrink safe area past 30%. Will leave a comment about that

src/browser/screen-shooter/composite-image/index.ts
shadowusr: added brief descriptions for each method.

PR #1281, opened 2026-06-21
fix: older browsers support in time travel and assertView

shadowusr:
### What's done?

All these changes tested on Chrome 53

- assertView now works.
- rrweb can't be run on chrome 53 (rrweb bundle has newer syntax), so this lack of support is handled gracefully
- fixed duplicate session close by unsubscribing from signalHandler events
- fixed incorrect rollback logic when safeArea size changes mid-capture after first chunk
- made selectors settle waiting compatible with older browsers (tested on chrome 67)

Not tested on internet explorer, because it turns out it was broken a long time ago: https://github.com/gemini-testing/testplane/blob/f540130eef618bde03546d61eafc6f7a8c813f51/src/browser/new-browser.ts#L223. In this chunk of code, we assign custom capability, which makes the IE throw on session creation. So I've kept this as-is.


src/browser/client-bridge/index.ts
```diff
@@ -28,13 +29,27 @@ export class ClientBridge<T extends Record<string, (...args: any[]) => any>> {
...
 
+        let debugBrowserId = "";
+        if (debug.enabled) {
+            debugBrowserId = `${(browser as WdioBrowser)?.capabilities?.browserName} ${
+                (browser as WdioBrowser)?.capabilities?.browserVersion
+            }`;
+        }
+
         if (bundlesCache[scriptFilePath]) {
+            debug(
```
KuznetsovRoman, line 40: Why would we need to know it?
This debug log does not seem to be usefull

src/browser/client-bridge/index.ts
```diff
@@ -89,10 +104,7 @@ export class ClientBridge<T extends Record<string, (...args: any[]) => any>> {
     }
 
     private async _inject(): Promise<void> {
-        debug(` > injecting script into namespace ${this._namespace}`);
-        if (debug.enabled) {
-            console.log(this._script);
-        }
+        debug(` > injecting script into namespace ${this._namespace}: ${this._script.slice(0, 256)}...`);
```
KuznetsovRoman, line 107: 👍

src/browser/client-bridge/index.ts
```diff
@@ -28,13 +29,27 @@ export class ClientBridge<T extends Record<string, (...args: any[]) => any>> {
         const scriptFileName = needsCompatLib ? "bundle.compat.js" : "bundle.native.js";
         const scriptFilePath = path.join(__dirname, "..", "client-scripts", namespace, "build", scriptFileName);
 
+        let debugBrowserId = "";
+        if (debug.enabled) {
+            debugBrowserId = `${(browser as WdioBrowser)?.capabilities?.browserName} ${
```
sipayRT, line 34: этого точно хватит для id? у нас же может быть 2 браузера одинаковой версии, но с разными капабилити (desktp/touch)

src/browser/history/rrweb.ts
```diff
@@ -23,19 +29,42 @@ export async function installRrwebAndCollectEvents(
...
         if (shouldSendRrwebCode) {
-            sessionsWithRrwebRequested.add(session);
+            sessionsWithRrwebSent.add(session);
         }
 
-        const result = await collectRrwebEvents(session, shouldSendRrwebCode ? rrwebCode : null);
+        let result = await collectRrwebEvents(session, shouldSendRrwebCode ? rrwebCode : undefined);
 
+        if (result.isRrwebSupported === false) {
+            debug("rrweb is not supported in this browser, error: %s", result.evalError);
```
KuznetsovRoman: > this browser

Maybe we should specify, what browser are we talking about? Like name and version

src/browser/history/rrweb.ts
```diff
@@ -86,21 +115,22 @@ export async function cleanupRrweb(session: WebdriverIO.Browser, callstack: Call
...
+    rrwebRecordFnCode: string | undefined,
 ): Promise<CollectRrwebEventsResult> {
     return session.execute(
         (rrwebRecordFnCode, serverTime) => {
             const isRrwebInstalled = (): boolean => {
                 try {
                     // @ts-expect-error
                     return Boolean(window.rrweb);
-                } catch {
+                } catch (e) {
```
sipayRT, line 138: ?

sipayRT (approved):

src/browser/history/rrweb.ts
```diff
@@ -57,8 +86,8 @@ export async function cleanupRrweb(session: WebdriverIO.Browser, callstack: Call
                     }
 
                     try {
-                        const colorSchemeMedia = rrwebData?.colorSchemeMedia;
-                        const colorSchemeListener = rrwebData?.colorSchemeListener;
+                        const colorSchemeMedia = rrwebData && rrwebData.colorSchemeMedia;
+                        const colorSchemeListener = rrwebData && rrwebData.colorSchemeListener;
```
KuznetsovRoman, line 95: Whats the difference?

src/browser/history/rrweb.ts
```diff
@@ -86,21 +115,22 @@ export async function cleanupRrweb(session: WebdriverIO.Browser, callstack: Call
...
-        sessionsWithRrwebRequested.delete(session);
+        sessionsWithRrwebSent.delete(session);
+        sessionsWithUnsupportedRrweb.delete(session);
     }
 }
 
 function collectRrwebEvents(
     session: WebdriverIO.Browser,
-    rrwebRecordFnCode: string | null,
+    rrwebRecordFnCode: string | undefined,
```
KuznetsovRoman, line 130: nit: -> `rrwebRecordFnCode?: string`

KuznetsovRoman (approved):

src/browser/history/rrweb.ts
shadowusr: The `catch {}` syntax became available from chrome 66, while we are targeting chrome 53: https://caniuse.com/wf-optional-catch-binding

src/browser/client-bridge/index.ts
shadowusr: Added sessionId

src/browser/client-bridge/index.ts
shadowusr: It actually is very useful, for example it highlights which version of the script — native or compat was chosen, which instantly makes it clear if calibration works as expected, etc.

src/browser/history/rrweb.ts
shadowusr: optional chaining is supported from chrome 80: https://caniuse.com/mdn-javascript_operators_optional_chaining

While we are targeting much earlier versions.

src/browser/history/rrweb.ts
shadowusr: Well it's actually important here.

The story is that in older versions, argument is expected to be string or undefined precisely (with null it throws an error). However, we don't mean that "argument can be omitted". We mean precisely that it must be passed, but can be either string or undefined.

PR #1285, opened 2026-06-29
Fix add timetravel timeout

sonic16x:
Before this change, if Time Travel snapshot collection hung, the test runner would wait forever with only a `Collecting Time Travel snapshots takes longer than expected. Waiting...` warning. Now collection is aborted after a timeout and the run continues, logging `Collecting Time Travel snapshots timed out after 5000ms` — the 
  snapshot is simply omitted from the report.


shadowusr (approved):
LGTM, but I would mark the session as broken in case snapshot collection time outs.

PR #1289, opened 2026-07-05
fix: add support for ignoreElements in viewport-mode assertView

shadowusr:
### What's done?

- Added support for `ignoreSelectors` in assertView viewport mode (e.g. `assertView('state-name', { ignoreElements: ['#ignore'] })`)


sonic16x (approved):

src/browser/client-scripts/screen-shooter/implementation.ts
```diff
@@ -368,13 +368,19 @@ export function scrollFullPage(
 }
 
 export function prepareViewportScreenshot(
-    opts: { usePixelRatio?: boolean; disableAnimation?: boolean; disableHover?: DisableHoverMode } = {}
+    opts: {
+        usePixelRatio?: boolean;
+        disableAnimation?: boolean;
+        disableHover?: DisableHoverMode;
+        ignoreSelectors?: string[];
```
DudaGod, line 375: Why did you decide to name it differently? I mean not `ignoreElements`. For example, for android, we can pass native locators

DudaGod (approved):

src/browser/client-scripts/screen-shooter/implementation.ts
shadowusr: It's not public API, it's an internal options object, so we can change this any time. It's consistent with this field:

https://github.com/gemini-testing/testplane/blob/81fa6873f8330dad26984f383adb95118dd42907/src/browser/client-scripts/screen-shooter/types.ts#L41

> For example, for android, we can pass native locators

Hmm, I doubt that will work currently. Right now we only work with css selectors in assertView. But yeah, in the future we could support that. In that case, we can always rename it

PR #1291, opened 2026-07-10
Fix increase command args length

sonic16x:
Need for https://github.com/gemini-testing/html-reporter/pull/780

### DIFF
--- a/src/browser/history/utils.ts
+++ b/src/browser/history/utils.ts
@@ -4,7 +4,7 @@ import _ from "lodash";
 import { TestStep, TestStepKey } from "../../types";
 import { TimeTravelMode } from "../../config";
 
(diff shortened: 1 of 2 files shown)


src/browser/history/utils.ts
```diff
@@ -4,7 +4,7 @@ import _ from "lodash";
 import { TestStep, TestStepKey } from "../../types";
 import { TimeTravelMode } from "../../config";
 
-const MAX_STRING_LENGTH = 50;
+const MAX_STRING_LENGTH = 500;
```
KuznetsovRoman, line 7: Unit tests need to be adjusted as well:

Image

KuznetsovRoman (approved):

PR #1300, opened 2026-07-26
fix: bring page to front when isolation is turned on

shadowusr:
### What's done?

Context:
- Isolation works by creating new browsing context, essentially meaning creating a new tab or window
- When we perform isolation on older browsers, the window becomes not focused until the next iteration
- This leads to "focus" states on elements to be absent in that window
- We didn't see this issue in Testplane 8, because `resetCursor` was true by default and cursor reset brings the window into focus
- On Testplane 9, `resetCursor` is false by default and this issue became apparent
- Interestingly, this issue won't reproduce on newer browsers like Chrome 144. Supposedly, due to this commit: https://chromium.googlesource.com/chromium/src/+/b5c6f8ac5bb0947e3b772b32350132dfa6426847 that changes the way session window works.

### How I tested?

Reproduced the issue on the following minimal case:
```
await browser.url("https://nda.ya.ru/t/Hbwgz3kb7jqnAM");

await browser.pause(3000);

await browser.saveScreenshot('screenshot-focus.png');
```

On chrome 123 running with selenoid locally, with `headless=false, resetCursor=false, isolation=true`.

Verified that this issue went away after the fix.

### DIFF
--- a/src/browser/existing-browser.ts
+++ b/src/browser/existing-browser.ts
@@ -419,7 +419,11 @@ export class ExistingBrowser extends Browser {
         const { sessionId } = await cdpTarget.attachToTarget(incognitoWindowId);
         // Important for chrome older than ~133, because otherwise "focus" states on the page will be lost in
         // headful mode, with isolation=true and resetCursor=false.
-        await this._cdp.page.bringToFront(sessionId);
+        try {
+            await this._cdp.page.bringToFront(sessionId);
+        } finally {
+            await cdpTarget.detachFromTarget(sessionId).catch(() => {});
+        }
     }
 
     protected async _performIsolation({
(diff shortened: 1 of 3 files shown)


src/browser/existing-browser.ts
```diff
@@ -415,6 +415,11 @@ export class ExistingBrowser extends Browser {
             ...browserContextIdsToClose.map(contextId => cdpTarget.disposeBrowserContext(contextId)),
             ...currentTargets.map(target => cdpTarget.closeTarget(target.targetId).catch(() => {})),
         ]);
+
+        const { sessionId } = await cdpTarget.attachToTarget(incognitoWindowId);
+        // Important for chrome older than ~133, because otherwise "focus" states on the page will be lost in
+        // headful mode, with isolation=true and resetCursor=false.
+        await this._cdp.page.bringToFront(sessionId);
```
KuznetsovRoman: Shouldn't we detach from target after it?

KuznetsovRoman (approved):

PR #1304, opened 2026-07-30
fix: type "NEW_BROWSER" event in worker context

DudaGod:
### What is done

- fix typings for `NEW_BROWSER`  worker-events now available after check `testplane.isWorker()`
- add type tests


test/type-tests/testplane-events.ts
```diff
@@ -0,0 +1,16 @@
+import type Testplane from "../../src";
```
DudaGod, line 1: add type tests which is checked when run  `npm run check-types` or `npm test`

src/testplane.ts
```diff
@@ -316,7 +322,7 @@ export class Testplane extends BaseTestplane {
         });
     }
 
-    isWorker(): boolean {
+    isWorker(): this is this & WorkerEventEmitter<this> {
```
DudaGod, line 325: after check type is expanding by worker events

sonic16x (approved):

KuznetsovRoman (approved):

PR #1307, opened 2026-08-04
fix: discard shrinking the calibration area too much

shadowusr:
### What's done?

Added a safeguard against shinking safe area too much. You can see example image, which previously resulted in calibrated area with width of 8px which is completely incorrect.

The better solution would be building an algorithm for finding the rectangle with the largest area that doesn't have any interferences, but this proved to be not so straightforward.


src/browser/calibrator.ts
```diff
@@ -100,7 +102,7 @@ export class Calibrator {
 
         for (let y = (imageHeight - 1) as Coord<"image", "device", "y">; y >= 0; y--) {
             const result = await findMarkerXBandInRow(y, image, searchColor);
-            if (result) {
+            if (result && result.width > imageWidth * 0.8) {
```
KuznetsovRoman: Why do we have to check `result.width > imageWidth * 0.8` multiple times? Can we only check it in one line of code (either L88 or L105)?

If thats inevitable, lets at least pull "0.8" magic constant to the top of the file

KuznetsovRoman (approved):

PR #1308, opened 2026-08-07
Retry session create with exp timeout

sonic16x:
Screenshot 2026-08-07 at 15 02 05


src/browser/new-browser.ts
```diff
@@ -13,6 +13,7 @@ import { LOCAL_GRID_URL, W3C_CAPABILITIES, VENDOR_CAPABILITIES } from "../consta
 import { Config } from "../config";
 import { BrowserConfig } from "../config/browser-config";
 import { BrowserName, type W3CBrowserName } from "./types";
+import { exponentiallyWait } from "../ws-connection/utils";
```
KuznetsovRoman: Browser should not directly depend on ws-connection utils module

If you want to use "exponentiallyWait" here, its better to pull "exponentiallyWait" to generic utils, which will be used both in ws-connection and browser

sonic16x: Fixed

PR #1311, opened 2026-08-17
fix(selectivity): stabilize navigation coverage

KuznetsovRoman:


src/browser/cdp/selectivity/click-command-guard.ts
KuznetsovRoman, line 1: Clicks need to be intercepted and queued, because chromedriver itself does not wait for navigation to complete, and we need to.
Order:
```
 startClick()
 WebDriver originalClick()
 beforeunload / coverage snapshot
 Debugger.resume
 navigation commit
 DOMContentLoaded или load
 page-start coverage flush
 click() returns the value
```

src/browser/cdp/selectivity/coverage-hooks.ts
KuznetsovRoman, line 1: Pulled out existing cdp hooks from index.ts to here

src/browser/cdp/selectivity/navigation-guard.ts
KuznetsovRoman, line 1: State machine module, that stores navigation state of the browser

KuznetsovRoman:
@codex review

src/browser/cdp/selectivity/click-command-guard.ts
```diff
@@ -0,0 +1,75 @@
...
+const guardedBrowsers = new WeakSet<WebdriverIO.Browser>();
+
+/**
+ * Install one persistent click wrapper because WebdriverIO commands cannot be restored after overwriteCommand.
+ * The active guard can still be replaced or disabled when a browser session is reused.
+ */
+export const enableClickCommandGuard = (browser: WebdriverIO.Browser, guard: ClickCommandGuard): (() => void) => {
+    activeGuards.set(browser, guard);
+
+    if (!guardedBrowsers.has(browser)) {
```
DudaGod: I would write like this:

```
const cb = () => {
    if (activeGuards.get(browser) === guard) {
        guard.cancelClick();
        activeGuards.delete(browser);
    }
};

if (guardedBrowsers.has(browser)) {
  return cb;
}

// a lot of code

return cb;
```

will allow to reduce the nesting a little bit

src/browser/cdp/selectivity/click-command-guard.ts
```diff
@@ -0,0 +1,75 @@
...
+    startClick: () => void;
+    waitForNavigationCompletion: () => Promise<void>;
+    cancelClick: () => void;
+}
+
+const activeGuards = new WeakMap<WebdriverIO.Browser, ClickCommandGuard>();
+const guardedBrowsers = new WeakSet<WebdriverIO.Browser>();
+
+/**
+ * Install one persistent click wrapper because WebdriverIO commands cannot be restored after overwriteCommand.
```
DudaGod, line 11: Are you sure that `install` is correct word here? in fact, nothing is installed here.

DudaGod (approved):

src/browser/cdp/selectivity/click-command-guard.ts
KuznetsovRoman: The click wrapper is installed
Given how hard to undo the "overwriteCommand" effect, it pretty much looks like installation.
e.g. hooks are installed too.

PR #1312, opened 2026-08-17
Update wdio to 9.6.2

sonic16x:


test/src/browser/session-429-retry.ts
```diff
@@ -0,0 +1,79 @@
+import * as http from "node:http";
+import { remote } from "@testplane/webdriverio";
+
+const RETRY_BASE_DELAY = 5000;
+const RETRY_JITTER = 1000;
+
+describe.only("session creation 429 retry", () => {
```
KuznetsovRoman: `.only` are forbidden in VCS.

test/src/browser/session-429-retry.ts
```diff
@@ -0,0 +1,145 @@
+import * as http from "node:http";
+import { URL as NodeURL } from "node:url";
+import { remote } from "@testplane/webdriverio";
+
+// wdio internally uses `URL` as a global — polyfill for envs where it's missing
+if (typeof (globalThis as never as Record<string, unknown>).URL === "undefined") {
+    (globalThis as never as Record<string, unknown>).URL = NodeURL;
+}
```
KuznetsovRoman: What do we need this polyfill for?

URL is available since node 6, and we only support 20+

Image

Therefore, this polyfill is useless

KuznetsovRoman (changes requested):

KuznetsovRoman (approved):

PR #1313, opened 2026-08-17
fix: fix pixel ratio mismatch on pages with out of process iframes

shadowusr:
### What's done?

- Migrated to a new cross-platform chromium image, that supports headful mode
- Implemented a e2e test that exposes an issue with out of process iframes
- Implemented a fix for this behavior

The bug reproduces in headful chrome with pixel ratio emulation, when page has iframe on different origin that creates an iframe on the original origin. In this case, window.devicePixelRatio gets overridden by value inside that iframe (without emultation, because pixel ratio in iframes is not emulated)


src/browser/commands/assert-view/index.js
```diff
@@ -19,8 +19,18 @@ const { AssertViewError } = require("./errors/assert-view-error");
 const makeDebug = require("debug");
 const debug = makeDebug("testplane:screenshots:assert-view");
 
+const HEADLESS_CHROME_ARG_RE = /^-{0,2}headless(?:=|$)/;
+
 const getShortDebugId = debugId => crypto.createHash("sha1").update(debugId).digest("hex").slice(0, 7);
 
+const isHeadlessBrowser = chromeOptions => (chromeOptions?.args || []).some(arg => HEADLESS_CHROME_ARG_RE.test(arg));
```
KuznetsovRoman, line 26: Is it only for chrome?
Shouldn't we remove "chrome" from "HEADLESS_CHROME_ARG_RE"?

KuznetsovRoman (approved):

src/browser/commands/assert-view/index.js
shadowusr: It's chrome-specific bug, yes. So only for chrome.

PR #1314, opened 2026-08-18
feat: testplane profiler

KuznetsovRoman:


src/profiler/schema/v1.ts
```diff
@@ -0,0 +1,302 @@
...
+    arch: string;
+    /** Number of logical CPU execution slots reported by Node.js for this host. */
+    availableParallelism: number;
+    /** Worker count configured for this Testplane run. */
+    configuredWorkers?: number;
+    /** Configured maximum browser sessions, grouped by browser configuration name. */
+    configuredSessionsPerBrowser?: Record<string, number>;
+}
+
+export interface ProfilerResultV1 {
```
KuznetsovRoman, line 273: This object is emitted as event and saved to file

src/profiler/retention/operation-store.ts
```diff
@@ -0,0 +1,131 @@
...
+                    getRetentionBucket(operation.kind) === bucket
+                ) {
+                    retained += 1;
+                }
+            }
+            truncation.push({
+                collector: bucket,
+                seen: state.seen,
+                retained,
+                rule: `top ${RETENTION_POLICY_V1.operationLimits[bucket]} by wall time; keep ancestors of retained operations`,
```
KuznetsovRoman, line 71: Rules, reasons, evidence, confidence and such are saved as human-readable values so we could display clear strings in profiler UI with at least as possible logic

src/profiler/retention/policy-v1.ts
```diff
@@ -0,0 +1,49 @@
+import type { EnabledProfilerLevel } from "../schema";
+
+export const RETENTION_POLICY_V1 = Object.freeze({
```
KuznetsovRoman, line 3: Object configures "how much slowest objects of each type we should keep"

src/profiler/retention/top-k.ts
```diff
@@ -0,0 +1,70 @@
...
+    score: number;
+    value: T;
+}
+
+interface TopKAddResult<T> {
+    accepted: boolean;
+    evicted?: T;
+}
+
+export class TopK<T> {
```
KuznetsovRoman, line 11: classic min-heap topK implementation on array
The least element is stored at the root

src/profiler/retention/streaming-statistics.ts
```diff
@@ -0,0 +1,238 @@
...
+export const DEFAULT_STATISTICS_RESERVOIR_SIZE = 256;
+
+export class StreamingStatistics {
+    private _count = 0;
+    private _sum = 0;
+    private _min = Infinity;
+    private _max = -Infinity;
+    private _mean = 0;
+    private _m2 = 0;
+    private readonly _reservoir: number[] = [];
```
KuznetsovRoman, line 31: [Reservoir Sampling](https://ru.wikipedia.org/wiki/Reservoir_sampling) for percentile readings

src/profiler/retention/top-k.ts
```diff
@@ -0,0 +1,70 @@
...
+                return;
+            }
+            [this._heap[parent], this._heap[index]] = [this._heap[index], this._heap[parent]];
+            index = parent;
+        }
+    }
+
+    private _bubbleDown(index: number): void {
+        for (;;) {
+            const left = index * 2 + 1;
```
KuznetsovRoman, line 52: for binary trees on arrays, child nodes has those indexes

src/profiler/retention/streaming-statistics.ts
```diff
@@ -0,0 +1,238 @@
...
+    p50?: number;
+    /** Estimated 95th percentile from `samples`: about 95% of the observations are at or below it. */
+    p95?: number;
+    /** Bounded random sample used to estimate percentiles without retaining every value. */
+    samples?: number[];
+}
+
+export const DEFAULT_STATISTICS_RESERVOIR_SIZE = 256;
+
+export class StreamingStatistics {
```
KuznetsovRoman, line 24: Besides topK, we also store aggregates: count/sum/min/max/mean/variance

src/profiler/retention/streaming-statistics.ts
```diff
@@ -0,0 +1,238 @@
...
+
+        const nextCount = this._count + 1;
+        const nextSum = this._sum + value;
+        const delta = value - this._mean;
+        const nextMean = this._mean + delta / nextCount;
+        const deviation = value - nextMean;
+        const m2Increment = delta * deviation;
+        const nextM2 = this._m2 + m2Increment;
+
+        if (
```
KuznetsovRoman, line 51: safety guard so we won't randomly pollute all the data with NaN

src/profiler/runtime/async-activity.ts
```diff
@@ -0,0 +1,127 @@
+import { createHook, executionAsyncId, type AsyncHook } from "node:async_hooks";
```
KuznetsovRoman, line 1: createHook is needed here so we would know "how many milliseconds did the code actually run in the Event Loop as part of this span" in order to determine "waiting" (I/O bound) vs "activeJs" (CPU bound)

src/profiler/runtime/bootstrap-probe.ts
KuznetsovRoman, line 1: While the config was not loaded yet, we have no idea "is profiler enabled or not"
So we profile what we can, and if after config parsing turns out "it is enabled", we use the stats. If it is disabled, discard them.

src/profiler/runtime/noop.ts
KuznetsovRoman, line 1: In order to have same API for "profiler enabled/disabled" cases, "NoopProfilerRuntime" was implemented.
Now we can just make records without thinking "will it be used or not"

src/profiler/runtime/types.ts
```diff
@@ -0,0 +1,109 @@
...
+    clockUncertaintyMs?: number;
+    process: ProcessRef;
+    operations: RetainedOperation[];
+    aggregates?: RuntimeAggregate[];
+    metrics?: RuntimeMetric[];
+    errors: ProfilerError[];
+    truncation: TruncationEntry[];
+}
+
+export interface ProfilerRuntimeLike {
```
KuznetsovRoman, line 87: Either real profiler runtime, either noop profiler runtime.

src/profiler/runtime/runtime.ts
```diff
@@ -0,0 +1,506 @@
...
+        this.originMonotonicMs = options.originMonotonicMs ?? this._clock.now();
+        this.originEpochMs =
+            options.originEpochMs ?? this._clock.epochNow() - (this._clock.now() - this.originMonotonicMs);
+
+        if (this.level >= 3) {
+            try {
+                this._asyncActivity = new AsyncActivityTracker(
+                    this._clock,
+                    () => this._context.getStore()?.spanId,
+                    error => this.recordError("runtime.asyncActivity", error),
```
KuznetsovRoman, line 165: In order to not just silently ignore profiler errors, but dont consider them as critical, they are recorded to profiler errors

src/profiler/runtime/runtime.ts
```diff
@@ -461,6 +536,205 @@ export class ProfilerRuntime {
...
+
+    private _startResourceSampler(): void {
+        if (typeof performance.eventLoopUtilization === "function") {
+            this._lastElu = performance.eventLoopUtilization();
+        }
+        try {
+            this._eventLoopDelay = monitorEventLoopDelay({
+                resolution: EVENT_LOOP_DELAY_RESOLUTION_MS,
+            });
+            this._eventLoopDelay.enable();
```
KuznetsovRoman: Discover cases when large synchronous task stops the whole worker from running multiple session tests

src/profiler/runtime/runtime.ts
```diff
@@ -109,6 +131,8 @@ export class ProfilerRuntime {
         originMonotonicMs?: number;
         /** Unix timestamp in milliseconds that corresponds to `originMonotonicMs`. */
         originEpochMs?: number;
+        /** Estimated maximum error when this runtime clock is aligned with another process. */
+        clockUncertaintyMs?: number;
```
KuznetsovRoman, line 148: We can't perfectly align several processes time, because we dont know "when worker sent the message exactly"
So we assume delay is half-ping time

src/profiler/runtime/runtime.ts
```diff
@@ -638,18 +849,22 @@ export class ProfilerRuntime {
                 .map(([key, dimension]) => [key.slice(0, 100), dimension]),
         );
         let normalizedMetric = metric;
-        let key = `${normalizedMetric} ${stableAttributesKey(normalizedDimensions)}`;
+        let key = `${normalizedMetric}\0${stableAttributesKey(normalizedDimensions)}`;
```
KuznetsovRoman, line 953: `\0` is join-character to generate uniq keys

src/profiler/runtime/runtime.ts
```diff
@@ -847,6 +1144,243 @@ function readHostCpu(): { busy: number; total: number } {
     return { busy, total };
 }
 
+type ValueGuard<T = unknown> = (value: unknown) => value is T;
+
+const isProcessRef = recordShape<ProcessRef>({
```
KuznetsovRoman, line 1254: Because worker sends messages of type "unknown" (and they can be sent by user's code) we have runtime validation of messages here, so random message, sent by user, would not spoil profiler measures.
With those, we can safely rely on typescript typings

src/profiler/analysis/aggregator.ts
```diff
@@ -0,0 +1,299 @@
...
+    operationName: ProfilerOperationName = "run",
+): NormalizedProfile => {
+    const errors = [...snapshot.errors];
+    const rootId = `${snapshot.runId}:operation`;
+    const timeline: RetainedOperation[] = snapshot.operations.map(operation => ({
+        ...operation,
+        timing: { ...operation.timing },
+        parentId: operation.parentId ?? rootId,
+    }));
+    const root: RetainedOperation = {
```
KuznetsovRoman, line 37: synthetic root for all child operations

src/profiler/analysis/compact.ts
```diff
@@ -0,0 +1,51 @@
...
+): { timeline: RetainedOperation[]; findings: Finding[] } {
+    const compactIdByOriginalId = new Map<string, string>();
+    const root = timeline.find(operation => operation.kind === "testplane.operation");
+    const operationsInIdOrder = root ? [root, ...timeline.filter(operation => operation !== root)] : timeline;
+
+    for (const operation of operationsInIdOrder) {
+        compactIdByOriginalId.set(operation.id, `op:${compactIdByOriginalId.size.toString(36)}`);
+    }
+
+    const getCompactId = (originalId: string): string => {
```
KuznetsovRoman, line 15: each operation has its uniq id
When they are generated as uuid, those are hugely increase result size, so compact readable id is used

src/profiler/analysis/policy-v1.ts
```diff
@@ -0,0 +1,39 @@
+export const ANALYSIS_POLICY_V1 = Object.freeze({
```
KuznetsovRoman, line 1: heuristics tresholds.

src/profiler/analysis/analyzers.ts
KuznetsovRoman, line 1: Commits "analyze *" introduce basic analyzers

Analyzers act on the following pattern:
- Accept profile
- check if there is a match on heuristics
- write finding with evidence and human-readable suggestions

src/profiler/sanitize.ts
KuznetsovRoman, line 1: In profiler we collect a lot of data, and we want users to be able to safely share profiler result with us
For that reason we strip sensitive info. We also strip "just very huge strings", non-standard objects, functions and other stuff

src/profiler/retention/result-budget.ts
```diff
@@ -0,0 +1,129 @@
...
+            }
+        }
+        if (!changed) {
+            return;
+        }
+    }
+}
+
+function trimDiagnostics(result: ProfilerResultV1, budget: number): void {
+    const arrays: unknown[][] = [result.profiler.collectionErrors, result.run.partialReasons, result.findings];
```
KuznetsovRoman, line 64: Those represent the order we strip data from json, when it weights too much

src/profiler/output/json.ts
```diff
@@ -0,0 +1,22 @@
...
+    cwd = process.cwd(),
+): Promise<void> => {
+    const destination = path.resolve(cwd, outputPath);
+    const directory = path.dirname(destination);
+    const temporary = path.join(directory, `.${path.basename(destination)}.${result.run.id}.tmp`);
+
+    await fs.mkdir(directory, { recursive: true });
+    try {
+        await fs.writeFile(temporary, `${JSON.stringify(result, null, 2)}\n`, { encoding: "utf8", flag: "wx" });
+        await fs.rename(temporary, destination);
```
KuznetsovRoman, line 17: atomic write through rename

src/profiler/output/console.ts
KuznetsovRoman, line 1: Outputs human-readable result with findings and suggestions to console

src/events/async-emitter/index.ts
```diff
@@ -1,15 +1,327 @@
...
 export class AsyncEmitter extends EventEmitter {
-    async emitAndWait(event: string | symbol, ...args: unknown[]): Promise<unknown[]> {
-        const results = await Promise.allSettled(
-            this.listeners(event).map(l => promiseMethod(l as (...args: unknown[]) => unknown).apply(this, args as [])),
+    private _eventObserver?: EventObserver;
+    private readonly _wrappedListeners = new WeakMap<Listener, WrappedListener>();
+    private readonly _onceWrappers = new WeakMap<Listener, Listener>();
+    private readonly _asyncDispatch = new WeakSet<Listener>();
+
+    constructor(observer?: EventObserver) {
```
KuznetsovRoman, line 33: AsyncEmitter can receive `EventObserver` to track event info

src/events/async-emitter/index.ts
```diff
@@ -1,15 +1,327 @@
 import { EventEmitter } from "events";
 import { promiseMethod } from "../../utils/promise";
 
+type Listener = {
+    bivarianceHack(this: unknown, ...args: unknown[]): unknown;
+}["bivarianceHack"];
```
KuznetsovRoman, line 6: We can't extract `(...args: unknown[]) => unknown` directly because of TS rules, so we use intermediate object in order to get "some listener" type

src/events/async-emitter/index.ts
```diff
@@ -1,15 +1,327 @@
...
+    observeAsyncEmission<T>(event: string | symbol, action: () => Promise<T>): Promise<T>;
+    error(stage: string, error: unknown): void;
+}
+
+interface WrappedListener {
+    original: Listener;
+    registration: unknown;
+}
+
 export class AsyncEmitter extends EventEmitter {
```
KuznetsovRoman, line 27: A lot of methods were implemented explicitly in order for event observer wrapper to function seamlessly

src/profiler/runtime/event-observer.ts
```diff
@@ -0,0 +1,177 @@
...
+import type { SourceRef } from "../schema";
+
+interface ProfilerEventRegistration {
+    event: string;
+    functionName: string;
+    origin: "internal" | "user";
+    source?: SourceRef;
+}
+
+class ProfilerEventObserver implements EventObserver {
```
KuznetsovRoman, line 18: Observes the time it took to wait for event handler with "best-effort" location discovery

src/config/index.ts
```diff
@@ -28,41 +34,60 @@ export class Config {
         }
     }
 
-    static async read(configPath: string): Promise<ConfigInputData> {
+    static async read(configPath: string, observer?: ConfigLifecycleObserver): Promise<ConfigInputData> {
+        const phase = observer?.startPhase("testplane.phase.config.read", "Read configuration file");
```
KuznetsovRoman, line 39: In some places this syntax (without wrapper) is cleaner

src/base-testplane.ts
```diff
@@ -72,6 +122,10 @@ export abstract class BaseTestplane extends AsyncEmitter {
         return Errors;
     }
 
+    async profileCliCommand<T>(command: string, action: () => T | Promise<T>): Promise<T> {
```
KuznetsovRoman, line 123: Other CLI commands are profiled too

src/utils/module-observer-registry.ts
KuznetsovRoman, line 1: both "selectivity" and "profiler" depend on tracking "which files are loaded"
So module observer is pulled to another mutual module, which both selectivity and profiler could use

src/browser/history/index.ts
KuznetsovRoman, line 1: Using existing history commands decorator to track "how much time it took to execute browser commands"

src/profiler/collectors/pool-observer.ts
KuznetsovRoman, line 1: tracks browser pool operations (time to create/reset/end session), counts new/reused/free sessions, observes pool state (active/running sessions, queue)
Is needed so we could distinguish between "slow session create" and "pool waits" in order to see, were we limited by "sessionsPerBrowser" or "parallelLimit"

src/runner/test-runner/regular-test-runner.js
KuznetsovRoman, line 1: Master tracks test:
- get/free session time
- full test.attempt time
- result handling

src/worker/runner/test-runner/index.js
KuznetsovRoman, line 1: Worker tracks test:
- prepare
- beforeEach/afterEach
- test body

src/profiler/collectors/module-graph.ts
```diff
@@ -0,0 +1,119 @@
...
+interface ModuleContext {
+    ownerFile: string;
+    stack: SpanHandle[];
+}
+
+interface LoadToken {
+    span: SpanHandle;
+}
+
+export class ProfilerModuleGraphCollector {
```
KuznetsovRoman, line 17: Level 3 only
Tracks time to load modules, require cache hit, binds transitive modules to parent module

src/constants/process-messages.ts
```diff
@@ -5,3 +5,7 @@ export const WORKER_INIT = "worker.init";
 export const WORKER_SYNC_CONFIG = "worker.syncConfig";
 export const WORKER_UNHANDLED_REJECTION = "worker.unhandledRejection";
 export const TEST_ASSIGNED_TO_WORKER = "worker.testAssignedToWorker";
+
+export const WORKER_PROFILER_BATCH = "worker.profilerBatch";
+export const MASTER_PROFILER_FLUSH = "master.profilerFlush";
+export const WORKER_PROFILER_FLUSHED = "worker.profilerFlushed";
```
KuznetsovRoman, line 11: Worker sends profile nodes in batches
in order to ensure "worker has empty batch", master sends "master.profilerFlush" command, then worker answers with "worker.profilerFlushed", and after master got "worker.profilerFlushed", we can be sure worker sent all profile data it had.

src/worker/testplane-facade.ts
```diff
@@ -56,11 +79,23 @@ module.exports = class TestplaneFacade {
...
+                        workerInstanceId?: string;
+                        clockSync?: {
+                            workerSentAtEpochMs: number;
+                            masterReceivedAtEpochMs: number;
+                            masterSentAtEpochMs: number;
+                        };
+                    };
                 }) => {
                     try {
+                        const clockAlignment = calculateClockAlignment(profiler?.clockSync, Date.now());
```
KuznetsovRoman, line 97: NTP-style clock synchronization between a worker and the master, because they have independent clocks.

src/worker/browser-env/runner/test-runner/index.ts
```diff
@@ -210,11 +210,30 @@ export class TestRunner extends NodejsEnvTestRunner {
                 });
 
                 this._socket.on(BrowserEventNames.runBrowserCommand, this._handleRunBrowserCommand(browser));
+                this._socket.on(BrowserEventNames.profilerFragment, payload => {
+                    this._profiler.recordMeasurement("browser.runnable", payload.wallMs, {
```
KuznetsovRoman, line 214: Recording somewhat same values on browser env, as we do on node.js runtime, but it lacks:
- detailed hook spans
- test body
- commands
- event loop utilization/delay
- CPU
We also have just one browser runnable here, which would just help to decide between "slow test" and "slow transport"

src/utils/exit-code.ts
KuznetsovRoman, line 1: After testplane profiler finalization, process might already have error code, but then get to another process.exit, which messes up exit code.
In this module we calculate origin exit code

src/cli/commands/config/index.ts
KuznetsovRoman, line 1: In "profile CLI commands" we wrap other CLI commands to same profiler flow

src/config/options.js
```diff
@@ -136,6 +136,35 @@ const rootSection = section(
             }),
         }),
 
+        profiler: section({
```
KuznetsovRoman, line 139: Describe profiler root section

src/config/defaults.js
```diff
@@ -113,6 +113,10 @@ module.exports = {
         output: ".testplane/failed.json",
         input: ".testplane/failed.json",
     },
+    profiler: {
+        level: 0,
```
KuznetsovRoman, line 117: Disabled by default in order to not waste extra CPU cycles in CI

KuznetsovRoman:
@codex review

src/cli/commands/install-deps/index.ts
```diff
@@ -31,78 +32,85 @@ export const registerCmd = (cliTool: typeof commander, testplane: Testplane): vo
...
+                    logResult(
+                        browsersInstallPerStatus,
+                        "error",
+                        BrowserInstallStatus.Error,
+                        "An error occured while trying to download these browsers:",
+                        ({ tag, reason }) => `- ${tag}: ${reason}`,
+                    );
+
+                    if (browsersInstallPerStatus[BrowserInstallStatus.Error].length) {
+                        process.exitCode = resolveExitCode(1);
```
DudaGod, line 110: looks like here callback ends successfully and return `undefined` in `action` inside ProfilerManager is this expected behavior? In this case `runOutcome` would be `passed`

src/runner/browser-env/vite/plugins/generate-index-html.ts
```diff
@@ -145,15 +157,31 @@ function generateTemplate(env: WorkerInitializePayload, runUuid: string): string
                 return mod;
             }
         </script>
-        <script type="module" src="${MODULE_NAMES.globals}"></script>
-        <script type="module" src="${MODULE_NAMES.mocha}"></script>
-        <script type="module" src="${MODULE_NAMES.browserRunner}"></script>
+        <script type="module" src="${viteFsUrl(modules.globals)}"></script>
+        <script src="${MOCHA_RUNTIME_URL}"></script>
```
DudaGod, line 161: why we move vite module loading with `@testplane/mocha` to script `/__testplane__/mocha.js` ?

src/profiler/output/json.ts
DudaGod: using public api we can call `readTests()` with configured profiler and then `run` method. Looks like in this case I will get only profiler with the last call. Is this intentionally behavior?

DudaGod (approved):
I suggest adding info about  new feature to docs

src/profiler/output/json.ts
KuznetsovRoman: Yes, this is intended behavior. `readTests()` and `run()` are separate top-level profiler operations. Each produces a new result with a new run.id. Both results are printed and emitted through `PROFILER_RESULT`
If user wants the combined profile, he can call:
```js
await testplane.run(testPaths);
```

src/cli/commands/install-deps/index.ts
KuznetsovRoman: Good catch, fixed in latest commit "fix: profile install dependency failures"

src/runner/browser-env/vite/plugins/generate-index-html.ts
KuznetsovRoman: There was a bug: Relative browser-env runtime URLs were resolved under /run-uuids/, causing Vite’s HTML fallback to return an error page instead of runtime JavaScript, which resulted in "Browser didn't connect to the Vite server in 10000ms"

The bug predates the profiler and was only discovered while browser runtime telemetry increased browser-env coverage.

Runtime ESM files now use rooted /@fs URLs, Mocha is served as a classic UMD script, route parsing is stricter, and the Testplane package root is allowed by Vite, which fixes the issue.

PR #1318, opened 2026-08-23
Fix halt test

sonic16x:
Add exit for test when halt and also add pass error to test in this case.

### DIFF
--- a/src/browser/commands/assert-view/errors/no-ref-image-error.ts
+++ b/src/browser/commands/assert-view/errors/no-ref-image-error.ts
@@ -30,5 +30,6 @@ export class NoRefImageError extends BaseStateError {
         super(stateName, currImg, refImg);
 
         this.message = `can not find reference image at ${this.refImg.path} for "${stateName}" state`;
+        this.stack = undefined;
     }
 }
--- a/src/browser/commands/assert-view/errors/no-ref-image-error.ts
+++ b/src/browser/commands/assert-view/errors/no-ref-image-error.ts
@@ -30,6 +30,5 @@ export class NoRefImageError extends BaseStateError {
         super(stateName, currImg, refImg);
 
         this.message = `can not find reference image at ${this.refImg.path} for "${stateName}" state`;
-        this.stack = undefined;
     }
 }
--- a/src/runner/browser-runner.ts
+++ b/src/runner/browser-runner.ts
@@ -26,6 +26,7 @@ export class BrowserRunner extends CancelableEmitter {
     private workers: Workers;
     private running: PromiseGroup;
     private profiler: ProfilerRuntimeLike;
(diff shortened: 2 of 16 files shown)


src/testplane.ts
```diff
@@ -362,6 +362,8 @@ export class Testplane extends BaseTestplane {
 
         this._fail();
 
+        signalHandler.emit(MasterEvents.EXIT, err);
```
shadowusr, line 457: Why is this one needed?

shadowusr (commented):
I'd wait with this one until issues in https://github.com/gemini-testing/html-reporter/pull/800 are resolved.

src/utils/format-error.ts
```diff
@@ -0,0 +1,15 @@
+export const formatError = (err: Error): Error => {
+    const TRIM_MARKERS = ["Tests were stopped by the user"];
```
sonic16x, line 2: Remove stack for some errors, now only for force stop

src/testplane.ts
sonic16x: Need for send event to reporter

src/browser/commands/assert-view/errors/no-ref-image-error.ts
```diff
@@ -30,5 +30,6 @@ export class NoRefImageError extends BaseStateError {
         super(stateName, currImg, refImg);
 
         this.message = `can not find reference image at ${this.refImg.path} for "${stateName}" state`;
+        this.stack = undefined;
```
sonic16x, line 33: Remove stack for no ref image error

shadowusr: imho error formatting should happen on reporter's side, not on the source side. This affects both this error and tests were stopped by user error. But that's not critical. Just more clean in my opinion — the source should give full info, the receiver should filter what's needed.

sonic16x: Removed here

sonic16x: Now realized on reporter side, also for 429 and no ref image errors

src/runner/browser-runner.ts
```diff
@@ -72,6 +74,10 @@ export class BrowserRunner extends CancelableEmitter {
         });
         const runner = TestRunner.create(test, this.config, browserAgent, this.profiler);
 
+        if (this.cancelError && typeof runner.cancel === "function") {
```
shadowusr, line 77: When is runner.cancel not a function?

shadowusr (approved):

src/runner/browser-runner.ts
sonic16x: It check for ts

PR #1326, opened 2026-09-09
fix: avoid leaking mocha types from profiler clock

KuznetsovRoman:


src/profiler/runtime/clock.ts
```diff
@@ -1,24 +1,29 @@
...
+    system: number;
+}
+
 export interface ProfilerClock {
     /** High-resolution monotonic time in milliseconds. It is suitable for measuring elapsed time. */
     now(): number;
     /** Unix wall-clock timestamp in milliseconds. It is suitable for timestamps shared between processes. */
     epochNow(): number;
     /** Process CPU usage since startup, or since `previousValue` when it is supplied. */
-    cpuUsage(previousValue?: NodeJS.CpuUsage): NodeJS.CpuUsage;
```
KuznetsovRoman, line 9: "NodeJS.CpuUsage" causes typescript to add reference to mocha types

shadowusr: hmmmmmm why is that so, I wonder?

shadowusr (approved):

src/profiler/runtime/clock.ts
KuznetsovRoman: Because NodeJS is a global namespace, which is expanded by:
- `@types/node`
- `@types/mocha`
- `expect-webdriverio`

So all those 3 are referenced

PR #1333, opened 2026-09-18
feat: support browser download mirrors

KuznetsovRoman:


KuznetsovRoman:
@codex review

docs/dealing-with-browsers.md
```diff
@@ -2,30 +2,104 @@
...
+You can also request explicit versions:
+
+```bash
+npx testplane install-deps chrome@130 firefox@128
 ```
 
-and run your server by executing
+If `install-deps` is not run first, Testplane can download missing local dependencies when the browser session starts. For a remote grid, set `gridUrl` to its WebDriver endpoint instead of `"local"`.
+
+For local browser and driver installation, a `browserVersion` consisting only of digits is normalized by adding `.0`: `"139"` is treated as `"139.0"`, and `"144"` as `"144.0"`. Versions containing dots or channel names are not rewritten by this normalization. The same rule applies to explicit `install-deps` versions such as `chrome@139`.
```
sipayRT, line 45: I think that one example with 139.0 is enough (the second example with 144 is not needed). But we need to add an example about 144.1

src/browser-installer/chrome/browser.ts
```diff
@@ -40,18 +51,23 @@ const installChromeBrowser = async (
...
 
     if (!canBeInstalled) {
         throw new Error(
-            [
-                `${browserName}@${version} can't be installed.`,
-                `Probably the version '${version}' is invalid, please try another version.`,
-                "Version examples: '120', '120.0'",
-            ].join("\n"),
+            mirror
+                ? "Couldn't download browser artifact from the configured mirror"
```
sipayRT, line 65: pls, add current mirror here

src/browser-installer/chrome/browser.ts
```diff
@@ -40,18 +51,23 @@ const installChromeBrowser = async (
...
+                : [
+                      `${browserName}@${version} can't be installed.`,
+                      `Probably the version '${version}' is invalid, please try another version.`,
+                      "Version examples: '120', '120.0'",
```
sipayRT, line 69: is example "120.0" still valid?

src/browser-installer/chrome/utils.ts
```diff
@@ -0,0 +1,70 @@
...
+    if (EXACT_VERSION_PATTERN.test(selector)) {
+        return selector;
+    }
+
+    const isChannel = CHANNELS.has(selector);
+    const isMilestone = MILESTONE_VERSION_PATTERN.test(selector);
+    let filename: string;
+
+    if (isChannel) {
+        const channel = selector === "latest" ? "canary" : selector;
```
sipayRT, line 37: why should "latest" be "canary"? is this the same?

src/browser-installer/firefox/browser.ts
```diff
@@ -57,31 +78,66 @@ const installFirefoxBrowser = async (version: string, { force = false } = {}): P
...
+            const response = await retryFetch(
+                getBrowserDownloadMirrorFileUrl(mirror, FIREFOX_VERSIONS_LATEST_VERSIONS_FILENAME),
+            );
+
+            if (!response.ok) {
+                throw new Error(MIRRORED_FIREFOX_VERSION_ERROR);
+            }
+
+            const data = (await response.json()) as { LATEST_FIREFOX_VERSION?: unknown } | null;
+            const version = data?.LATEST_FIREFOX_VERSION;
```
sipayRT, line 130: you can use `trim` here once

src/browser-installer/mirrors.ts
```diff
@@ -0,0 +1,76 @@
...
+    if (!trimmedMirror) {
+        throw new Error(`"${optionName}" must not be empty`);
+    }
+
+    let url: URL;
+
+    try {
+        url = new URL(trimmedMirror);
+    } catch {
+        throw new Error(`"${optionName}" must be an absolute http: or https: URL`);
```
sipayRT, line 23: maybe we can change error text here to something like "wrong url". you already have an error about http/https below

src/browser/new-browser.ts
```diff
@@ -103,33 +112,52 @@ export class NewBrowser extends Browser {
...
-        } catch (e) {
-            warn(`WARNING: Can not close session: ${(e as Error).message}`);
-            this._wdProcess?.kill();
-        } finally {
-            this._wdProcess = null;
+        } catch {
+            return;
         }
+
+        await this._cleanup(true);
```
sipayRT: lets change it to `{ canReuseWebdriver: true}`

src/config/defaults.js
```diff
@@ -6,6 +6,11 @@ const { TimeTravelMode } = require("./types");
 
 module.exports = {
     baseUrl: "http://localhost",
+    browserDownloadMirrors: {
```
sipayRT, line 9: why is this not a browser option?

sipayRT (approved):

src/browser-installer/chrome/browser.ts
KuznetsovRoman: Yes, why not?

src/browser-installer/chrome/utils.ts
KuznetsovRoman: By puppeteer logic, yes
What you probably mean by "latest" is "stable", but "latest" === "canary"

src/browser-installer/firefox/browser.ts
KuznetsovRoman: Not quite. Version could be not a string here, so by using "trim" here once we have to throw error twice, so, no difference

